PR #40, opened 2015-11-15
Move opt/spinnaker/config into application's main

gregturn:


cfieber:
Fixed via #41

gregturn:
That works too! Great.

PR #41, opened 2015-11-17
inject the packaging specific spring config location via startScripts extension

cfieber:
@ewiseblatt @dstengle FYI

This generates a legit start script for the deb file without interfering with the bootRun task

### DIFF
--- a/echo-web/echo-web.gradle
+++ b/echo-web/echo-web.gradle
@@ -23,7 +23,7 @@ apply plugin: 'project-report'
 apply plugin: 'nebula.ospackage'
 
 applicationName = 'echo'
-applicationDefaultJvmArgs = ["-Djava.security.egd=file:/dev/./urandom", "-Dspring.config.location=/opt/spinnaker/config/"]
+applicationDefaultJvmArgs = ["-Djava.security.egd=file:/dev/./urandom"]
 applicationDistribution.from(project.file('config')) {
     into 'config'
 }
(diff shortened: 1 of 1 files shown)


echo-web/echo-web.gradle
```diff
@@ -88,8 +88,15 @@ jar {
 }
 
 startScripts {
+    defaultJvmOpts = applicationDefaultJvmArgs + ["-Dspring.config.location=/opt/spinnaker/config/"]
     doLast {
-        unixScript.text = unixScript.text.replace('DEFAULT_JVM_OPTS=', '\n if [ -f /etc/default/spinnaker ];then \n set -a\n . /etc/default/spinnaker \n set +a\n fi \n DEFAULT_JVM_OPTS=')
+        unixScript.text = unixScript.text.replace('DEFAULT_JVM_OPTS=', '''\
```
cfieber: just a bit more readable..

echo-web/echo-web.gradle
```diff
@@ -133,5 +140,5 @@ ospackage {
 
 
 buildDeb {
-  dependsOn installApp
+  dependsOn installDist
```
cfieber: deprecation warning fix (unrelated really to this change)

ewiseblatt:
Yea, this looks good. It's how it was before (at least for bootRun) so should work. This probably also restores $HOME/.spinnaker if that is still baked into the Main class initialization.

PR #43, opened 2015-11-20
allow configuration of multiple rest endpoints to send events to

tomaslin:
[irrelevant]


tomaslin:
@spinnaker/netflix-reviewers - PTAL

PR #60, opened 2016-01-12
add an endpoint that will accept arbitrary events for webhooks

tomaslin:


tomaslin:
@ajordens PTAL

echo-webhooks/src/main/groovy/com/netflix/spinnaker/echo/controllers/WebhooksController.groovy
```diff
@@ -0,0 +1,48 @@
...
+
+@RestController
+@Slf4j
+class WebhooksController {
+
+  @Autowired
+  EventPropagator propagator
+
+  @RequestMapping(value = '/webhooks/{type}/{source}', method = RequestMethod.POST)
+  void forwardEvent(@RequestBody Map postedEvent, @PathVariable String type, @PathVariable String source) {
```
ajordens: The only thing that stood out (and it's super minor) ... but for consistency across service controllers I'd suggest putting the `@RequestBody` as the last parameter rather than the first.

ajordens:
Nice and simple, LGTM.

echo-webhooks/src/test/groovy/com/netflix/spinnaker/echo/controllers/WebhooksControllerSpec.groovy
```diff
@@ -0,0 +1,46 @@
...
+ */
+
+package com.netflix.spinnaker.echo.controllers
+
+import com.netflix.spinnaker.echo.events.EventPropagator
+import spock.lang.Specification
+
+class WebhooksControllerSpec extends Specification {
+
+  void 'emits a tranformed event for every webhook event'() {
```
duftler: s/tranformed/transformed

netwolfuk:
Is there an example payload to send to the webhook?
I've cobbled together some JSON which contains just project and application, but it's not starting my pipeline.
My pipeline does not currently have a trigger configured. Would I need to configure a trigger which "listens" for my webhook POST?

tomaslin:
@netwolfuk you would still need to add the processing of the trigger to the pipeline listener so that a pipeline will fire based on your events https://github.com/spinnaker/echo/commit/8bb344c2bca5164d105328e9ef8d88614ead7a90

tomaslin:
and add it to deck

PR #66, opened 2016-02-04
stash trigger

tomaslin:
add the ability to trigger off a pipeline based on a git trigger from stash

### DIFF
--- a/echo-pipelinetriggers/src/main/java/com/netflix/spinnaker/echo/pipelinetriggers/BuildEventMonitor.java
+++ b/echo-pipelinetriggers/src/main/java/com/netflix/spinnaker/echo/pipelinetriggers/BuildEventMonitor.java
@@ -28,8 +28,8 @@ import rx.functions.Func1;
 @Slf4j
 public class BuildEventMonitor implements EchoEventListener {
 
-  public static final String ECHO_EVENT_TYPE = "build";
-  public static final String TRIGGER_TYPE = "jenkins";
+  public static final String JENKINS_TRIGGER_TYPE = "jenkins";
+  public static final String GIT_TRIGGER_TYPE = "git";
 
   private final ObjectMapper objectMapper = new ObjectMapper();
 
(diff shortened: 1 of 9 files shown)


tomaslin:
@robfletcher PTAL

echo-pipelinetriggers/src/main/java/com/netflix/spinnaker/echo/pipelinetriggers/BuildEventMonitor.java
```diff
@@ -121,8 +136,13 @@ private void onMatchingPipeline(Pipeline pipeline) {
     log.info("Found matching pipeline {}:{}", pipeline.getApplication(), pipeline.getName());
     val id = registry.createId("pipelines.triggered")
       .withTag("application", pipeline.getApplication())
-      .withTag("name", pipeline.getName())
-      .withTag("job", pipeline.getTrigger().getJob());
+      .withTag("name", pipeline.getName());
+    if (pipeline.getTrigger().getType() == JENKINS_TRIGGER_TYPE) {
+      id.withTag("job", pipeline.getTrigger().getJob());
+    }
+    if (pipeline.getTrigger().getType() == GIT_TRIGGER_TYPE) {
```
robfletcher: should this be `else if`?

echo-scheduler/src/test/groovy/com/netflix/spinnaker/echo/scheduler/actions/pipeline/PipelineConfigsPollingAgentSpec.groovy
```diff
@@ -36,7 +36,7 @@ class PipelineConfigsPollingAgentSpec extends Specification {
 
     void 'when a new pipeline trigger is added, a scheduled action instance is registered with an id same as the trigger id'() {
         given:
-        Trigger trigger = new Trigger(true, null, 'cron', null, null, null, null, '* 0/30 * * * ? *')
+        Trigger trigger = new Trigger(true, null, 'cron', null, null, null, null, '* 0/30 * * * ? *', null, null, null, null)
```
robfletcher: Feels like we could do with some constructor overloads or a factory method here but it can wait.

PR #73, opened 2016-03-02
Factor BuildEvent into BuildEvent and GitEvent

lwander:
@tomaslin


echo-pipelinetriggers/src/main/java/com/netflix/spinnaker/echo/pipelinetriggers/GitEventMonitor.java
```diff
@@ -0,0 +1,140 @@
...
+      .subscribe(triggerEachMatchFrom(pipelineCache.getPipelines()));
+  }
+
+  private void onEchoResponse(final GitEvent event) {
+    registry.gauge("echo.events.per.poll", 1);
+  }
+
+  private Action1<GitEvent> triggerEachMatchFrom(final List<Pipeline> pipelines) {
+    return event -> {
+      Observable.from(pipelines)
```
tomaslin: feels like we can move some of this stuff into the superclass without having to repeat it for every provider type

tomaslin:
@robfletcher

echo-pipelinetriggers/src/main/java/com/netflix/spinnaker/echo/pipelinetriggers/GitEventMonitor.java
lwander: I thought the same thing, but GitEvent and BuildEvent don't inherit from Event, making it tricky. Do we need to keep Event around as is?

robfletcher: I agree with separating the events into different types but is there a pressing reason not to have an event interface or something so that we can unify some of this.

lwander: I agree, but do you know if we need to keep `Event.content` as a `Map`: https://github.com/spinnaker/echo/blob/master/echo-model/src/main/groovy/com/netflix/spinnaker/echo/model/Event.groovy?

tomaslin: otherwise we would need to map every possible field from orca

Would suggest adding a TriggerEvent class for orca-pipelinetriggers and model Git, Cron and BuildEvents off that since it is essentially a different module and we don't have to deal with orca events as triggers.

lwander: SGTM

lwander:
@tomaslin PTAL

tomaslin:
defer to @robfletcher since he's the original author.

echo-model/src/main/groovy/com/netflix/spinnaker/echo/model/trigger/GitEvent.groovy
```diff
@@ -0,0 +1,44 @@
...
+ * limitations under the License.
+ */
+
+package com.netflix.spinnaker.echo.model.trigger
+
+import com.fasterxml.jackson.annotation.JsonIgnore
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties
+import groovy.transform.Canonical
+
+@Canonical
```
robfletcher: Does it have any implications that this isn't on the superclass?

echo-model/src/main/groovy/com/netflix/spinnaker/echo/model/trigger/TriggerEvent.groovy
```diff
@@ -0,0 +1,23 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.spinnaker.echo.model.trigger
+
+import com.netflix.spinnaker.echo.model.Metadata
+
+class TriggerEvent {
```
robfletcher: Should this be abstract?

robfletcher:
Looks good. I have a couple of questions above but they're minor things.

echo-model/src/main/groovy/com/netflix/spinnaker/echo/model/trigger/GitEvent.groovy
lwander: If I understand it correctly, it should be fine since we are never instantiating the superclass. But I'll defer to you on that.

echo-model/src/main/groovy/com/netflix/spinnaker/echo/model/trigger/TriggerEvent.groovy
lwander: Good point

echo-model/src/main/groovy/com/netflix/spinnaker/echo/model/trigger/GitEvent.groovy
robfletcher: It means any fields on the superclass won't participate in `equals`, `hashCode` or `toString` but that's not an issue here I guess.

lwander: Ah, I thought it was used for generating constructors. None of those methods are used for the `TriggerEvent` subclasses, so it should be fine. 

Ready to merge?

robfletcher: Yep

PR #81, opened 2016-04-06
Add Webhook trigger to echo: PR for code review

netwolfuk:
This PR is not ready to merge yet. It's for review only.

The main point of discussion is probably that I added a  "category" to the Trigger.
This is because the EventListener in echo usually checks the "type" of the trigger.
The initial proposal by @tomaslin was for the webhook URL to contain the type and source, but makes it difficult to determine that the trigger should only fire for webhook triggers, not any trigger.

I have added a new webhooks end point:   /webhooks/trigger/{category}/{source}
This is because in the WebHookEvent monitor, it made sense to first check that the event was triggered by the webhook POST, not just any "type".

Otherwise, it would be possible to create a Webhook trigger in Deck with a type of Cron or Jenkins and it would allow webhooks to emulate other trigger types. I wasn't really sure this was a good plan or not, but am happy to hear other views. Perhaps this is the power of the webhook endpoint and provides a way to trigger other random triggers intentionally. I just thought it would be too easy to trigger them unintentionally, which would be confusing.

### DIFF
--- a/echo-model/src/main/groovy/com/netflix/spinnaker/echo/model/Metadata.groovy
+++ b/echo-model/src/main/groovy/com/netflix/spinnaker/echo/model/Metadata.groovy
@@ -23,7 +23,6 @@ package com.netflix.spinnaker.echo.model
 class Metadata {
     String source
     String type
-    String category
     String created = new Date().time
     String organization
     String project
--- a/echo-model/src/main/groovy/com/netflix/spinnaker/echo/model/Metadata.groovy
+++ b/echo-model/src/main/groovy/com/netflix/spinnaker/echo/model/Metadata.groovy
@@ -23,7 +23,7 @@ package com.netflix.spinnaker.echo.model
 class Metadata {
     String source
     String type
-    String triggeredBy
+    String category
     String created = new Date().time
     String organization
     String project
--- a/echo-model/src/main/groovy/com/netflix/spinnaker/echo/model/Metadata.groovy
+++ b/echo-model/src/main/groovy/com/netflix/spinnaker/echo/model/Metadata.groovy
@@ -23,6 +23,7 @@ package com.netflix.spinnaker.echo.model
 class Metadata {
     String source
     String type
+    String triggeredBy
     String created = new Date().time
     String organization
     String project
--- a/echo-model/src/main/groovy/com/netflix/spinnaker/echo/model/trigger/TriggerEvent.groovy
+++ b/echo-model/src/main/groovy/com/netflix/spinnaker/echo/model/trigger/TriggerEvent.groovy
@@ -20,4 +20,5 @@ import com.netflix.spinnaker.echo.model.Metadata
 
 abstract class TriggerEvent {
   Metadata details
+  Map payload
 }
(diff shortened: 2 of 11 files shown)


tomaslin:
I don't think the new endpoint is necessary. Events that are not trigger based are sent to the top level and the webhooks endpoint really serves as an identifier that the contents of the webhook is unknown and needs to be translated into an event format spinnaker understands.

I am not a fan of category because this means you have to add several webhooks for each type. The trigger for webhook should match on type, source and a list of attributes that are defined at trigger time ( I.e job, build, category, etc )

tomaslin:
To be clear, what I am saying is that instead of a static category field there should be a map of properties and values that match the trigger content. Only if all those attributes are matched should a new pipeline fire

netwolfuk:
Thanks for your comments. I have a couple of questions.

1. Do all matching triggers fire? If the Webhook trigger matches will other triggers get a chance to fire too?
2. Are you envisaging users are able to specify the properties map to match on in the deck UI?

What I was planning to do was to enhance the [WebhookEventMonitor#matchTriggerFor](https://github.com/netwolfuk/echo/blob/teamcity-trigger/echo-pipelinetriggers/src/main/java/com/netflix/spinnaker/echo/pipelinetriggers/monitor/WebhookEventMonitor.java#L99-L113) to check for matches in the payload, like application, pipeline name, etc. I just didn't get that running last night, and didn't want to go too far in case I was on the wrong track.

tomaslin:
all matching triggers fire

There does not need to be a strict mapping between what the UI does and how the triggering action happens. The goal here is to allow triggers to be created and added without needing to change much code in the backend. So you can have a teamcity trigger with fixed fields in the UI that will translate to the trigger structure expected in the backend. 

If you look at the Jenkins trigger, we fetch job names, masters and builds to generate dropdowns. I think it is a friendlier user experience if you don't require your user to be an expert in which properties they can set up.

netwolfuk:
I think I get it now. Sorry for being slow.

1. Create a nice trigger in Deck for each CI you want to support. I was realy hoping to avoid that, and just let people define the properties they want to capture.
1. That page creates the various properties and persists them against that trigger (I have this already working for the webhook stuff)
1. When the webhook request comes in to the WebhookController, it just sends the event to the propagator (again, you've already done this part).
1. I just need to change the WebhookEventListener so that it does try to check the source is "webhook". Rather just gets any trigger that matches the type and fetches the list of properties it's expecting.
1. Match those properties against the items in the payload (or content as it's referred to in your code) and if they all match, trigger the pipeline.

It would be nice to be able to pass some of the content in the payload into the pipeline. By that I mean set the buildNumber, define artifacts for bake, etc. How would that be accomplished? Can I just write into the trigger when it's created? Is that what TriggerMonitor#buildTrigger(Pipeline pipeline, TriggerEvent event) should be doing?

tomaslin:
for jenkins we load the additional details in orca - https://github.com/spinnaker/orca/blob/34dfc93848a41ec9c7c923c44b6b4b8f67ab54ab/orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/OperationsController.groovy#L116 

Instead of converting the event into a web hook event, the contents of the original event should be passed as the trigger field for web hooks it feels. Otherwise we seem to be losing information that we could use later

ajordens:
Seems like we were pretty closing to landing on a reasonable solution.

Where are we at now?

netwolfuk:
Last night I spent some time learning Spock because my tests were failing. Functionally this PR is working but the tests are failing.

netwolfuk:
Actually there is one last thing to fix. Webhook triggers are working but the bake step never rebakes. I'm not sure yet what I need to pass in the webhook to force a rebake.

netwolfuk:
I have rebased and created a new branch and PR #85 
I'll close this one.

PR #82, opened 2016-04-13
Externalize timezone for CRON evaluation as config property.

duftler:
- Default to America/Los_Angeles.
  - Respect timezone when determining if triggers and action instance are in sync.

### DIFF
--- a/echo-scheduler/src/main/groovy/com/netflix/spinnaker/echo/scheduler/actions/pipeline/impl/PipelineConfigsPollingAgent.groovy
+++ b/echo-scheduler/src/main/groovy/com/netflix/spinnaker/echo/scheduler/actions/pipeline/impl/PipelineConfigsPollingAgent.groovy
@@ -45,19 +45,22 @@ class PipelineConfigsPollingAgent extends AbstractPollingAgent {
   private final PipelineCache pipelineCache
   private final long intervalMs
   private final ActionsOperator actionsOperator
+  private final String timeZoneId
 
   @Autowired
   PipelineConfigsPollingAgent(CounterService counterService,
                               GaugeService gaugeService,
                               PipelineCache pipelineCache,
                               ActionsOperator actionsOperator,
-                              @Value('${scheduler.pipelineConfigsPoller.pollingIntervalMs:30000}') long intervalMs) {
+                              @Value('${scheduler.pipelineConfigsPoller.pollingIntervalMs:30000}') long intervalMs,
+                              @Value('${scheduler.cron.timezone:America/Los_Angeles}') String timeZoneId) {
     super()
     this.counterService = counterService
     this.gaugeService = gaugeService
     this.pipelineCache = pipelineCache
     this.actionsOperator = actionsOperator
     this.intervalMs = intervalMs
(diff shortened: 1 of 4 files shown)


duftler:
@cfieber please review.

echo-scheduler/src/main/groovy/com/netflix/spinnaker/echo/scheduler/actions/pipeline/impl/PipelineTriggerConverter.groovy
```diff
@@ -51,26 +52,27 @@ class PipelineTriggerConverter {
...
   }
 
-  static boolean isInSync(ActionInstance actionInstance, Trigger trigger) {
+  static boolean isInSync(ActionInstance actionInstance, Trigger trigger, String timeZoneId) {
     if (trigger.type == Trigger.Type.CRON.toString()) {
       return (
         actionInstance.trigger instanceof CronTrigger &&
-          trigger.cronExpression == ((CronTrigger) actionInstance.trigger).cronExpression
+          trigger.cronExpression == ((CronTrigger) actionInstance.trigger).cronExpression &&
+          timeZoneId == actionInstance?.context?.parameters?.triggerTimeZoneId
```
cfieber: I guess in the case of existing triggers where timeZoneId doesn't explicitly exist, they will just get updated here? (isInSync will be false the first time it is checked, which will re-persist the trigger including the configured timezone)

Also applies in the case of a persisted trigger with a timezone, and changing the system configured timezone?

echo-scheduler/src/main/groovy/com/netflix/spinnaker/echo/scheduler/actions/pipeline/impl/PipelineConfigsPollingAgent.groovy
```diff
@@ -46,6 +46,9 @@ class PipelineConfigsPollingAgent extends AbstractPollingAgent {
   private final long intervalMs
   private final ActionsOperator actionsOperator
 
+  @Value('${scheduler.cron.timezone:America/Los_Angeles}')
```
cfieber: Could you add a constructor arg and put the `@Value` annotation there, and make the field final

cfieber:
nice, thanks LGTM

echo-scheduler/src/main/groovy/com/netflix/spinnaker/echo/scheduler/actions/pipeline/impl/PipelineTriggerConverter.groovy
duftler: Yup, exactly.
I verified this behavior with a running system as well.

echo-scheduler/src/main/groovy/com/netflix/spinnaker/echo/scheduler/actions/pipeline/impl/PipelineConfigsPollingAgent.groovy
duftler: Done (sorry, hadn't spotted the existing `@Value` there).

PR #84, opened 2016-04-19
Add in-memory as an option

gregturn:
Make cassandra configuration conditional, but enabled by default for backwards compatibility.

### DIFF
--- a/echo-scheduler/src/main/groovy/com/netflix/spinnaker/echo/config/SchedulerConfiguration.groovy
+++ b/echo-scheduler/src/main/groovy/com/netflix/spinnaker/echo/config/SchedulerConfiguration.groovy
@@ -15,14 +15,16 @@
  */
 
 package com.netflix.spinnaker.echo.config
-
 import com.netflix.astyanax.Keyspace
 import com.netflix.fenzo.triggers.TriggerOperator
+import com.netflix.fenzo.triggers.persistence.InMemoryTriggerDao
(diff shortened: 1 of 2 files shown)


gregturn:
This introduces the option to switch off cassandra (critical for PCF) and switch to using the in-memory implementations.

@spinnaker/reviewers PTAL

echo-scheduler/src/main/groovy/com/netflix/spinnaker/echo/config/SchedulerConfiguration.groovy
```diff
@@ -44,21 +46,42 @@ import java.util.concurrent.TimeUnit
...
     }
 
     @Bean
+    @ConditionalOnExpression('${spinnaker.cassandra.enabled:true}')
     TriggerDao triggerDao(Keyspace keyspace) {
         new CassandraTriggerDao(keyspace)
     }
 
     @Bean
+    @ConditionalOnExpression('${spinnaker.in-memory.enabled:false}')
```
tomaslin: change to inMemory to be consistent with all other projects

tomaslin:
LGTM after changing property

gregturn:
Done.

tomaslin:
why do this instead of cassandra.embedded : true

gregturn:
Huh? This has nothing to do with Cassandra, embedded or not.

tomaslin:
the embedded cassandra functionality would work the same as the in-memory implementation of the scheduler

PR #86, opened 2016-05-23
triggers: fix webhook bug

lwander:
The webhook trigger added in https://github.com/spinnaker/echo/pull/85 never verified that the trigger type was correct, but only that the incoming trigger type matched that of an arbitrary registered trigger. The result is that with this code, an incoming jenkins trigger would kick off every jenkins triggered pipeline once, and the correct jenkins triggered pipeline a second time. I'm guessing this wasn't caught since that code was never released.

@spinnaker/reviewers PTAL


duftler:
LGTM.

PR #87, opened 2016-05-25
docker: Avoid duplicate triggers

lwander:
@duftler When a tag isn't specified, don't trigger on changes to "latest", since 1. specifying "latest" as the source tag is anti-pattern, and 2. this will cause duplicate builds when "latest" is a valid tag, and a change to separate tag was pushed (and expected).


duftler:
LGTM.
That method is now complicated enough that it would benefit from a test showing all the various combinations.

lwander:
Will tack on a unit test once we figure out the other trigger error.

PR #92, opened 2016-06-21
Pager Duty Support

ajordens:
- New INCIDENT Notification Type ... PagerDuty is just one possible implementation (payload should be generic enough to support alternative incident response systems)
- PRs in `orca` and `deck` are forthcoming

### DIFF
--- a/echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/pagerduty/PagerDutyNotificationService.groovy
+++ b/echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/pagerduty/PagerDutyNotificationService.groovy
@@ -1,0 +1,64 @@
+/*
+ * Copyright 2016 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+
+package com.netflix.spinnaker.echo.pagerduty
+
+import com.netflix.spinnaker.echo.api.Notification
+import com.netflix.spinnaker.echo.notification.NotificationService
+import com.netflix.spinnaker.echo.services.Front50Service
+import groovy.util.logging.Slf4j
+import org.springframework.beans.factory.annotation.Autowired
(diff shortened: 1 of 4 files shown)


ajordens:
PTAL @tomaslin

echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/api/Notification.groovy
```diff
@@ -29,13 +29,15 @@ class Notification {
...
+    String user
   }
 
   static enum Type {
     HIPCHAT,
     EMAIL,
     SMS,
-    SLACK
+    SLACK,
+    INCIDENT
```
tomaslin: PAGERDUTY instead of INCIDENT

echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/pagerduty/PagerDutyNotificationService.groovy
```diff
@@ -0,0 +1,67 @@
...
+
+  @Override
+  void handle(Notification notification) {
+    notification.to.each {
+      def application = front50Service.getApplication(it)
+      if (application.pdApiKey) {
+        pagerDuty.createEvent(
+          "Token token=${token}",
+          new PagerDutyService.PagerDutyCreateEvent(
+            service_key: application.pdApiKey as String,
```
tomaslin: Probably a better idea to keep the front50 lookup logic in orca so that a warning could be generated if pdApiKey is not set rather than fail silently

echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/api/Notification.groovy
ajordens: I'm not convinced we want this style of integration to require changes in `echo` + `orca` + `deck` ... which is why I purposely went generic with specific logic only existing at the edge in `echo`. 

I don't see any particular reason why `echo` shouldn't be able to surface failures back to callers ... maybe the `pdApiKey` doesn't exist, or perhaps it does exist but is otherwise invalid.

echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/pagerduty/PagerDutyNotificationService.groovy
ajordens: I'd rather throw an exception here and propagate it upwards.

ajordens:
Follow-up to have the `PagerDutyNotificationService` raise an exception if any applications it's trying to page don't have a `pdApiKey` (that's a terrible name as it's actually the pager duty service key).

In theory this supports paging multiple applications in one call but in reality it'll be singular.

ajordens:
Ended up going with a concrete `PAGER_DUTY` implementation. 

Can revisit if/when another pager integration comes along ... only a few lines of code either way between `orca` and `echo`.

PR #96, opened 2016-07-01
add connectionPool configuration that seems to soothe the OkHttpClient

cfieber:
the spring config around retrofit is a horrible horrible mess in this app due to the way it has
frankensteined together all the previous services, but this config change should only be
necessary for the connections to front50 and orca where it will get picked up

@spinnaker/netflix-reviewers FYI


ajordens:
LGTM

PR #102, opened 2016-10-10
allow custom message text for email/slack notification

robfletcher:


robfletcher:
Example of notification JSON: 

```{
  "notifications": [
    {
      "level": "pipeline",
      "when": [
        "pipeline.failed"
      ],
      "type": "email",
      "address": "delivery-engineering@netflix.com",
      "message": {
        "pipeline.failed": {
          "text": "O NOES!!"
        }
      }
    }
  ]
}```

File: echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/notification/EmailNotificationAgent.groovy [hunk 1]
```diff
@@ -16,90 +16,107 @@
...
-        sendMessage(
-            [preference.address] as String[],
-            event,
-            subject,
-            config.type,
-            status,
-            config.link
-        )
+  @Autowired
+  EmailNotificationService mailService
```
ajordens: PR party foul ... changing spaces

echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/notification/EmailNotificationAgent.groovy
```diff
@@ -16,90 +16,107 @@
...
+        executionId: event.content?.execution?.id,
+        type       : type,
+        status     : status,
+        link       : link,
+        name       : event.content?.execution?.name ?: event.content?.execution?.description
+      ]
+    )
+
+    if (customMessage) {
+      body += "\n\n" + customMessage
```
tomaslin: feels we want a better separator for this?

tomaslin:
why sms get no love?

echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/notification/EmailNotificationAgent.groovy
ajordens: Emails are full on plain text right now aren't they?

echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/notification/EmailNotificationAgent.groovy
```diff
@@ -16,90 +16,107 @@
...
     }
```
robfletcher: append custom message to body from velocity template

echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/notification/EmailNotificationAgent.groovy
tomaslin: well, if they go onto gmail it should be able to render fun characters

robfletcher: Could pass it to Velocity & then we can render it any way we like

ajordens:
LGTM (haven't tested the templates but seems reasonable given everything is plain text afaik)

echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/notification/EmailNotificationAgent.groovy
robfletcher: Updated PR

robfletcher:
@tomaslin I didn't do SMS because we could blow the character limit. If that's not a concern it would be easy to add.

PR #108, opened 2016-10-28
Take into account the composite action instance ids of InMemoryAction…

duftler:
…InstanceDao when making comparisons between action instance and trigger ids.

### DIFF
--- a/echo-scheduler/src/main/groovy/com/netflix/spinnaker/echo/scheduler/actions/pipeline/impl/PipelineConfigsPollingAgent.groovy
+++ b/echo-scheduler/src/main/groovy/com/netflix/spinnaker/echo/scheduler/actions/pipeline/impl/PipelineConfigsPollingAgent.groovy
@@ -122,7 +122,8 @@ class PipelineConfigsPollingAgent extends AbstractPollingAgent {
       }
 
       actionInstances.each { actionInstance ->
-        Trigger trigger = triggers.find { it.id == actionInstance.id }
+        // InMemoryActionInstanceDao builds composite action instance ids that end with the trigger id.
+        Trigger trigger = triggers.find { it.id == actionInstance.id || actionInstance.id.endsWith(":$it.id") }
         try {
           if (trigger) {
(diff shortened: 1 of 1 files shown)


duftler:
@lwander please review.
@robzienert fyi

echo-scheduler/src/main/groovy/com/netflix/spinnaker/echo/scheduler/actions/pipeline/impl/PipelineConfigsPollingAgent.groovy
```diff
@@ -176,7 +182,7 @@ class PipelineConfigsPollingAgent extends AbstractPollingAgent {
        */
       List<Pipeline> pipelinesWithNewTriggers = pipelines.findAll {
         def cronTriggers = it.triggers.findAll { it.type == Trigger.Type.CRON.toString() }
-        cronTriggers.any { !actionIds.contains(it.id) }
+        cronTriggers.any { !actionIds.contains(it.id) && !actionIds.find { actionId -> actionId.endsWith(":$it.id") } }
```
lwander: This "endsWith" check could probably be factored out. In fact, the whole id matching probably should be to avoid accidentally writing code that compares IDs incorrectly.

lwander:
LGTM

echo-scheduler/src/main/groovy/com/netflix/spinnaker/echo/scheduler/actions/pipeline/impl/PipelineConfigsPollingAgent.groovy
duftler: The composite ids end with a `:` followed by a uuid, so no real chance of a false positive. Feels like the right place to do some cleanup around this would be in scheduled-actions. The two instance action daos have different id naming schemes, so it might make sense to delegate the matching to those classes (or at lease expose the methods to generate the id's so we can use them here).
When we dig back into scheduled-actions to work on that cancellation-relation race condition, maybe we can do more cleanup around this.

PR #109, opened 2016-10-28
Adds 'runAsUser' field to Trigger with option to use this user as the…

ttomsu:
… X-SPINNAKER-USER call to Orca.

This is in support of Fiat's service accounts, and will enable the ability to run automated triggers against protected resources.

Additional PRs are coming to enforce Front50 enforcing checking that the user saving the pipeline has the access to the specified service account. This PR won't be submitted until that PR is ready and submitted.

@cfieber @tomaslin @jtk54 @duftler PTAL

tag: https://github.com/spinnaker/fiat/issues/36


jtk54:
LGTM, I'll defer for the other reviewers to merge.

PR #111, opened 2016-11-03
Conditionally include the 'runAsUser' iff Fiat is enabled

ttomsu:
@jtk54 @cfieber @ajordens @duftler PTAL

tag: https://github.com/spinnaker/fiat/issues/36


tomaslin:
LGTM

PR #113, opened 2016-11-11
Graceful handling of null triggers list

robzienert:
Fixes NPE where pipeline triggers list object is null.

cc @ajordens


ajordens:
LGTM

PR #124, opened 2017-01-05
Base echo halconfig

lwander:
@duftler @jtk54 PTAL

### DIFF
--- a/halconfig/echo.yml
+++ b/halconfig/echo.yml
@@ -1,0 +1,30 @@
+# halconfig
+
+server:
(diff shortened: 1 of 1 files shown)


halconfig/echo.yml
```diff
@@ -0,0 +1,35 @@
+# halconfig
+
+server:
+  port: ${services.echo.port:8089}
+  address: ${services.echo.host:localhost}
+
+cassandra:
```
duftler: Should probably just leave out the `cassandra` bits here.

PR #125, opened 2017-01-11
Adds runAsUser to CRON triggers

ttomsu:
@duftler PTAL

### DIFF
--- a/echo-scheduler/src/main/groovy/com/netflix/spinnaker/echo/scheduler/actions/pipeline/impl/PipelineTriggerConverter.groovy
+++ b/echo-scheduler/src/main/groovy/com/netflix/spinnaker/echo/scheduler/actions/pipeline/impl/PipelineTriggerConverter.groovy
@@ -37,7 +37,7 @@ class PipelineTriggerConverter {
     if (trigger.runAsUser) {
       params.runAsUser = trigger.runAsUser
     }
-    params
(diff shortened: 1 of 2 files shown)


echo-scheduler/src/main/groovy/com/netflix/spinnaker/echo/scheduler/actions/pipeline/impl/PipelineTriggerConverter.groovy
```diff
@@ -26,30 +26,37 @@ import com.netflix.spinnaker.echo.scheduler.actions.pipeline.PipelineTriggerActi
...
       triggerId            : trigger.id,
       triggerType          : trigger.type,
       triggerCronExpression: trigger.cronExpression,
       triggerTimeZoneId    : timeZoneId,
       triggerEnabled       : Boolean.toString(trigger.enabled)
     ]
+    if (trigger.runAsUser) {
+      params.runAsUser = trigger.runAsUser
+    }
+    params
```
duftler: Please add `return` here.

ttomsu: Done.

ttomsu:
Updated.

PR #126, opened 2017-01-12
Add optional basic auth support to outgoing rest webhooks.

duftler:


duftler:
@tomaslin please review.

PR #128, opened 2017-01-13
Support the ping events from github webhooks

gardleopard:
Github ships a ping request when you set up webhook, this results in an error using echo.
This PR addresses this issue.

https://developer.github.com/enterprise/2.8/webhooks/#ping-event

screen shot 2017-01-13 at 16 17 31

### DIFF
--- a/echo-webhooks/src/main/groovy/com/netflix/spinnaker/echo/controllers/WebhooksController.groovy
+++ b/echo-webhooks/src/main/groovy/com/netflix/spinnaker/echo/controllers/WebhooksController.groovy
@@ -50,10 +50,15 @@ class WebhooksController {
         }
       }
       if (source == 'github') {
-        event.content.hash = postedEvent.after
-        event.content.branch = postedEvent.ref.replace('refs/heads/', '')
-        event.content.repoProject = postedEvent.repository.owner.name
-        event.content.slug = postedEvent.repository.name
+        if (event.content.hook_id) {
+          log.info("Webook ping received from github hook_id=${event.content.hook_id} repository=${event.content.repository.full_name}")
+          sendEvent = false
+        } else {
+          event.content.hash = postedEvent.after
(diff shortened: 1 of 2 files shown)


echo-webhooks/src/main/groovy/com/netflix/spinnaker/echo/controllers/WebhooksController.groovy
```diff
@@ -50,10 +50,16 @@ class WebhooksController {
         }
       }
       if (source == 'github') {
-        event.content.hash = postedEvent.after
-        event.content.branch = postedEvent.ref.replace('refs/heads/', '')
-        event.content.repoProject = postedEvent.repository.owner.name
-        event.content.slug = postedEvent.repository.name
+        if (event.content.hook_id) {
+          log.debug("Webook ping received from github hook_id=${event.content.hook_id} repository=${event.content.repository.full_name}")
```
duftler: If this ping is only sent once when the webhook is first established, I think it would make sense to make this a `log.info` instead. Otherwise, one may get in a situation where they have to modify the default logger settings, and the remove/re-establish the webhook while troubleshooting.

echo-webhooks/src/main/groovy/com/netflix/spinnaker/echo/controllers/WebhooksController.groovy
```diff
@@ -50,10 +50,16 @@ class WebhooksController {
...
+          sendEvent = false
+        }
```
duftler: Please put the `else {` on the same line.

gardleopard:
Thanks @duftler, I have fixed the things you pointed out in the comments and I have tested this in our setup. It works.

duftler:
@tomaslin please review.

tomaslin:
LGTM

PR #131, opened 2017-01-18
Enables checking of 'X-Hub-Signature' in Github webhook requests to c…

ttomsu:
…onfirm source of webhook.

We don't currently authenticate the source of webhooks, so anyone on the internet could send a well-formed message to an unprotected Spinnaker instance. Github webhooks have an optional ability to include a signature in the HTTP headers, which this PR uses to confirm the authenticity of the sender before triggering the target pipeline.

The secret, unfortunately, is to be stored in the pipeline's trigger config. This is not optimal, but there is no Spinnaker key-server like HashiCorp's Vault. I think this is a fair tradeoff in order to close the unauthenticated webhook hole mentioned above.

PTAL @tomaslin @ajordens @duftler @jtk54 @lwander 

source: https://developer.github.com/webhooks/securing/
tag: https://github.com/spinnaker/fiat/issues/134 (Deck work still needs to be done to specify `secret`)


echo-pipelinetriggers/src/main/java/com/netflix/spinnaker/echo/pipelinetriggers/monitor/GitEventMonitor.java
```diff
@@ -94,7 +100,8 @@ protected boolean isSuccessfulTriggerEvent(TriggerEvent event) {
       && trigger.getSource().equalsIgnoreCase(source)
       && trigger.getProject().equalsIgnoreCase(project)
       && trigger.getSlug().equalsIgnoreCase(slug)
-      && (trigger.getBranch() == null || trigger.getBranch().equals("") || matchesPattern(branch, trigger.getBranch()));
+      && (trigger.getBranch() == null || trigger.getBranch().equals("") || matchesPattern(branch, trigger.getBranch()))
+      && hasValidGitHubSecureSignature(event, trigger);
```
tomaslin: This should just apply to github triggers. I believe this is also used by Stash triggers

ttomsu: A stash webhook call would not have the `X-Hub-Signature`, so `hasValidGitHubSecureSignature` would return true (line 128)

tomaslin:
LGTM

PR #143, opened 2017-04-26
feat(core): Allow configurable headers for rest endpoints

andrewbackes:
This feature started because I needed to be able to set the Authorization header in order to send events to Splunk.

This PR will allow you to set headers in two ways.

1. Directly in `echo-local.yml`:
```yaml
rest:
  enabled: true
  endpoints:
    - wrap: false
      url: http://www.example.com
      headers:
        Authorization: Splunk MYSECRETKEYHERE
```

2. In a file that contains the headers you want to include:
```yaml
rest:
  enabled: true
  endpoints:
    - wrap: false
      url: http://www.example.com
      headersFile: /mnt/mysecrets/splunkKey
```
where `/mnt/mysecrets/splunkKey` looks like:
```
Authorization: Splunk MYSECRETKEYHERE
```

In the examples above, I only used the `Authorization` header. However, you can specify any header you want. 

If there are conflicts with headers then the header in the `headersFile` will be used over what is in the `headers` field. If username/password fields are present in `echo-local.yml` and an Authorization header is specified, then the Authorization header will override the basic auth header created from the username/password.

### DIFF
--- a/echo-rest/src/main/groovy/com/netflix/spinnaker/echo/config/RestConfig.groovy
+++ b/echo-rest/src/main/groovy/com/netflix/spinnaker/echo/config/RestConfig.groovy
@@ -16,11 +16,10 @@
 
 package com.netflix.spinnaker.echo.config
 
-import static retrofit.Endpoints.newFixedEndpoint
-
-import org.apache.commons.codec.binary.Base64
 import com.netflix.spinnaker.echo.rest.RestService
 import groovy.transform.CompileStatic
+import groovy.util.logging.Slf4j
+import org.apache.commons.codec.binary.Base64
 import org.springframework.beans.factory.annotation.Value
 import org.springframework.beans.factory.config.ConfigurableBeanFactory
 import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty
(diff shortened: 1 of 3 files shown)


echo-rest/src/main/groovy/com/netflix/spinnaker/echo/config/RestConfig.groovy
```diff
@@ -68,16 +68,37 @@ class RestConfig {
...
+      }
+
+      if (endpoint.headers) {
+        headers += endpoint.headers
+      }
+
+      if (endpoint.headersFile) {
+        new File(endpoint.headersFile).eachLine { line ->
+          def pair = line.split(":")
+          if (pair.length == 2) {
```
duftler: Could use a warning here if the number of tokens is not 2.

echo-rest/src/main/groovy/com/netflix/spinnaker/echo/config/RestConfig.groovy
```diff
@@ -68,16 +68,37 @@ class RestConfig {
         .setLogLevel(retrofitLogLevel)
         .setConverter(new JacksonConverter())
 
+
```
duftler: Extraneous blank line.

duftler:
This looks nice.
Could use some unit tests, especially demonstrating the order of precedence.
LGTM once the unit tests are added and the comments are addressed.

andrewbackes:
Thanks again for the feedback @duftler, I added some unit tests to demonstrate the order of precedence for headers. I also address the formatting issues. Let me know what you think =)

PR #145, opened 2017-05-16
fix(triggers): fix NPE regression on BuildEventMonitor.

cfieber:
travis trigger check switched the comparison on trigger.getType to not be null safe

### DIFF
--- a/echo-pipelinetriggers/src/main/java/com/netflix/spinnaker/echo/pipelinetriggers/monitor/BuildEventMonitor.java
+++ b/echo-pipelinetriggers/src/main/java/com/netflix/spinnaker/echo/pipelinetriggers/monitor/BuildEventMonitor.java
@@ -112,6 +112,6 @@ public class BuildEventMonitor extends TriggerMonitor {
   }
 
   private boolean isBuildTrigger(Trigger trigger) {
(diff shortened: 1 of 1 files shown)


jeyrschabu:
LGTM

echo-pipelinetriggers/src/main/java/com/netflix/spinnaker/echo/pipelinetriggers/monitor/BuildEventMonitor.java
```diff
@@ -112,6 +112,6 @@ protected void onMatchingPipeline(Pipeline pipeline) {
   }
 
   private boolean isBuildTrigger(Trigger trigger) {
-    return Arrays.stream(BUILD_TRIGGER_TYPES).anyMatch(triggerType -> trigger.getType().equals(triggerType));
```
cfieber: This was previously implemented [here](https://github.com/spinnaker/echo/commit/30cd0f7b22bedd35dd7998f639692033804f86aa#diff-e8626a7b50789fd70f554891c4d3af21L88) as
```
(BUILD_TRIGGER_TYPE.equals(trigger.getType())
```
which handles the case of `trigger.getType()` returning null

PR #146, opened 2017-05-26
feat(rest): extensible JSON payloads for Spinnaker events

imosquera:
When configuring webhooks in Echo, it only supports two types of payloads:
```
{ "content":[..], "details":[..]}
```
or using the `wrap` property which forces the payload to have a field called `eventName`.   This PR introduces a new config property called `template` which allows a user to specify a simple string template.  Currently the only string that will be interpolated is `${event}`.  Below is an example of a rest endpoint configured with a template.

```
rest:
  endpoints:
    - wrap: true
      url: https://events.spinnaker.myteam.net/
      template: '{"myCustomEventFieldName":${event}}'
```

Note that the template must be valid JSON _after_ the interpolation.  The template will contain the event and the entire content details will be placed in `${event}` such that a rendered template will look like:

```
"myCustomEventFieldName": {
    "details": { ... },
    "content": { ... }
}
```

### DIFF
--- a/echo-rest/src/main/groovy/com/netflix/spinnaker/echo/events/RestEventListener.groovy
+++ b/echo-rest/src/main/groovy/com/netflix/spinnaker/echo/events/RestEventListener.groovy
@@ -39,6 +39,9 @@ class RestEventListener implements EchoEventListener {
   @Autowired
   RestUrls restUrls
 
+  @Autowired
+  RestEventTemplateEngine restEventTemplateEngine
+
   @Value('${rest.defaultEventName:spinnaker_events}')
   String eventName
 
@@ -51,15 +54,21 @@ class RestEventListener implements EchoEventListener {
       try {
         Map eventAsMap = mapper.convertValue(event, Map)
         Map sentEvent = eventAsMap
+
         if (service.config.flatten) {
           eventAsMap.content = mapper.writeValueAsString(eventAsMap.content)
           eventAsMap.details = mapper.writeValueAsString(eventAsMap.details)
         }
+
         if (service.config.wrap) {
-          sentEvent = [
-            "eventName": "${service.config.eventName ?: eventName}" as String,
-          ]
-          sentEvent["${service.config.fieldName ?: fieldName}" as String] = eventAsMap
+          if (service.config.template) {
+            sentEvent = restEventTemplateEngine.render(service.config.template as String, sentEvent)
+          } else {
+            sentEvent = [
+              eventName: "${service.config.eventName ?: eventName}" as String,
+            ]
+            sentEvent["${service.config.fieldName ?: fieldName}" as String] = eventAsMap
+          }
         }
         service.client.recordEvent(sentEvent)
       } catch (e) {
--- a/echo-rest/src/main/groovy/com/netflix/spinnaker/echo/events/RestEventTemplateEngine.groovy
+++ b/echo-rest/src/main/groovy/com/netflix/spinnaker/echo/events/RestEventTemplateEngine.groovy
@@ -1,0 +1,38 @@
+/*
+ * Copyright 2017 Armory, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License")
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.spinnaker.echo.events
+
+import groovy.util.logging.Slf4j
+import org.springframework.stereotype.Component
+import com.fasterxml.jackson.databind.ObjectMapper
+
+interface RestEventTemplateEngine {
+    Map render(String template, Map eventMap)
+}
+
+
+@Component
+@Slf4j
+class SimpleEventTemplateEngine implements RestEventTemplateEngine {
+
+    ObjectMapper objectMapper = new ObjectMapper()
+
+    Map render(String templateString, Map eventMap) {
+        String renderedResult = templateString.replace('${event}', objectMapper.writeValueAsString(eventMap))
+        return objectMapper.readValue(renderedResult, Map)
+    }
+}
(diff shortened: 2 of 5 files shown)


echo-rest/src/main/groovy/com/netflix/spinnaker/echo/events/RestEventListener.groovy
```diff
@@ -51,15 +54,20 @@ class RestEventListener implements EchoEventListener {
...
         if (service.config.wrap) {
-          sentEvent = [
-            "eventName": "${service.config.eventName ?: eventName}" as String,
-          ]
-          sentEvent["${service.config.fieldName ?: fieldName}" as String] = eventAsMap
+          if(service.config.template) {
+            sentEvent = restEventTemplateEngine.render(service.config.template as String, sentEvent)
+          } else {
+            sentEvent = [
+                    "eventName": "${service.config.eventName ?: eventName}" as String,
```
lwander: Indention seems a 4 spaces too much

echo-rest/src/main/groovy/com/netflix/spinnaker/echo/events/RestEventListener.groovy
```diff
@@ -51,15 +54,20 @@ class RestEventListener implements EchoEventListener {
...
         if (service.config.flatten) {
           eventAsMap.content = mapper.writeValueAsString(eventAsMap.content)
           eventAsMap.details = mapper.writeValueAsString(eventAsMap.details)
         }
         if (service.config.wrap) {
-          sentEvent = [
-            "eventName": "${service.config.eventName ?: eventName}" as String,
-          ]
-          sentEvent["${service.config.fieldName ?: fieldName}" as String] = eventAsMap
+          if(service.config.template) {
```
lwander: Missing space

echo-rest/src/main/groovy/com/netflix/spinnaker/echo/events/RestEventTemplateEngine.groovy
```diff
@@ -0,0 +1,38 @@
...
+
+
+@Component
+@Slf4j
+class SimpleEventTemplateEngine implements RestEventTemplateEngine {
+
+    ObjectMapper objectMapper = new ObjectMapper()
+
+    Map render(String templateString, Map eventMap) {
+        String renderedResult = templateString.replace('\"${event}\"', objectMapper.writeValueAsString(eventMap))
```
lwander: Why include the `\"` here? Is it to ensure that the pre-transformed template is valid JSON? The reason I ask is that it might be a pain to make backwards-compatible if we start to flesh out this functionality with a more advanced templating solution.

echo-rest/src/main/groovy/com/netflix/spinnaker/echo/events/RestEventListener.groovy
```diff
@@ -51,15 +54,21 @@ class RestEventListener implements EchoEventListener {
...
         if (service.config.wrap) {
-          sentEvent = [
-            "eventName": "${service.config.eventName ?: eventName}" as String,
-          ]
-          sentEvent["${service.config.fieldName ?: fieldName}" as String] = eventAsMap
+          if (service.config.template) {
+            sentEvent = restEventTemplateEngine.render(service.config.template as String, sentEvent)
+          } else {
+            sentEvent = [
+              "eventName": "${service.config.eventName ?: eventName}" as String,
```
duftler: Don't need the quotes around the `eventName` key.

echo-rest/src/test/groovy/com/netflix/spinnaker/echo/events/RestEventListenerSpec.groovy
```diff
@@ -34,18 +34,40 @@ class RestEventListenerSpec extends Specification {
...
+            ]
+        ]
+    ]
+
+    when:
+    listener.processEvent(event)
+
+    then:
+    1 * restService.recordEvent({
+      it.myCustomEventField == listener.mapper.convertValue(event, Map)
```
duftler: Would rather see the assertion compare the value of `myCustomEventField` with a literal map, instead of relying on `listener.mapper.convertValue(event, Map)`.

echo-rest/src/test/groovy/com/netflix/spinnaker/echo/events/RestEventListenerSpec.groovy
```diff
@@ -34,18 +34,40 @@ class RestEventListenerSpec extends Specification {
...
+    })
+  }
+
   void 'wraps events when wrap is set'() {
```
duftler: Was all the re-formatting here intentional? I'd think just making the new test case indented at 2 spaces and leaving the others at 2 is a shorter path.

echo-rest/src/test/groovy/com/netflix/spinnaker/echo/events/SimpleEventTemplateSpec.groovy
```diff
@@ -0,0 +1,47 @@
...
+package com.netflix.spinnaker.echo.events
+
+import spock.lang.Specification
+import spock.lang.Subject
+
+class SimpleEventTemplateSpec extends Specification {
+
+    @Subject
+    RestEventTemplateEngine restEventTemplateEngine = new SimpleEventTemplateEngine()
+    Map event = [
```
duftler: Prefer 2 spaces of indentation over 4.

echo-rest/src/test/groovy/com/netflix/spinnaker/echo/events/SimpleEventTemplateSpec.groovy
```diff
@@ -0,0 +1,47 @@
...
+              'action': 'resize asg'
+            ],
+            'details': [
+              'user': 'anonymous'
+            ]
+    ]
+
+    void 'renders template with event map'() {
+        given:
+        def template = '{"myEventField":"${event}"}'
```
duftler: I think you need to drop the quotes around `${event}` to get the unit test to pass.

echo-rest/src/test/groovy/com/netflix/spinnaker/echo/events/RestEventListenerSpec.groovy
imosquera: It was not intentional.  I'll go ahead and update.

echo-rest/src/test/groovy/com/netflix/spinnaker/echo/events/RestEventListenerSpec.groovy
duftler: ah, good point. Let's leave it as-is then.

PR #148, opened 2017-06-13
feat(rest): introduces a new configuration option,`insecure` which creates a https client which is insecure by ignoring host verification and does not validate certificate chains

imosquera:
(rest): This PR introduces a new configuration option,`insecure` which accepts a boolean value and creates a https client which is insecure by ignoring host verification (e.g. always return true for hosts) and does not validate certificate chains.

This is need due to Splunk's management of certificates which doesn't always use a common CA.  Additionally, this is useful for testing purposes.  Below is how you would configure an insecure client

```
rest:
  enabled: true
  endpoints:
    - wrap: true
      url: https://myinstance.splunkcloud.com/services/collector/event?
      headers:
        Authorization: Splunk AAAAAAA-AAA-AAAA-AAAA-AAAAB
      insecure: true
```

### DIFF
--- a/echo-rest/src/main/groovy/com/netflix/spinnaker/echo/events/RestClientFactory.groovy
+++ b/echo-rest/src/main/groovy/com/netflix/spinnaker/echo/events/RestClientFactory.groovy
@@ -1,0 +1,75 @@
+/*
+ * Copyright 2017 Armory, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License")
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.spinnaker.echo.events
+
+import com.squareup.okhttp.OkHttpClient
+import org.springframework.beans.factory.annotation.Autowired
+import org.springframework.stereotype.Component
+import retrofit.client.Client
+import retrofit.client.OkClient
+
+import javax.net.ssl.HostnameVerifier
+import javax.net.ssl.SSLContext
+import javax.net.ssl.SSLSocketFactory
+import javax.net.ssl.TrustManager
+import javax.net.ssl.X509TrustManager
+
+
+interface OkHttpClientFactory {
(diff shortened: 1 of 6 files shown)


echo-rest/src/main/groovy/com/netflix/spinnaker/echo/events/RestClientFactory.groovy
```diff
@@ -0,0 +1,75 @@
...
+}
+
+@Component
+class RestClientFactory {
+
+  @Autowired
+  OkHttpClientFactory httpClientFactory
+
+  Client getClient(Boolean insecure) {
+    if (insecure == false) {
```
lwander:
```groovy
if (insecure) {
  return new OkClient(httpClientFactory.getInsecureClient())
} else {
  return new OkClient()
}
```

reads a little better IMO

lwander:
Please `git commit --amend` to match changelog requirements: https://www.spinnaker.io/community/contributing/submitting/

echo-rest/src/test/groovy/com/netflix/spinnaker/echo/events/OkHttpClientFactorySpec.groovy
```diff
@@ -0,0 +1,27 @@
...
+
+import javax.net.ssl.HostnameVerifier
+
+class OkHttpClientFactorySpec extends Specification {
+
+  @Subject
+  OkHttpClientFactory clientFactory = new OkHttpClientFactoryImpl()
+  OkHttpClient insecureClient
+
+  void setup() {
```
duftler: Can drop this method if it has no body.

echo-rest/src/test/groovy/com/netflix/spinnaker/echo/events/OkHttpClientFactorySpec.groovy
```diff
@@ -0,0 +1,27 @@
...
+  }
+
+  void 'insecure client does not verify hostname'() {
+
+    when:
+    insecureClient = clientFactory.getInsecureClient()
+
+    then:
+    insecureClient.getHostnameVerifier().verify("mockdomain", null) == true
```
duftler: Is the `== true` necessary?

echo-rest/src/main/groovy/com/netflix/spinnaker/echo/events/RestClientFactory.groovy
imosquera: Agreed.

echo-rest/src/test/groovy/com/netflix/spinnaker/echo/events/OkHttpClientFactorySpec.groovy
imosquera: It's not.  I'll go ahead and remove

PR #152, opened 2017-06-24
refactor(logs): Structured logs

robzienert:
Kicking around the idea of more structured logs for our services so things may be easier to index in search. This converts the entirety of echo (from what I saw as worthwhile) as a POC before biting off bigger services.

Uses https://github.com/logstash/logstash-logback-encoder

@spinnaker/reviewers WDYT? If it's interesting, I'll update spinnaker-dependencies.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -36,7 +36,7 @@ allprojects {
     group = "com.netflix.spinnaker.echo"
 
     ext {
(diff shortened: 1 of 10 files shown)


cfieber:
makes a ton of sense to me, and I particularly like the possibility of using the appender directly to publish instead of running the agent sidecar

PR #162, opened 2017-08-22
feat(templates): removes deprecated velocity engine in favour of freemarker

cfieber:
existing templates were auto converted using cavalry tool [0] then hand tweaked after for null handling

[0] http://doc.justonetech.com/freemarker-2.3.10/docs/usCavalry.html


echo-web/config/echo.yml
```diff
@@ -17,3 +17,5 @@ search:
   index: 'event_history'
 
 management.health.mail.enabled: false
+
+spring.freemarker.enabled: false
```
cfieber: this disables freemarker as a web view source (the `Configuration` object still gets instantiated)

PR #166, opened 2017-08-24
fix(triggers): fix handling of blank regex in docker registry triggers

polarbizzle:
When you remove a previously existing regex from the trigger in the
UI, deck keeps a tag field with an empty string for the value in the
json. This arrives as non null to the DockerEventMonitor, which  causes
the trigger to not fire for any new tag until you manually update the
pipeline json to remove the field. This change makes the logic more
resilient to malformed input. Blank or empty tag regex doesn't make
sense so it should behave the same as when the regex is null.

### DIFF
--- a/echo-pipelinetriggers/src/main/java/com/netflix/spinnaker/echo/pipelinetriggers/monitor/DockerEventMonitor.java
+++ b/echo-pipelinetriggers/src/main/java/com/netflix/spinnaker/echo/pipelinetriggers/monitor/DockerEventMonitor.java
@@ -26,6 +26,7 @@ import com.netflix.spinnaker.echo.model.trigger.TriggerEvent;
 import com.netflix.spinnaker.echo.pipelinetriggers.PipelineCache;
 import lombok.NonNull;
 import lombok.val;
+import org.apache.commons.lang.StringUtils;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Component;
 import rx.Observable;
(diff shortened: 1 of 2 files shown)


echo-pipelinetriggers/src/main/java/com/netflix/spinnaker/echo/pipelinetriggers/monitor/DockerEventMonitor.java
```diff
@@ -103,12 +103,19 @@ private boolean matchTags(String suppliedTag, String incomingTag) {
     DockerEvent dockerEvent = (DockerEvent) event;
     String account = dockerEvent.getContent().getAccount();
     String repository = dockerEvent.getContent().getRepository();
-    String tag = dockerEvent.getContent().getTag();
     return trigger -> trigger.getType().equals(TRIGGER_TYPE) &&
-      trigger.getRepository().equals(repository) &&
-      trigger.getAccount().equals(account) &&
-      ((trigger.getTag() == null && !tag.equals("latest"))
-        || trigger.getTag() != null && matchTags(trigger.getTag(), tag));
+              trigger.getRepository().equals(repository) &&
```
tomaslin: should just be `return isMatchingTrigger(trigger)` and put all the logic in that method

polarbizzle: I just pushed a commit that I think does what you're asking for. Had to drag along the DockerEvent also, though.

echo-pipelinetriggers/src/main/java/com/netflix/spinnaker/echo/pipelinetriggers/monitor/DockerEventMonitor.java
```diff
@@ -100,15 +100,22 @@ private boolean matchTags(String suppliedTag, String incomingTag) {
...
     String repository = dockerEvent.getContent().getRepository();
-    String tag = dockerEvent.getContent().getTag();
-    return trigger -> trigger.getType().equals(TRIGGER_TYPE) &&
-      trigger.getRepository().equals(repository) &&
-      trigger.getAccount().equals(account) &&
-      ((trigger.getTag() == null && !tag.equals("latest"))
-        || trigger.getTag() != null && matchTags(trigger.getTag(), tag));
+    String eventTag = dockerEvent.getContent().getTag();
+    String triggerTagPattern = null;
+    if(trigger.getTag() != null && trigger.getTag().trim().length() > 0) {
```
jtk54: You can replace this if clause with `StringUtils.isEmpty()` from apache.lang. Also please add a space before the opening `(`.

ngaya-ll: Actually this would be equivalent to `!StringUtils.isBlank()`.

jtk54: SGTM.

polarbizzle:
fixed and squashed, thanks for the review 👍

jtk54:
Thanks for the contribution.

PR #167, opened 2017-08-24
feat(pubsub): Process message events and forward matching triggers.

jtk54:
Adds specs for a crucial classes in this code path. Also refactors the 10+ arg `Trigger` constructors into `builder`s.

### DIFF
--- a/echo-model/src/main/java/com/netflix/spinnaker/echo/model/Trigger.java
+++ b/echo-model/src/main/java/com/netflix/spinnaker/echo/model/Trigger.java
@@ -22,12 +22,13 @@ import lombok.Builder;
 import lombok.ToString;
 import lombok.Value;
 import lombok.experimental.Wither;
+
 import java.util.Map;
 
 @JsonDeserialize(builder = Trigger.TriggerBuilder.class)
-@Builder
+@Builder(toBuilder = true)
 @Wither
-@ToString(of = {"type", "master", "job", "cronExpression", "source", "project", "slug", "account", "repository", "tag", "constraints", "branch", "runAsUser"}, includeFieldNames = false)
+@ToString(of = {"type", "master", "job", "cronExpression", "source", "project", "slug", "account", "repository", "tag", "constraints", "branch", "runAsUser", "subscriptionName", "pubsubType"}, includeFieldNames = false)
 @Value
 public class Trigger {
   public enum Type {
@@ -35,7 +36,8 @@ public class Trigger {
     GIT("git"),
     JENKINS("jenkins"),
     DOCKER("docker"),
-    WEBHOOK("webhook");
+    WEBHOOK("webhook"),
+    PUBSUB("pubsub");
 
     private final String type;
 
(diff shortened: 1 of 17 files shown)


echo-model/src/main/java/com/netflix/spinnaker/echo/model/Trigger.java
```diff
@@ -69,29 +71,80 @@ public String toString() {
   String branch;
   String runAsUser;
   String secret;
+  String subscriptionName;
+  String pubsubType;
 
   public Trigger atBuildNumber(final int buildNumber) {
-    return new Trigger(enabled, id, type, master, job, buildNumber, propertyFile, cronExpression, source, project, slug, null, account, repository, null, digest, null, branch, runAsUser, secret);
+    return this.toBuilder()
```
lwander: Noice

echo-pipelinetriggers/src/main/java/com/netflix/spinnaker/echo/pipelinetriggers/monitor/PubsubEventMonitor.java
```diff
@@ -0,0 +1,125 @@
...
+
+  @Override
+  protected Predicate<Trigger> matchTriggerFor(final TriggerEvent event) {
+    PubsubEvent pubsubEvent = (PubsubEvent) event;
+    MessageDescription description = pubsubEvent.getContent().getMessageDescription();
+
+    // TODO(jacobkiefer): Need to apply filters specified in triggers here.
+    return trigger -> trigger.getType().equalsIgnoreCase(PUBSUB_TRIGGER_TYPE)
+        && trigger.getPubsubType().equalsIgnoreCase(description.getPubsubType().toString())
+        && trigger.getSubscriptionName().equalsIgnoreCase(description.getSubscriptionName());
```
lwander: If you have multiple files in one GCS bucket, do each have their own subscription?

echo-pipelinetriggers/src/main/java/com/netflix/spinnaker/echo/pipelinetriggers/monitor/PubsubEventMonitor.java
```diff
@@ -0,0 +1,125 @@
...
+  }
+
+  @Override
+  protected void onMatchingPipeline(Pipeline pipeline) {
+    super.onMatchingPipeline(pipeline);
+    val id = registry.createId("pipelines.triggered")
```
lwander: Why is this happening outside the `isPubsubTrigger` guard below? Will this be applied to all triggered pipelines, not just pubsub ones?

echo-pipelinetriggers/src/main/java/com/netflix/spinnaker/echo/pipelinetriggers/monitor/PubsubEventMonitor.java
```diff
@@ -0,0 +1,125 @@
...
+      id.withTag("pubsubType", pipeline.getTrigger().getPubsubType());
+      id.withTag("subscriptionName", pipeline.getTrigger().getSubscriptionName());
+    }
+
+    registry.counter(id).increment();
+  }
+
+  private boolean isPubsubTrigger(Trigger trigger) {
+    return PUBSUB_TRIGGER_TYPE.equals(trigger.getType())
+        && (trigger.getSubscriptionName() != null && !trigger.getSubscriptionName().isEmpty())
```
lwander: StringUtils.isEmpty from apache.lang3 I think

echo-pipelinetriggers/src/main/java/com/netflix/spinnaker/echo/pipelinetriggers/monitor/PubsubEventMonitor.java
jtk54: No, there's one subscription per bucket in the GCS -> Pubsub setup. This is where artifact filters in pipeline triggers come in so you don't trigger on every single change to a bucket. This is obviously not included in this PR -- hopefully arriving shortly.

echo-pipelinetriggers/src/main/java/com/netflix/spinnaker/echo/pipelinetriggers/monitor/PubsubEventMonitor.java
jtk54: This only applies to a pipeline with a trigger matching `matchTriggerFor` for a pubsub event. This is more or less just an extra guard so we don't try to dereference a field that is null in the Trigger.

echo-pipelinetriggers/src/main/java/com/netflix/spinnaker/echo/pipelinetriggers/monitor/PubsubEventMonitor.java
jtk54: Right on.

PR #168, opened 2017-08-31
feat(pubsub): Parse artifacts from pub/sub message contents.

jtk54:
Adds proper artifact model parsing from pub/sub messages and propagates artifacts to Orca. The spec gives an idea what jinja is capable of, but we'll have to document some of the cookie-cutter use cases.

### DIFF
--- a/echo-pubsub/src/main/java/com/netflix/spinnaker/echo/pubsub/utils/MessageArtifactTranslator.java
+++ b/echo-pubsub/src/main/java/com/netflix/spinnaker/echo/pubsub/utils/MessageArtifactTranslator.java
@@ -1,0 +1,87 @@
+/*
+ * Copyright 2017 Google, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.spinnaker.echo.pubsub.utils;
+
+import com.fasterxml.jackson.core.type.TypeReference;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.hubspot.jinjava.Jinjava;
+import com.netflix.spinnaker.kork.artifacts.model.GenericArtifact;
+import lombok.NoArgsConstructor;
+import lombok.extern.slf4j.Slf4j;
+import org.apache.commons.lang.StringUtils;
+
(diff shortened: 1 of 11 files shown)


echo-pubsub/src/main/java/com/netflix/spinnaker/echo/pubsub/utils/MessageArtifactTranslator.java
```diff
@@ -0,0 +1,90 @@
...
+        jinjaTemplate = new String(Files.readAllBytes(Paths.get(templatePath)));
+      } catch (IOException ioe) {
+        log.error(ioe.getMessage());
+      }
+    }
+  }
+
+  public List<GenericArtifact> parseArtifacts(String messagePayload) {
+    if (!StringUtils.isNotBlank(messagePayload)) {
+      log.warn("Attempted to parse artifacts from empty message.");
```
lwander: Does this need a warning? It seems like an acceptable use of the triggers.

echo-pubsub/src/main/java/com/netflix/spinnaker/echo/pubsub/utils/MessageArtifactTranslator.java
```diff
@@ -0,0 +1,90 @@
...
+    String derp = jinja.render(jinjaTemplate, context);
+    return jinja.render(jinjaTemplate, context);
+  }
+
+  private Map readMapValue(ObjectMapper mapper, String messagePayload) {
+    Map context = null;
+    try {
+      context = mapper.readValue(messagePayload, Map.class);
+    } catch (IOException ioe) {
+      log.error(ioe.getMessage());
```
lwander: Why not fail instead?

echo-pubsub/src/test/groovy/com/netflix/spinnaker/echo/pubsub/utils/MessageArtifactTranslatorSpec.groovy
```diff
@@ -0,0 +1,106 @@
...
+    def artifacts = messageArtifactTranslator.parseArtifacts(payload)
+
+    then:
+    artifacts.size() == 1
+    artifacts.any { it.getReference() == 'gs://this/is/my/id' && it.getType() == 'gcs/object' }
+  }
+
+  def "can deserialize several artifacts from a list with jinja magic"() {
+    String template = '''
+    [
```
lwander: Very neat :)

echo-pubsub/src/main/java/com/netflix/spinnaker/echo/pubsub/utils/MessageArtifactTranslator.java
```diff
@@ -0,0 +1,90 @@
...
+    }
+    return context;
+  }
+
+  private List<GenericArtifact> readArtifactList(ObjectMapper mapper, String hydratedTemplate) {
+    try {
+      return mapper.readValue(hydratedTemplate, artifactListReference);
+    } catch (IOException ioe) {
+      log.error(ioe.getMessage());
```
lwander: Again this might be a good time to fail

PR #171, opened 2017-09-13
feat(pubsub): Enforce expected artifacts in triggers.

jtk54:
Triggers can specify expected artifacts. This change ensures that at least one of the artifacts parsed from the message satisfies each expected artifact specified in the Trigger.


jtk54:
Ping

jtk54:
I'm going to add some tests before I merge this, on @gardleopard's suggestion.

PR #176, opened 2017-10-02
feat(dryrun) capability for triggering dry run executions in prestaging

robfletcher:
The idea is that a successful execution of certain pipelines in production will trigger a "dry-run" execution in another environment that can confirm the same behavior of stage topology, expression evaluation, etc. using a newer Orca.

See https://github.com/spinnaker/orca/pull/1662 for corresponding Orca PR


echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/config/DryRunConfig.java
```diff
@@ -0,0 +1,55 @@
...
+import retrofit.RestAdapter;
+import retrofit.client.OkClient;
+import static retrofit.Endpoints.newFixedEndpoint;
+
+@Configuration
+@ConditionalOnProperty("dryRun.enabled:false")
+public class DryRunConfig {
+
+  @Bean
+  Endpoint dryRunEndpoint(@Value("dryRun.url") String url) {
```
robzienert: `dryRun.front50Url` seems a better option.

robfletcher:
The idea is that the echo notification can trigger a run in another env. Definitely don’t want it to run in prod.

Regards,
Rob Fletcher

PR #177, opened 2017-10-04
fix(pipelinetriggers): Add retry on network & http errors

robzienert:
@spinnaker/reviewers

### DIFF
--- a/echo-pipelinetriggers/src/main/java/com/netflix/spinnaker/echo/pipelinetriggers/orca/PipelineInitiator.java
+++ b/echo-pipelinetriggers/src/main/java/com/netflix/spinnaker/echo/pipelinetriggers/orca/PipelineInitiator.java
@@ -7,10 +7,14 @@ import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.beans.factory.annotation.Value;
 import org.springframework.boot.actuate.metrics.CounterService;
 import org.springframework.stereotype.Component;
+import retrofit.RetrofitError;
+import retrofit.RetrofitError.Kind;
 import rx.Observable;
 import rx.functions.Action1;
+import rx.functions.Func1;
 
 import javax.annotation.PostConstruct;
+import java.util.concurrent.TimeUnit;
 
 /**
  * Triggers a {@link Pipeline} by invoking _Orca_.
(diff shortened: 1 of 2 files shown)


echo-pipelinetriggers/src/main/java/com/netflix/spinnaker/echo/pipelinetriggers/orca/PipelineInitiator.java
```diff
@@ -48,29 +50,42 @@ public void call(Pipeline pipeline) {
...
-      if (fiatEnabled && runAsUser != null && !runAsUser.isEmpty()) {
-        response = orca.trigger(pipeline, pipeline.getTrigger().getRunAsUser());
-      } else {
-        response = orca.trigger(pipeline);
-      }
-      response.subscribe(this::onOrcaResponse, this::onOrcaError);
+      createTriggerObservable(pipeline)
+        .retry((count, error) -> {
+          if (count <= 2 && isRetryable(error)) {
+            log.error("Retrying pipeline trigger, attempt {}/3: {}", count + 1, pipeline);
```
jtk54: Planning on making the number of retries configurable?

robzienert: Yes good thought.

jtk54: LGTM otherwise, not sure if you're gating on me or waiting for ajordens :D

ajordens: Drowning in PR emails so I missed this one.

ajordens: General comment about the variety of ways we're doing retrys throughout Spinnaker, some with configurable retries, some with exponential back-off, some in loops, etc.

Specific to this, maybe some backoff and a higher default retry count. It'd probably be fine if we retried for upwards of 30s?

ajordens:
General comment about the variety of ways we're doing retrys throughout Spinnaker, some with configurable retries, some with exponential back-off, some in loops, etc. 

Specific to this, maybe some backoff and a higher default retry count. It'd probably be fine if we retried for upwards of 30s?

Otherwise, if you're happy with his, :shipit:.

robzienert:
Would probably be worthwhile to consolidate some common retry logic into kork. Will add it as a personal todo.

PR #190, opened 2017-10-25
feat(pubsub): Support firing triggers from configured subscriptions.

jtk54:


echo-pubsub/src/main/java/com/netflix/spinnaker/echo/pubsub/google/GooglePubsubSubscriber.java
```diff
@@ -88,7 +88,7 @@ public static GooglePubsubSubscriber buildSubscriber(GooglePubsubProperties.Goog
     String jsonPath = subscription.getJsonPath();
 
     GooglePubsubMessageReceiver messageReceiver = new GooglePubsubMessageReceiver(subscription.getAckDeadlineSeconds(),
-        formatSubscriptionName(project, subscriptionName),
+        subscription.getName(),
```
lwander: I thought `name` was the user-assigned field that's only used for readability like in the rest of Spinnaker

jtk54: It is, except that we need to feed the `name` into the pub/sub trigger matching logic since that's what is configured in the UI.

lwander:
LGTM

PR #194, opened 2017-11-01
feat(pubsub): Trigger pipeline matches on expectedArtifact IDs.

jtk54:


jtk54:
Depends on https://github.com/spinnaker/deck/pull/4349 and https://github.com/spinnaker/orca/pull/1763.

echo-model/src/main/java/com/netflix/spinnaker/echo/model/Trigger.java
```diff
@@ -80,7 +80,7 @@ public String toString() {
    */
   String subscriptionName;
   String pubsubSystem;
-  List<ExpectedArtifact> expectedArtifacts;
+  List<String> expectedArtifacts;
```
lwander: Should we call these `expectedArtifactIds`?

jtk54: Yeah, let me rename.

jtk54: Changed this everywhere and re-tested, should be ok now.

PR #195, opened 2017-11-13
feat(pagerduty): Support details field and add error handling/reporting

jrsquared:
Related orca PR: https://github.com/spinnaker/orca/pull/1797

### DIFF
--- a/echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/pagerduty/PagerDutyNotificationService.groovy
+++ b/echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/pagerduty/PagerDutyNotificationService.groovy
@@ -17,14 +17,21 @@
 
 package com.netflix.spinnaker.echo.pagerduty
 
+import com.fasterxml.jackson.databind.ObjectMapper
 import com.netflix.spinnaker.echo.api.Notification
 import com.netflix.spinnaker.echo.notification.NotificationService
 import com.netflix.spinnaker.echo.services.Front50Service
+import groovy.transform.InheritConstructors
 import groovy.util.logging.Slf4j
 import org.springframework.beans.factory.annotation.Autowired
 import org.springframework.beans.factory.annotation.Value
 import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty
+import org.springframework.http.HttpStatus
 import org.springframework.stereotype.Component
+import org.springframework.web.bind.annotation.ResponseStatus
+import retrofit.RetrofitError
+import retrofit.mime.TypedByteArray
+import retrofit.mime.TypedInput
 
 import static net.logstash.logback.argument.StructuredArguments.kv
 
@@ -33,6 +40,9 @@ import static net.logstash.logback.argument.StructuredArguments.kv
 @ConditionalOnProperty('pagerDuty.enabled')
 class PagerDutyNotificationService implements NotificationService {
   private static Notification.Type TYPE = Notification.Type.PAGER_DUTY
+
+  @Autowired
+  ObjectMapper mapper
 
   @Autowired
   PagerDutyService pagerDuty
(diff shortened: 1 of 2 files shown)


echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/pagerduty/PagerDutyNotificationService.groovy
```diff
@@ -34,6 +40,8 @@ import static net.logstash.logback.argument.StructuredArguments.kv
 class PagerDutyNotificationService implements NotificationService {
   private static Notification.Type TYPE = Notification.Type.PAGER_DUTY
 
+  private final ObjectMapper mapper = new ObjectMapper()
```
ajordens: Usually we'll autowire an `ObjectMapper`

echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/pagerduty/PagerDutyNotificationService.groovy
```diff
@@ -50,17 +58,91 @@ class PagerDutyNotificationService implements NotificationService {
 
   @Override
   void handle(Notification notification) {
+    Map details = new LinkedHashMap()
```
ajordens: Mr. Groovy has left you a voice mail ... it says:

`def details = [details: notification.additionalContext.details]`

If you care about LinkedHashMap you should be able to `as LinkedHashMap`

echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/pagerduty/PagerDutyNotificationService.groovy
```diff
@@ -50,17 +58,91 @@ class PagerDutyNotificationService implements NotificationService {
...
+            service_key: it,
+            client: "Spinnaker (${notification.source.user})",
+            description: notification.additionalContext.message,
+            details: details
+          )
         )
-      )
 
-      log.info('Sent page {} {}', kv('serviceKey', it), kv('message', notification.additionalContext.message))
+        if (response.status.equals("success")) {
```
ajordens: On the off chance there's a null in here ... you could:

`"success".equals(response?.status)`

echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/pagerduty/PagerDutyNotificationService.groovy
```diff
@@ -50,17 +58,91 @@ class PagerDutyNotificationService implements NotificationService {
...
-        "Token token=${token}",
-        new PagerDutyService.PagerDutyCreateEvent(
-          service_key: it,
-          client: "Spinnaker (${notification.source.user})",
-          description: notification.additionalContext.message
+      try {
+        Map response = pagerDuty.createEvent(
+          "Token token=${token}",
+          new PagerDutyService.PagerDutyCreateEvent(
+            service_key: it,
```
ajordens: Rather than using `it` ... it might be more readable to `notification.to.each { String serviceKey ->`

I realize we were using `it` before but you've added some more code (and error handling) and it's likely worth it being a tad more readable.

echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/pagerduty/PagerDutyNotificationService.groovy
```diff
@@ -50,17 +58,91 @@ class PagerDutyNotificationService implements NotificationService {
...
+        log.error('Failed to send page {} {} {}',  kv('serviceKey', it), kv('message', notification.additionalContext.message), kv('error', errorMessage))
+        errors.put(it, errorMessage)
+      }
+    }
+
+    // If some errors occurred
+    if (errors.size() > 0 || pdErrors.size() > 0) {
+      String message = ""
+      String pagerDutyError = pdErrors.collect { "${it.key}: ${it.value}" }.join(", ")
+      if (pagerDutyError != null && !pagerDutyError.isEmpty()) {
```
ajordens: can be `if (pagerDutyErrors)` and that'll null _and_ empty check

echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/pagerduty/PagerDutyNotificationService.groovy
```diff
@@ -50,17 +58,91 @@ class PagerDutyNotificationService implements NotificationService {
...
+
     notification.to.each {
-      pagerDuty.createEvent(
-        "Token token=${token}",
-        new PagerDutyService.PagerDutyCreateEvent(
-          service_key: it,
-          client: "Spinnaker (${notification.source.user})",
-          description: notification.additionalContext.message
+      try {
+        Map response = pagerDuty.createEvent(
```
ajordens: Value in typing the response from `pagerDuty.createEvent()` to something better than `Map`?

echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/pagerduty/PagerDutyNotificationService.groovy
```diff
@@ -50,17 +58,91 @@ class PagerDutyNotificationService implements NotificationService {
...
+          // Page successful
+          log.info('Sent page {} {}', kv('serviceKey', it), kv('message', notification.additionalContext.message))
+        } else {
+          pdErrors.put(it, (String)response.message)
+        }
+      } catch (RetrofitError error) {
+        String errorMessage = error.response.reason
+        TypedInput responseBody = error.response.getBody()
+        if (responseBody != null) {
+          PagerDutyErrorResponseBody errorResponse = mapper.readValue(new String(((TypedByteArray)responseBody).getBytes()), PagerDutyErrorResponseBody)
```
ajordens: IMO try and keep to 120 line width (RobF does 80 but that's _crazy_) ...

```
def errorResponse = mapper.readValue(
  new String(....),
  PagerDutyErrorResponseBody
)
```

is more readable IMO.

echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/pagerduty/PagerDutyNotificationService.groovy
```diff
@@ -50,17 +58,91 @@ class PagerDutyNotificationService implements NotificationService {
...
+            errorMessage = errorResponse.errors.join(", ")
+          }
+        }
+        log.error('Failed to send page {} {} {}',  kv('serviceKey', it), kv('message', notification.additionalContext.message), kv('error', errorMessage))
+        errors.put(it, errorMessage)
+      }
+    }
+
+    // If some errors occurred
+    if (errors.size() > 0 || pdErrors.size() > 0) {
```
ajordens: `if (errors || pdErrors)`

echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/pagerduty/PagerDutyNotificationService.groovy
```diff
@@ -50,17 +58,91 @@ class PagerDutyNotificationService implements NotificationService {
...
+      if (pagerDutyError != null && !pagerDutyError.isEmpty()) {
+        message += " ${pagerDutyError}."
+      }
+      String requestError = errors.collect { "${it.key}: ${it.value}" }.join(", ")
+      if (requestError != null && !requestError.isEmpty()) {
+        message += " ${requestError}."
+      }
+      // If some successes occurred
+      if (errors.size() + pdErrors.size() < notification.to.size()) {
+        throw new PartialPagerDutySuccessException(message)
```
ajordens: I suspect that normally we would actually just throw one type of exception and pass in the full message.

It's a tad strange to pass in `message` but have it actually only form part of the exception message.

jrsquared: Yeah, good call. Was separate for different properties, but rolled it all into message.

PR #197, opened 2017-11-19
feat(webhooks): Support artifacts

lwander:
The result of this is that if someone POSTs to `$GATE_URL/webhooks/artifacts/<source>` with either

1. `"artifacts": []` specified in the payload, or
2. an artifact transforming jinja template registered for `<source>`.

The artifacts will be injected into any pipelines that declare that webhook (and have matching expected artifacts) as a trigger.

This is useful, because with (1) external systems can either explicitly say which artifacts they want to trigger pipelines with, or, any sort of "push" based trigger system with (2) can have their output converted into something Spinnaker understands without us having to supply code for it.

For example, look at the message format in [quay's notification guide](https://docs.quay.io/guides/notifications.html).

A user (or really us, for common cases like this) can supply a jinja template like:

```jinja
[
{% for tag in updated_tags %}
{
  "type": "docker/image",
  "name": "{{ docker_url }}"
  "reference": "{{ docker_url }}:{{ tag }}"
  "version": "{{ tag }}"
} {% if not loop.last %} , {% endif %}
{% endfor %}
]
```

for the source `quay`, and then configure quay to send push notifications to `$GATE_URL/webhooks/artifact/quay`.

```yaml
# echo.yml
webhooks:
  artifacts:
    enabled: true
    sources:
    - source: quay
      templatePath: /path/to/above/jinja-file # in the future some might ship with spinnaker
```

In their pipeline, they would register this "expected artifact": 
quay

and add a "webhook" trigger that requires that artifact to be present. Now only changes to images matching that tag will be allowed to trigger their pipeline, and any stage that understands how to deploy a docker image will be able to do so.

### DIFF
--- a/echo-pubsub/src/main/java/com/netflix/spinnaker/echo/pubsub/google/GooglePubsubSubscriber.java
+++ b/echo-pubsub/src/main/java/com/netflix/spinnaker/echo/pubsub/google/GooglePubsubSubscriber.java
@@ -26,13 +26,14 @@ import com.google.cloud.pubsub.v1.Subscriber;
 import com.google.common.util.concurrent.MoreExecutors;
 import com.google.pubsub.v1.PubsubMessage;
 import com.google.pubsub.v1.SubscriptionName;
+import com.netflix.spinnaker.echo.artifacts.MessageArtifactTranslator;
 import com.netflix.spinnaker.echo.config.google.GooglePubsubProperties;
 import com.netflix.spinnaker.echo.model.pubsub.MessageDescription;
 import com.netflix.spinnaker.echo.model.pubsub.PubsubSystem;
 import com.netflix.spinnaker.echo.pubsub.PubsubMessageHandler;
 import com.netflix.spinnaker.echo.pubsub.model.PubsubSubscriber;
-import com.netflix.spinnaker.echo.pubsub.utils.MessageArtifactTranslator;
 import com.netflix.spinnaker.echo.pubsub.utils.NodeIdentity;
+import com.netflix.spinnaker.kork.artifacts.model.Artifact;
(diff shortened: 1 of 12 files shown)


echo-pubsub/src/main/java/com/netflix/spinnaker/echo/pubsub/google/GooglePubsubSubscriber.java
```diff
@@ -143,17 +146,26 @@ public GooglePubsubMessageReceiver(Integer ackDeadlineSeconds,
...
           .ackDeadlineMillis(5 * TimeUnit.SECONDS.toMillis(ackDeadlineSeconds)) // Set a high upper bound on message processing time.
           .retentionDeadlineMillis(TimeUnit.DAYS.toMillis(7)) // Expire key after max retention time, which is 7 days.
-          .artifacts(messageArtifactTranslator.parseArtifacts(messagePayload))
           .build();
       GoogleMessageAcknowledger acknowledger = new GoogleMessageAcknowledger(consumer);
+
+      try {
+        List<Artifact> artifacts = messageArtifactTranslator.parseArtifacts(messagePayload);
+        description.setArtifacts(artifacts);
+        log.info("artifacts {}", String.join(", ", artifacts.stream().map(Artifact::toString).collect(Collectors.toList())));
```
lwander: FYI @jtk54 this should fix that bug where a malformed jinja template causes echo to retry processing a message forever

lwander:
Does anyone know what happened to this? https://github.com/spinnaker/deck/pull/2214

lwander:
https://github.com/spinnaker/spinnaker/issues/2159

PR #201, opened 2017-11-28
fix(pubsub): Use message IDs instead of payload hash.

jtk54:


echo-pubsub/src/main/java/com/netflix/spinnaker/echo/pubsub/google/GooglePubsubSubscriber.java
```diff
@@ -146,6 +146,7 @@ public GooglePubsubMessageReceiver(Integer ackDeadlineSeconds,
     @Override
     public void receiveMessage(PubsubMessage message, AckReplyConsumer consumer) {
       String messagePayload = message.getData().toStringUtf8();
+      String messageId = message.getMessageId();
```
lwander: Globally unique?

jtk54: > Guaranteed to be unique within the topic.

From the javadoc.

PR #202, opened 2017-11-29
feat(pubsub): support constraints

lwander:


echo-pipelinetriggers/src/main/java/com/netflix/spinnaker/echo/pipelinetriggers/artifacts/ArtifactMatcher.java
```diff
@@ -48,4 +49,26 @@ public static Boolean anyArtifactsMatchExpected(List<Artifact> messageArtifacts,
...
+   * @param payload A map of the payload contents POST'd in the Webhook.
+   * @return Whether every key (and value if applicable) in the constraints map is represented in the payload.
+   */
+  public static boolean isConstraintInPayload(final Map constraints, final Map payload) {
+    for (Object key : constraints.keySet()) {
+      if (!payload.containsKey(key) || payload.get(key) == null) {
+        log.info("Webhook trigger ignored. Item " + key.toString() + " was not found in payload");
+        return false;
+      }
+      if (!constraints.get(key).equals("") && (!constraints.get(key).equals(payload.get(key)))){
```
jtk54: Do we want this to `equals` or just `match`?

lwander: Mmm match is a good idea

PR #207, opened 2017-11-30
Adding ability for custom webhook triggers to pass parameters.

TrevinTeacutter:
We wanted to add the ability for echo webhook triggers to pass on parameters if they are in the webhook payload. We are currently using this endpoint to prevent customization of automagically generated pipelines by using this trigger and tying a service account through RunAsuser that only the thing that can automagically generate pipelines has access to, we do this so that custom pipelines can be created but without impacting generated ones.

It is a bit of a mess but we couldn't really think of a reason why we shouldn't try and contribute this back up as it may be useful for other webhooks potentially.

### DIFF
--- a/echo-model/src/main/java/com/netflix/spinnaker/echo/model/Trigger.java
+++ b/echo-model/src/main/java/com/netflix/spinnaker/echo/model/Trigger.java
@@ -65,6 +65,7 @@ public class Trigger {
   String project;
   String slug;
   String hash;
+  Map parameters;
   String account;
   String repository;
   String tag;
@@ -127,11 +128,12 @@ public class Trigger {
         .build();
   }
 
-  public Trigger atConstraints(final Map constraints) {
+  public Trigger atConstraints(final Map parameters, final Map constraints) {
     return this.toBuilder()
         .buildNumber(null)
         .hash(null)
         .digest(null)
+        .parameters(parameters)
         .constraints(constraints)
         .subscriptionName(null)
         .pubsubSystem(null)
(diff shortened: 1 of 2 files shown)


echo-pipelinetriggers/src/main/java/com/netflix/spinnaker/echo/pipelinetriggers/artifacts/ArtifactMatcher.java
```diff
@@ -32,10 +33,9 @@
 public class ArtifactMatcher {
   public static Boolean anyArtifactsMatchExpected(List<Artifact> messageArtifacts, Trigger trigger, Pipeline pipeline) {
     List<String> expectedArtifactIds = trigger.getExpectedArtifactIds();
-    List<ExpectedArtifact> expectedArtifacts = pipeline.getExpectedArtifacts()
-        .stream()
-        .filter(e -> expectedArtifactIds.contains(e.getId()))
-        .collect(Collectors.toList());
+    List<ExpectedArtifact> expectedArtifacts = pipeline.getExpectedArtifacts() == null ? new ArrayList<>() : pipeline.getExpectedArtifacts();
```
TrevinTeacutter: While testing this we ran into a weird issue where a pipeline was returning null for expectedArtifacts so implemented a naive way around null since wasn't too sure what would be the expected response in that case.

TrevinTeacutter:
@spinnaker/reviewers could someone take a look at this?

jtk54:
Thanks for the contribution, sorry for the delay.

PR #212, opened 2017-12-12
fix(webhooks) Webhook trigger events should be treated as immutable.

TrevinTeacutter:
This should fix an issue we ran into with our prior contribution: https://github.com/spinnaker/echo/pull/207

Since events aren't filtered for each pipeline by the point the trigger is built for a pipeline, only the first pipeline processed would have the parameters while the rest would never have them (because we were using remove and not just get).


jtk54:
LG2M2.

PR #213, opened 2017-12-13
feat(triggers):(DO NOT MERGE) Adding ability to skip triggering pipelines on holidays

jeyrschabu:
- a pipeline trigger with `skipHolidays: true` should prevent that pipeline from being triggered on a defined holiday

### DIFF
--- a/echo-model/src/main/java/com/netflix/spinnaker/echo/model/Trigger.java
+++ b/echo-model/src/main/java/com/netflix/spinnaker/echo/model/Trigger.java
@@ -29,7 +29,7 @@ import lombok.experimental.Wither;
 @JsonDeserialize(builder = Trigger.TriggerBuilder.class)
 @Builder(toBuilder = true)
 @Wither
-@ToString(of = {"type", "master", "job", "cronExpression", "source", "project", "slug", "account", "repository", "tag", "parameters", "constraints", "branch", "runAsUser", "subscriptionName", "pubsubSystem", "expectedArtifactIds"}, includeFieldNames = false)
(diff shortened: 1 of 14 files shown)


echo-pipelinetriggers/src/test/groovy/com/netflix/spinnaker/echo/pipelinetriggers/monitor/SkipHolidaysTriggerCriteriaSpec.groovy
```diff
@@ -0,0 +1,48 @@
+/*
```
jeyrschabu: This test is currently incomplete

jeyrschabu:
@cfieber what do you think? Open to alternatives.

jeyrschabu:
Supports basic & google calendar. Basic calendar can be constructed as below:
```
pipelineTriggers:
  calendar:
    timezone: America/Los_Angeles
    google:
      enabled: false
      calendarId: en.usa#holiday@group.v.calendar.google.com
      credentials: ...
    
    basic:
      enabled: true
      holidaySpecs:
        - name: Chrismas eve
          day: 24
          month: 12

        - name: Chrismas
          day: 25
          month: 12

        - name: New Year's Eve
          day: 31
          month: 12
```

robzienert:
@jeyrschabu Like the idea. What's the fate?

jeyrschabu:
@robzienert this is still a thing. Has been on hold for a bit but needs to be revisited

jeyrschabu:
Going a different direction.

PR #214, opened 2017-12-14
fix(artifacts/github): Github artifacts should be type file

andrewbackes:
@lwander - Right now when GitHub webhooks to Echo, the artifacts aren't extracted. I think that artifact type should be `file`. However, I'm not exactly sure under what circumstances type `git` would be used, so could be wrong.


andrewbackes:
nvm.

PR #215, opened 2017-12-15
fix(webhooks): fix disabled webhook startup

lwander:


andrewbackes:
Looks good, thanks for fixing this up.

kevinawoo:
lgtm!

PR #218, opened 2017-12-20
fix(pubsub): Add attribute constraints to triggers.

jtk54:


jtk54:
UI is here: https://github.com/spinnaker/deck/pull/4599.

jtk54:
I'm going to change the field name from `constraints` to `payloadConstraints`, but don't merge until I add a pipeline migrator to Front50.

echo-model/src/main/java/com/netflix/spinnaker/echo/model/pubsub/MessageDescription.java
```diff
@@ -37,6 +38,13 @@
 
   private String messagePayload;
 
+  /**
+   * Optional, additional message metadata sent from the pub/sub system.
+   *
+   * May be null.
+   */
+  private Map<String, Object> messageAttributes;
```
lwander: Given that payload is a string, should we leave this as a string as well? Seems it will be more kafka-friendly when that time comes (that uses an 8 bit flag object)

jtk54:
Here's the migrator: https://github.com/spinnaker/front50/pull/299.

PR #219, opened 2017-12-29
fix(triggers): Include image as a received artifact for Docker

wjoel:
We prefer small, well tested pull requests.

Please refer to [Contributing to Spinnaker](https://spinnaker.io/community/contributing/).

When filling out a pull request, please consider the following:

* Follow the commit message conventions [found here](http://www.spinnaker.io/v1.0/docs/how-to-submit-a-patch).
* Provide a descriptive summary for your changes.
* If it fixes a bug or resolves a feature request, be sure to link to that issue.
* Add inline code comments to changes that might not be obvious.
* Squash your commits as you keep adding changes.
* Add a comment to @spinnaker/reviewers for review if your issue has been outstanding for more than 3 days.

Note that we are unlikely to accept pull requests that add features without prior discussion. The best way to propose a feature is to open an issue first and discuss your ideas there before implementing them.


echo-pipelinetriggers/src/main/java/com/netflix/spinnaker/echo/pipelinetriggers/monitor/DockerEventMonitor.java
```diff
@@ -77,7 +80,19 @@ protected boolean isSuccessfulTriggerEvent(final TriggerEvent event) {
   @Override
   protected Function<Trigger, Pipeline> buildTrigger(Pipeline pipeline, TriggerEvent event) {
     DockerEvent dockerEvent = (DockerEvent) event;
-    return trigger -> pipeline.withTrigger(trigger.atTag(dockerEvent.getContent().getTag()));
+    DockerEvent.Content content = dockerEvent.getContent();
+
+    String name = content.getRegistry() + "/" + content.getRepository();
+    String reference = name + ":" + content.getTag();
```
jtk54: Are `registry`, `repository` and `tag` always present on a DockerEvent?

echo-pipelinetriggers/src/test/groovy/com/netflix/spinnaker/echo/test/RetrofitStubs.groovy
```diff
@@ -75,7 +75,7 @@ trait RetrofitStubs {
       tag = inTag
     }
     def res = new DockerEvent()
-    res.content = new DockerEvent.Content("registry", "repository", tag, "sha")
+    res.content = new DockerEvent.Content("registry", "registry", "repository", tag, "sha")
```
jtk54: What was `registry` bound to before? If it was already present, do we need the `DockerEvent.Content` model change?

wjoel: It was, and still is, bound to `account` (so using the string "registry" as its value is arguably confusing).

echo-pipelinetriggers/src/main/java/com/netflix/spinnaker/echo/pipelinetriggers/monitor/DockerEventMonitor.java
```diff
@@ -77,7 +80,19 @@ protected boolean isSuccessfulTriggerEvent(final TriggerEvent event) {
...
+    List<Artifact> artifacts = Collections.singletonList(Artifact.builder()
+      .type("docker/image")
+      .name(name)
+      .version(content.getTag())
+      .location(reference)
```
lwander: I don't think "location" has any special meaning for docker images (yet). I would drop this call here.

echo-pipelinetriggers/src/main/java/com/netflix/spinnaker/echo/pipelinetriggers/monitor/DockerEventMonitor.java
wjoel: Yes, but to confirm this I went looking for where the events are coming from, which lead me to [this code in Igor](https://github.com/spinnaker/igor/blob/f2f0e3c4c6964fcb369af6132b5fc8ef30c147ad/igor-web/src/main/groovy/com/netflix/spinnaker/igor/docker/DockerMonitor.groovy#L134). It already includes artifacts, so we could use that; however, the artifacts are of the type `"docker"`, while the Kubernetes v2 provider is expecting `"docker/image"`.

They need to be the same (unless we change the type here, but that would be terribly ugly and confusing). `"docker"` is slightly older, while `"docker/image"` is more consistent with the naming other artifact types and has been used in the [new documentation](https://www.spinnaker.io/reference/artifacts/in-kubernetes-v2/).

@lwander, @gardleopard, any thoughts on this?

wjoel: As far as I can tell, the artifact in the event is quite different (of type `GenericArtifact` from Igor) from what we expect. Perhaps Igor should be refactored to use kork? Echo already uses `com.netflix.spinnaker.kork.artifacts.model.Artifact` in a number of places. Anyway, that would be a different commit.

I've updated the tests to use `"account"` for the account.

echo-pipelinetriggers/src/main/java/com/netflix/spinnaker/echo/pipelinetriggers/monitor/DockerEventMonitor.java
wjoel: Done.

wjoel:
@lwander I've added the `anyArtifactsMatchExpected(getArtifacts(dockerEvent), trigger, pipeline)` check and ran a test on my cluster, still works as expected so it should be good.

PR #221, opened 2018-01-17
fix(pubsub): Add payload to pubsub triggers.

jtk54:
Also fixes mismatch in trigger field/value in webhook monitor.

### DIFF
--- a/echo-model/src/main/java/com/netflix/spinnaker/echo/model/Trigger.java
+++ b/echo-model/src/main/java/com/netflix/spinnaker/echo/model/Trigger.java
@@ -29,7 +29,7 @@ import java.util.Map;
 @JsonDeserialize(builder = Trigger.TriggerBuilder.class)
 @Builder(toBuilder = true)
 @Wither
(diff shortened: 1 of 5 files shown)


echo-model/src/main/java/com/netflix/spinnaker/echo/model/Trigger.java
```diff
@@ -133,16 +134,21 @@ public Trigger atTag(final String tag) {
         .build();
   }
 
-  public Trigger atPayloadConstraints(final Map parameters, final Map payloadConstraints) {
+  public Trigger atPayload(final Map payload) {
+    return this.toBuilder()
+      .buildNumber(null)
```
lwander: Shouldn't these only set the necessary fields? This might delete stuff especially when these are chained together.

jtk54: I initially preserved the `null`s because the first at* functions had them, but I agree. I'll see where we can reasonably get rid of these.

jtk54:
Cleaned up the erroneous nulls in the builders, I think the initial ones were relevant to tags/branches/buildNumbers, so I left those in.

PR #231, opened 2018-02-12
feat(pubsub): adding support for AWS SQS/SNS

emjburns:
A first pass at adding pubsub support for AWS.
- Throwing custom exception types instead of `RuntimeExceptions`.
- Allowing you to have no artifacts in your message (`MessageArtifactTranslator` will no longer throw a runtime exception if an artifact cannot be parsed from the message. The assumption is that your pipeline will be looking for something in that original message that's not an artifact).
- SNS topic must be created ahead of time, and must allow SQS subscriptions from the account that Spinnaker is running in.
- If the queue you've specified as a subscriber doesn't exist, Spinnaker will create it. If there's no value in specifying the queue name I can instead auto-create a queue with a name like `...echo-$ENV-topic_name`, or make queue arn an optional override for the echo-generated name. 

Config in echo looks like
```
pubsub:
  amazon:
    enabled: true
    subscriptions:
    - name: my_subscription
      topicARN: arn:aws:sns:us-west-2:100:my_topic
      queueARN: arn:aws:sqs:us-west-2:100:my_queue
      # templatePath: /Users/emburns/.spinnaker/emburns-jinja.json # Optional artifact template, overrides messageFormat, if specified as anything other than "custom"
      # messageFormat: s3 # Optional, specify messageFormat of s3 to use the built in s3 jinfa artifact template
```

@lwander a small fix to `DockerHubArtifactExtractor` because of upgrading to a newer spinnaker-dependencies version


echo-core/src/main/groovy/com/netflix/spinnaker/echo/artifacts/MessageArtifactTranslator.java
```diff
@@ -87,7 +88,11 @@ private Map readMapValue(ObjectMapper mapper, String messagePayload) {
     try {
       return mapper.readValue(hydratedTemplate, artifactListReference);
     } catch (IOException ioe) {
-      throw new RuntimeException(ioe);
+      // Failure to parse artifacts from the message indicates either
+      // the message payload does not match the provided template or
+      // there is no template and no artifacts are expected
+      log.error("Unable to parse artifact from {}", hydratedTemplate);
```
emjburns: Allows messages with no artifacts in them to pass through for all subscription types.

echo-pubsub/src/main/java/com/netflix/spinnaker/echo/config/amazon/AmazonPubsubProperties.java
```diff
@@ -0,0 +1,119 @@
...
+  @Valid
+  private List<AmazonPubsubSubscription> subscriptions;
+
+  @Data
+  public static class AmazonPubsubSubscription {
+
+    private static final Logger log = LoggerFactory.getLogger(AmazonPubsubSubscription.class);
+
+    @NotEmpty
+    String name;
```
jtk54: These fields can be private.

echo-pubsub/src/main/java/com/netflix/spinnaker/echo/pubsub/amazon/ARN.java
```diff
@@ -0,0 +1,48 @@
...
+
+import java.util.Collection;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+public class ARN {
+
+  static final Pattern PATTERN = Pattern.compile("arn:aws(?:-cn|-us-gov)?:.*:(.*):(\\d+):(.*)");
+
+  String arn;
```
jtk54: Access modifiers on fields here.

echo-pubsub/src/main/java/com/netflix/spinnaker/echo/pubsub/amazon/SQSSubscriber.java
```diff
@@ -0,0 +1,255 @@
...
+    while (true) {
+      try {
+        listenForMessages();
+      } catch (Throwable e) {
+        log.error("Unexpected error running " + getWorkerName() + ", restarting", e);
+      }
+    }
+  }
+
+  private void listenForMessages() {
```
jtk54: Could break up the logical chunks into helpers to make this function more readable.

echo-pubsub/src/main/java/com/netflix/spinnaker/echo/pubsub/amazon/SQSSubscriber.java
```diff
@@ -0,0 +1,255 @@
...
+    this.queueId = ensureQueueExists(
+      amazonSQS, queueARN, topicARN, subscription.getSqsMessageRetentionPeriodSeconds()
+    );
+    subscribeToTopic(amazonSNS, topicARN, queueARN);
+
+    while (true) {
```
jtk54: I think I initially envisioned the pub/sub polling parts/agent looking similar to [Igor's polling parts](https://github.com/spinnaker/igor/blob/279f2f951747acb19eea2d4ae60896267dec9e02/igor-web/src/main/groovy/com/netflix/spinnaker/igor/jenkins/JenkinsBuildMonitor.groovy). Looking back at the fact that Google pub/sub is async and that SQS has its own polling flavor, it seems like that pattern is not a great fit. I still think the interface for integrating to pub/sub systems should be more uniform (i.e. if we add 5 more integrations where there is no standard interface, the code becomes hard to maintain). I think I'm going to take another look at this when I get around to implementing the Kafka integration since that will also be polling to make it more uniform. Thoughts?

echo-pubsub/src/main/java/com/netflix/spinnaker/echo/pubsub/amazon/NotificationMessageWrapper.java
```diff
@@ -0,0 +1,49 @@
...
+
+package com.netflix.spinnaker.echo.pubsub.amazon;
+
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+@JsonIgnoreProperties(ignoreUnknown = true)
+public class NotificationMessageWrapper {
+  @JsonProperty("Type")
+  String type;
```
jtk54: Access modifiers.

lwander:
Are you planning on adding halyard support too? most of the stuff is in place for google: https://www.spinnaker.io/reference/halyard/commands/#hal-config-pubsub-google-subscription-add

echo-pubsub/src/main/java/com/netflix/spinnaker/echo/pubsub/amazon/SQSSubscriber.java
emjburns: We use SQS and SNS for some bits in `clouddriver`, and I took most of this logic/flow from https://github.com/spinnaker/clouddriver/blob/master/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/lifecycle/InstanceTerminationLifecycleWorker.java. This is quite well tested and was easy to reason about, which is why I followed that pattern.

I agree that it would be great to have a better definition for polling implementations since aws isn't async. If you have suggestions now I'm super happy to refactor. I did struggle a bit to get aws to conform to the async model, but overall I think it's not too bad.

emjburns:
@lwander I'll add halyard support after I've rolled this out and tested it widely.

echo-pubsub/src/main/java/com/netflix/spinnaker/echo/config/amazon/AmazonPubsubProperties.java
```diff
@@ -0,0 +1,119 @@
...
+    MessageFormat messageFormat;
+
+    int visibilityTimeout = 30;
+    int sqsMessageRetentionPeriodSeconds = 120;
+    int waitTimeSeconds = 5;
+
+    public AmazonPubsubSubscription() {
+    }
+
+    public AmazonPubsubSubscription(
```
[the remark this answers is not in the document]
emjburns: I find Lombok more opaque than just generating the constructors.

echo-pubsub/src/main/java/com/netflix/spinnaker/echo/pubsub/amazon/ARN.java
```diff
@@ -0,0 +1,53 @@
...
+  static final Pattern PATTERN = Pattern.compile("arn:aws(?:-cn|-us-gov)?:.*:(.*):(\\d+):(.*)");
+
+  private String arn;
+  private String region;
+  private String name;
+
+  ARN(String arn) {
+    this.arn = arn;
+
+    Matcher sqsMatcher = PATTERN.matcher(arn);
```
ajordens: `s/sqs/arn`

echo-pubsub/src/main/java/com/netflix/spinnaker/echo/pubsub/amazon/SQSSubscriber.java
```diff
@@ -0,0 +1,267 @@
...
+        log.error("Unexpected error running " + getWorkerName() + ", restarting", e);
+      }
+    }
+  }
+
+  private void listenForMessages() {
+    this.queueId = ensureQueueExists(
+      amazonSQS, queueARN, topicARN, subscription.getSqsMessageRetentionPeriodSeconds()
+    );
+    subscribeToTopic(amazonSNS, topicARN, queueARN);
```
ajordens: Is there a reason these checks cannot be done up in the constructor vs. down here nested within a `while (true)`?

echo-pubsub/src/main/java/com/netflix/spinnaker/echo/pubsub/amazon/SQSSubscriber.java
```diff
@@ -0,0 +1,267 @@
...
+    while (true) {
+      ReceiveMessageResult receiveMessageResult = amazonSQS.receiveMessage(
+        new ReceiveMessageRequest(queueId)
+          .withMaxNumberOfMessages(AWS_MAX_NUMBER_OF_MESSAGES)
+          .withVisibilityTimeout(subscription.getVisibilityTimeout())
+          .withWaitTimeSeconds(subscription.getWaitTimeSeconds())
+      );
+
+      if (receiveMessageResult.getMessages().isEmpty()) {
+        log.debug("Received no messages");
```
ajordens: Probably missing some context in this log message ...

echo-pubsub/src/main/java/com/netflix/spinnaker/echo/pubsub/amazon/SQSSubscriber.java
```diff
@@ -0,0 +1,267 @@
...
+        continue;
+      }
+
+      receiveMessageResult.getMessages().forEach(message -> {
+        handleMessage(message);
```
ajordens: Anything to worry about if an exception arises during `handleMessage()`?

Right now that would cause any other messages in the same batch to be skipped

echo-pubsub/src/main/java/com/netflix/spinnaker/echo/pubsub/amazon/SQSSubscriber.java
```diff
@@ -0,0 +1,267 @@
...
+  }
+
+  private void handleMessage(Message message) {
+    String messageId = message.getMessageId();
+    String messagePayload = unmarshallMessageBody(message.getBody());
+
+    Map<String, String> stringifiedMessageAttributes = message.getMessageAttributes().entrySet().stream()
+      .collect(Collectors.toMap(Map.Entry::getKey, e -> String.valueOf(e.getValue())));
+
+    log.debug("Received Amazon sqs message: {} with payload: {}\n and attributes: {}", messageId, messagePayload, stringifiedMessageAttributes);
```
ajordens: Are line breaks in log messages normal?

echo-pubsub/src/main/java/com/netflix/spinnaker/echo/pubsub/amazon/SQSSubscriber.java
```diff
@@ -0,0 +1,267 @@
...
+        registry.counter(getProcessedMetricId(subscriptionName())).increment();
+      });
+    }
+  }
+
+  private void handleMessage(Message message) {
+    String messageId = message.getMessageId();
+    String messagePayload = unmarshallMessageBody(message.getBody());
+
+    Map<String, String> stringifiedMessageAttributes = message.getMessageAttributes().entrySet().stream()
```
ajordens: Curious why we went with `Map<String, String>` vs `Map<String, Object>` @jtk54?

echo-pubsub/src/main/java/com/netflix/spinnaker/echo/pubsub/amazon/SQSSubscriber.java
```diff
@@ -0,0 +1,267 @@
...
+      .collect(Collectors.toMap(Map.Entry::getKey, e -> String.valueOf(e.getValue())));
+
+    log.debug("Received Amazon sqs message: {} with payload: {}\n and attributes: {}", messageId, messagePayload, stringifiedMessageAttributes);
+
+    MessageDescription description = MessageDescription.builder()
+      .subscriptionName(subscriptionName())
+      .messagePayload(messagePayload)
+      .messageAttributes(stringifiedMessageAttributes)
+      .pubsubSystem(pubsubSystem)
+      .ackDeadlineMillis(5 * TimeUnit.SECONDS.toMillis(10)) // Set a high upper bound on message processing time.
```
ajordens: `TimeUnit.SECONDS.toMillis(50)` ?

echo-pubsub/src/main/java/com/netflix/spinnaker/echo/pubsub/amazon/SQSSubscriber.java
```diff
@@ -0,0 +1,267 @@
...
+
+    Statement sqsStatement = new Statement(Statement.Effect.Allow).withActions(SQSActions.SendMessage, SQSActions.GetQueueUrl);
+    sqsStatement.setPrincipals(Principal.All);
+    sqsStatement.setResources(Collections.singletonList(new Resource(queue.getArn())));
+
+    return new Policy("allow-sns-or-sqs-send", Arrays.asList(snsStatement, sqsStatement));
+  }
+
+  Id getProcessedMetricId(String subscriptionName) {
+    return registry.createId("echoAmazonPubSub.totalProcessed", "subscriptionName", subscriptionName);
```
ajordens: `echo.pubsub.amazon.totalProcessed`?

echo-pubsub/src/main/java/com/netflix/spinnaker/echo/pubsub/amazon/SQSSubscriber.java
```diff
@@ -0,0 +1,267 @@
...
+      NotificationMessageWrapper wrapper = objectMapper.readValue(messagePayload, NotificationMessageWrapper.class);
+      if (wrapper != null && wrapper.getMessage() != null) {
+        messagePayload = wrapper.getMessage();
+      }
+    } catch (IOException e) {
+      // Try to unwrap a notification message; if that doesn't work,
+      // we're dealing with a message we can't parse. The template or
+      // the pipeline potentially knows how to deal with it.
+      log.error("Unable unmarshal NotificationMessageWrapper. Unknown message type. (body: {})", messageBody, e);
+      registry.counter(getNotAmazonNotificationMetricId(subscriptionName())).increment();
```
ajordens: Not sure what this metric is meant to tell us? The `NotAmazonNotification` sounds a little strange?

echo-pubsub/src/main/java/com/netflix/spinnaker/echo/pubsub/amazon/SQSSubscriber.java
```diff
@@ -0,0 +1,267 @@
...
+                                         ARN queueARN) {
+    amazonSNS.subscribe(topicARN.getArn(), "sqs", queueARN.getArn());
+    return topicARN.getArn();
+  }
+
+  /**
+   * This policy allows operators to choose whether or not to have pubsub messages to be sent via SNS for fanout, or
+   * be sent directly to an SQS queue from the autoscaling group.
+   */
+  private static Policy buildSQSPolicy(ARN queue, ARN topic) {
```
ajordens: a few of these methods were lifted from `clouddriver`?

might be worth also extracting them vs. duplicating?

echo-core/src/main/groovy/com/netflix/spinnaker/echo/artifacts/MessageArtifactTranslator.java
```diff
@@ -78,7 +79,7 @@ private Map readMapValue(ObjectMapper mapper, String messagePayload) {
     try {
       context = mapper.readValue(messagePayload, Map.class);
     } catch (IOException ioe) {
-      throw new RuntimeException(ioe);
+      throw new InvalidPayloadFormatException(ioe);
```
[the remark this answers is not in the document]
emjburns: I'll use a more general kork exception, as well as logging the message payload.

echo-pubsub/src/main/java/com/netflix/spinnaker/echo/config/amazon/AmazonPubsubConfig.java
```diff
@@ -0,0 +1,46 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.spinnaker.echo.config.amazon;
+
+import com.netflix.spinnaker.clouddriver.aws.bastion.BastionConfig;
```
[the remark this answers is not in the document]
emjburns: @ajordens as in pulled out of clouddriver and duplicated in echo?

echo-pubsub/src/main/java/com/netflix/spinnaker/echo/config/amazon/AmazonPubsubConfig.java
```diff
@@ -0,0 +1,46 @@
...
+import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
+import org.springframework.boot.context.properties.EnableConfigurationProperties;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+import org.springframework.context.annotation.Import;
+
+import javax.validation.Valid;
+
+@Configuration
+@Slf4j
```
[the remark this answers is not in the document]
emjburns: Removing. Initially this class did much more than it currently does.

echo-pubsub/src/main/java/com/netflix/spinnaker/echo/config/amazon/AmazonPubsubProperties.java
```diff
@@ -0,0 +1,119 @@
...
+      String name,
+      String accountName,
```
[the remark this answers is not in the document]
emjburns: This is the account that has permission to download the artifact from: https://www.spinnaker.io/reference/artifacts/#format

I agree this isn't clear. I'll remove this field from the config and stop setting it on parsed artifacts, and instead use the account from the arn.

echo-pubsub/src/main/java/com/netflix/spinnaker/echo/config/amazon/AmazonPubsubProperties.java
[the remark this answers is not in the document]
robzienert: Quite like the `echo-pubsub-aws` route, so we're not forced into the javas.

echo-pubsub/src/main/java/com/netflix/spinnaker/echo/config/amazon/AmazonPubsubProperties.java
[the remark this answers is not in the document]
robzienert: Does this mean that we'll start passing around the account ID? The ARN isn't going to have the friendly name that most of the rest of Spinnaker speaks in. Will we translate?

echo-pubsub/src/main/java/com/netflix/spinnaker/echo/pubsub/amazon/ARN.java
```diff
@@ -0,0 +1,53 @@
...
+    if (!sqsMatcher.matches()) {
+      throw new IllegalArgumentException(arn + " is not a valid SNS or SQS ARN");
```
robzienert: Similarly, if this is going to be a generic ARN matcher, this exception message should also be generic..

echo-pubsub/src/main/java/com/netflix/spinnaker/echo/pubsub/amazon/AmazonMessageAcknowledger.java
```diff
@@ -0,0 +1,74 @@
...
+        message.getReceiptHandle()
+      );
+    }
+  }
+
+  @Override
+  public void nack() {
+    // Set visibility timeout to 0, so that the message can be processed by another worker
+    try {
+      amazonSQS.changeMessageVisibility(queueUrl, message.getReceiptHandle(), 0);
```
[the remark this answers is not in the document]
robzienert: I think changing the message visibility is probably the right thing to do here. The queue or message may be configured for especially long visibility timeouts (12 hours!), in which case we wouldn't want it to simply stagnate.

echo-pubsub/src/main/java/com/netflix/spinnaker/echo/pubsub/amazon/SQSSubscriber.java
```diff
@@ -0,0 +1,267 @@
...
+
+  @Override
+  public void run() {
+    log.info("Starting " + getWorkerName());
+
+    while (true) {
+      try {
+        listenForMessages();
+      } catch (Throwable e) {
+        log.error("Unexpected error running " + getWorkerName() + ", restarting", e);
```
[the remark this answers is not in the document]
robzienert: Also for consideration: If the listen op fails repeatedly, do we want to slow our roll for a little while? I'd hate to get into a downstream failure in AWS and then we just thrash the downstream services.

echo-pubsub/src/main/java/com/netflix/spinnaker/echo/config/amazon/AmazonPubsubProperties.java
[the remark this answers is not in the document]
emjburns: I'll take the `echo-pubsub-aws` route.

echo-pubsub/src/main/java/com/netflix/spinnaker/echo/config/amazon/AmazonPubsubProperties.java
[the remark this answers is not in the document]
emjburns: Pulling the account ID out is just for logging/troubleshooting. AWS uses the ARN to do things, and we need the full ARN for configuration. When dealing with sqs in clouddriver we do some converting to nice name, but in front50 we don't.

iiro:
Just a quick question - just to clarify... This is all based on a Spinnaker setup running on AWS (you @emjburns mention this in the first commit - but just to confirm)... Just was thinking, this could be an awesome feature to connect a non-AWS Spinnaker to AWS as well (we actually have this use case). The Google Pub/Sub support seems to have this credential file support.

echo-pubsub/src/main/java/com/netflix/spinnaker/echo/pubsub/amazon/AmazonMessageAcknowledger.java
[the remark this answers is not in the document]
emjburns: +1 to what Rob said. Overall I think having this "messageAcknowledger" is me forcing the Google Pubsub model to fit the Amazon SQS/SNS pieces. I may revisit this in the future. 

That being said, setting the 'visibility' to zero is the best indication I found to tell AWS that this worker will not process the message, and another worker should.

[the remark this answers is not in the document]
emjburns: I'll also add metrics here.

echo-pubsub/src/main/java/com/netflix/spinnaker/echo/pubsub/amazon/SQSSubscriber.java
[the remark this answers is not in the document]
emjburns: Clarified message and added a sleep

echo-pubsub/src/main/java/com/netflix/spinnaker/echo/pubsub/amazon/SQSSubscriber.java
```diff
@@ -0,0 +1,267 @@
...
+  @Override
+  public String getName() {
+    return getWorkerName();
+  }
+
+  @Override
+  public void run() {
+    log.info("Starting " + getWorkerName());
+
+    while (true) {
```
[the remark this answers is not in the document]
emjburns: Yes. Fixing.

echo-pubsub/src/main/java/com/netflix/spinnaker/echo/pubsub/amazon/SQSSubscriber.java
emjburns: Changing to initialize the queue once, and then handle if the queue gets deleted while the worker is running.

echo-pubsub/src/main/java/com/netflix/spinnaker/echo/pubsub/amazon/SQSSubscriber.java
emjburns: Added error handling, and a todo about dead-letter queues

echo-pubsub/src/main/java/com/netflix/spinnaker/echo/config/amazon/AmazonPubsubConfig.java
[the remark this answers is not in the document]
ajordens: 👍

echo-pubsub/src/main/java/com/netflix/spinnaker/echo/config/amazon/AmazonPubsubProperties.java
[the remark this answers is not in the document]
ajordens: 👍

echo-pubsub/src/main/java/com/netflix/spinnaker/echo/config/amazon/AmazonPubsubProperties.java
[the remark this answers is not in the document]
ajordens: 👍

echo-pubsub/src/main/java/com/netflix/spinnaker/echo/pubsub/amazon/AmazonMessageAcknowledger.java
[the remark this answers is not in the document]
ajordens: Only other comment I have is that the `changeMessageVisibility` seems like an optimization (which isn't bad!), I cannot comment as to whether _every_ other possible failure scenario is suitably wrapped in a try/catch but my expectation is that the queue has a reasonable default message visibility (the sqs default of 30s seems reasonable to me).

echo-pubsub/src/main/java/com/netflix/spinnaker/echo/pubsub/amazon/SQSSubscriber.java
ajordens: 👍 

The queue getting deleted while in normal operation is an interesting case ... not sure we handle that particularly well in clouddriver?

echo-pubsub/src/main/java/com/netflix/spinnaker/echo/pubsub/amazon/SQSSubscriber.java
ajordens: 👍

echo-pubsub/src/test/groovy/com/netflix/spinnaker/echo/pubsub/amazon/AmazonSQSSubscriberSpec.groovy
```diff
@@ -0,0 +1,100 @@
...
+    1 * amazonSQS.setQueueAttributes("my-queue-url", [
+      "Policy": SQSSubscriber.buildSQSPolicy(queueARN, topicARN).toJson(),
+      "MessageRetentionPeriod": "1"
+    ])
+    0 * _
+  }
+
+  def 'should unmarshall an sns notification message'() {
+    given:
+    String payload = "{\"Records\":[{" +
```
ajordens: Not sure how tedious it was to create this ... but you could use a multiline string which might look a little better

"""
{"Records": [{ ... }]}
"""

echo-pubsub/src/main/java/com/netflix/spinnaker/echo/config/amazon/AmazonPubsubProperties.java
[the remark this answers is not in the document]
jtk54: We're pretty far along reviewing this for a large refactor like moving this into a separate submodule and rewriting in a different language, particularly in a language I'm unfamiliar with. Echo currently has no Kotlin in it, so I'm also hesitant to introduce another language.

I would be happy to discuss any refactoring ideas you guys have, but I want to get this module functionally complete before making larger scale changes.

echo-pubsub/src/main/java/com/netflix/spinnaker/echo/pubsub/amazon/SQSSubscriber.java
emjburns: No, removing.

echo-pubsub/src/main/java/com/netflix/spinnaker/echo/pubsub/amazon/SQSSubscriber.java
emjburns: It does sound strange. This is meant to track any message delivered to the queue that is not from SNS (message type != Notification). I wanted to see if that's something that will happen, and track this, so that I can work in support for parsing messages from those other sources (like SQS). 

But, after thinking about it, I don't think this will happen often enough to track a metric on it. Removing.

echo-pubsub/src/main/java/com/netflix/spinnaker/echo/pubsub/amazon/SQSSubscriber.java
emjburns: I've put a todo in to extract to `kork-aws`.

echo-pubsub/src/main/java/com/netflix/spinnaker/echo/config/amazon/AmazonPubsubProperties.java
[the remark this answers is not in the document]
emjburns: @jtk54 when I pull it to a different module I'll keep it in Java. I'll do that in a separate PR, this one has had a lot of review. But, for data classes, Kotlin is quite nice. We are trying to gradually shift to Kotlin as the language of choice. First step to that is to not add any more Groovy code because the two don't mix well.

We can chat more offline about why we like the idea of pulling the aws bits to a new module, the google bits to a new module, and making the core pubsub functionality one module.

echo-pubsub/src/main/java/com/netflix/spinnaker/echo/pubsub/amazon/AmazonMessageAcknowledger.java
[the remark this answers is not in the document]
emjburns: I'll watch metrics for this and see how much it happens, and I've added a todo to come back and evaluate once I get this into prestaging to test.

echo-pubsub/src/main/java/com/netflix/spinnaker/echo/pubsub/amazon/SQSSubscriber.java
emjburns: I assume because we're just putting them in the pipeline context as strings from here.

emjburns:
@iiro this is based on Spinnaker having access to an amazon account as a deploy target. From what I see in `hal` and the config options for google pubsub you need to add account credentials in the same way - in the clouddriver config or by adding an account. What are you looking at that indicates otherwise?

lwander:
I don't see which role/permissions each pubsub subscriber will assume - I think that's what @iiro is getting at.

iiro:
@emjburns ok this sounds good - deploy target being AWS and using the accounts-configuration there. I just got confused on "...and must allow SQS subscriptions from the account that Spinnaker is running in" - got me sidetracked. As it has support for defining the "accountName" - everything should be perfectly fine IMO.

@lwander - I did not mean this - at least by purpose - but role assumption is a good question - to clarify. But - I guess this is not an issue - as it uses the "accountName"'s...

emjburns:
@iiro actually I've taken accountName out, because the account can be pulled from the queueUrl. That account name was actually being used for setting artifact permissions, which I'm no longer doing. As I iterate through this please feel free to test it and let me know if it isn't working for you!

iniinikoski:
@emjburns I noticed this one was available now in 1.6.0 - and I tried to get it working like this in echo-local.yml:

```
pubsub:
  amazon:
    enabled: true
    subscriptions:
    - name: sqssnstest
      topicARN: arn:aws:sns:eu-central-1:666677358638:spinnakerTest
      queueARN: arn:aws:sqs:eu-central-1:666677358638:spinnakerTest
      messageFormat: s3
```

The deployment goes through fine - and echo's logs look fine - but if I look the Spinnaker, UI, this is what I see:

screenshot 2018-02-22 16 22 55

screenshot 2018-02-22 14 41 47

Should there be a system type "amazon"? I can only find these guys under "google" or "kafka"...

I do have a Google Pub/Sub configured also with Halyard, maybe this mixed things...

I can do a proper GH issue about this one instead of using this one.

jtk54:
@iiro I think the UI support isn't in yet.

emjburns:
@iiro UI support isn't in yet through Halyard, it's only in through normal Deck settings file. Feel free to create an issue and mention me in it!

PR #233, opened 2018-02-14
refactor(pubsub): adding redisClientDelegate

emjburns:
Using the redis client delegate to talk to redis so that we can add in dynomite support (coming in a later PR). 

Creating autowired constructor instead of properties.

### DIFF
--- a/echo-pubsub/src/main/java/com/netflix/spinnaker/echo/pubsub/PubsubMessageHandler.java
+++ b/echo-pubsub/src/main/java/com/netflix/spinnaker/echo/pubsub/PubsubMessageHandler.java
@@ -22,12 +22,11 @@ import com.netflix.spinnaker.echo.model.Metadata;
 import com.netflix.spinnaker.echo.model.pubsub.MessageDescription;
 import com.netflix.spinnaker.echo.pipelinetriggers.monitor.PubsubEventMonitor;
 import com.netflix.spinnaker.echo.pubsub.model.MessageAcknowledger;
+import com.netflix.spinnaker.kork.jedis.RedisClientDelegate;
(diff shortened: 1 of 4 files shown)


echo-pubsub/src/main/java/com/netflix/spinnaker/echo/pubsub/PubsubMessageHandler.java
```diff
@@ -41,15 +40,21 @@
...
-  @Autowired
   ObjectMapper objectMapper;
 
+  private RedisClientDelegate redisClientDelegate;
+
+  @Autowired
+  public PubsubMessageHandler(PubsubEventMonitor pubsubEventMonitor,
+                              ObjectMapper objectMapper,
+                              RedisClientDelegate redisClientDelegate) {
+    this.pubsubEventMonitor = pubsubEventMonitor;
```
lwander: These can all be `final` now I believe

echo-pubsub/src/main/java/com/netflix/spinnaker/echo/pubsub/PubsubMessageHandler.java
```diff
@@ -41,15 +40,21 @@
...
   private PubsubEventMonitor pubsubEventMonitor;
 
-  @Autowired
   ObjectMapper objectMapper;
 
+  private RedisClientDelegate redisClientDelegate;
+
+  @Autowired
+  public PubsubMessageHandler(PubsubEventMonitor pubsubEventMonitor,
+                              ObjectMapper objectMapper,
```
lwander: I think it's standard for these to continue at double indent level

PR #235, opened 2018-02-16
fix(sqs): fix startup

lwander:
fyi @emjburns


emjburns:
@lwander sorry about this. When does this happen? Even when `amazon` isn't enabled?

lwander:
No worries - only when amazon isn't enabled on echo

PR #240, opened 2018-02-27
feat(artifacts): Support artifacts filtering in Travis triggers

jervi:
Corresponds with https://github.com/spinnaker/igor/pull/223


jervi:
@spinnaker/reviewers PTAL

echo-model/src/main/groovy/com/netflix/spinnaker/echo/model/Artifact.java
```diff
@@ -0,0 +1,46 @@
...
+
+import com.fasterxml.jackson.annotation.JsonAnySetter;
+
+import java.util.HashMap;
+import java.util.Map;
+
+public class Artifact extends com.netflix.spinnaker.kork.artifacts.model.Artifact {
+
+  @Override
+  public Map<String, Object> getMetadata() {
```
jtk54: Are these two methods necessary if they are strictly calling the `super` method?

echo-model/src/main/groovy/com/netflix/spinnaker/echo/model/Artifact.java
```diff
@@ -0,0 +1,46 @@
...
+
+  @Override
+  public void setMetadata(Map<String, Object> metadata) {
+    super.setMetadata(metadata);
+  }
+
+  // Add extra data in the JSON from Igor to metadata:
+  @JsonAnySetter
+  public void putMetadata(String key, Object value) {
+    if (super.getMetadata() == null) {
```
jtk54: Maybe this logic belongs in the superclass in Kork?

echo-model/src/main/groovy/com/netflix/spinnaker/echo/model/Artifact.java
```diff
@@ -0,0 +1,46 @@
...
+ */
+
+package com.netflix.spinnaker.echo.model;
+
+import com.fasterxml.jackson.annotation.JsonAnySetter;
+
+import java.util.HashMap;
+import java.util.Map;
+
+public class Artifact extends com.netflix.spinnaker.kork.artifacts.model.Artifact {
```
lwander: This probably shouldn't live in echo

echo-model/src/main/groovy/com/netflix/spinnaker/echo/model/Artifact.java
```diff
@@ -0,0 +1,46 @@
...
+  public Map<String, Object> getMetadata() {
+    return super.getMetadata();
+  }
+
+  @Override
+  public void setMetadata(Map<String, Object> metadata) {
+    super.setMetadata(metadata);
+  }
+
+  // Add extra data in the JSON from Igor to metadata:
```
lwander: This feels like it'll surprise us/bite us when strange objects are serialized entirely into artifacts by virtue of being placed into metadata.

jervi:
Thanks Lars and Jacob.
I extended the Kork `Artifact` as I thought of this as a special case not relevant in other places (and it only adds stuff for Jackson deserialization, it doesn't actually change the model), but I have no problem adding it directly to the superclass in Kork instead.
The rationale for the change is that the Jenkins/Travis artifacts from Igor have fields that Kork `Artifact` don't have: https://github.com/spinnaker/igor/blob/master/igor-web/src/main/groovy/com/netflix/spinnaker/igor/build/model/GenericArtifact.groovy
So it seemed like a simple way of supporting arbitrary fields.
A Jenkins/Travis artifact then looks like this in Echo:
```
{
    "metadata": {
        "displayPath": "<somePath>",
        "fileName": "<someFileName>",
        "relativePath": "<someRelativePath>"
    },
    "name": "<name>",
    "reference": "<reference>",
    "type": "<type>",
    "version": "<version>"
}
```

Another solution could be to add `displayPath`, `fileName` and `relativePath` to Kork's `Artifact`?

echo-model/src/main/groovy/com/netflix/spinnaker/echo/model/Artifact.java
jervi: True, fixed

jervi:
@jtk54 and @lwander, any preferences on what to do with the missing fields (`displayPath`, `fileName` and `relativePath`)? 

Suggestions:

- Ignore them
- Add the fields to the Kork `Artifact` class
- Put the data in the metadata map
  - As implemented in this PR
  - Or add the `@JsonAnySetter` to the Kork `Artifact` class

jtk54:
@jervi Sorry for the response delay -- I think adding to the metadata as implemented here works well.

jervi:
Ref Slack chat: I've moved the `@JsonAnySetter` to Kork: https://github.com/spinnaker/kork/pull/138

jtk54:
Tested with the new Spinnaker dependenies bump? Good to merge?

jervi:
Yes, it's good to merge 👍

PR #246, opened 2018-03-27
feat(artifacts): Docker registry event artifact parser

wjoel:
Extracts Docker images from push events sent by the Docker registry.

Example config for a Docker registry Kubernetes manifest:
```
        - name: REGISTRY_NOTIFICATIONS_ENDPOINTS
          value: |
            - name: spinnaker
              url: http://spin-gate.spinnaker.svc:8084/webhooks/webhook/dockerregistry
              timeout: 500ms
              threshold: 5
              backoff: 1s
```

### DIFF
--- a/echo-webhooks/src/main/groovy/com/netflix/spinnaker/echo/artifacts/DockerRegistryArtifactExtractor.java
+++ b/echo-webhooks/src/main/groovy/com/netflix/spinnaker/echo/artifacts/DockerRegistryArtifactExtractor.java
@@ -1,0 +1,116 @@
+/*
+ * Copyright 2018 Joel Wilsson
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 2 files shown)


lwander:
Can you link to the spec this was implemented under?

wjoel:
Sure, it's from https://docs.docker.com/registry/notifications/

ethanfrogers:
Are these implemented by any of the public Docker registries? Quay or Dockerhub?

ethanfrogers:
I guess I can answer my own question: https://docs.quay.io/guides/notifications.html

ethanfrogers:
this documentation could be out of date but even Dockerhub doesn't seem to to follow that spec: https://docs.docker.com/docker-hub/repos/#webhooks

@wjoel are you guys hosting an internal Docker registry that conforms to that contract?

wjoel:
No, this is what is supported by https://docs.docker.com/registry/ and I'm using it for my own private setup, not at work. Some companies may be using it internally (certainly many individuals are), but we use GCR at my employer, and use its pubsub notification feature.

Docker hub notification support was added in https://github.com/spinnaker/echo/pull/228, while this is for Docker registry (the self-hosted application/container). This is important in order to support non-cloud Kubernetes setups, which I hope will also make it possible to create a standalone codelab without external dependencies (probably using minikube).

@ethanfrogers Similar support for notifications from Quay would be great as well.

ethanfrogers:
aaaaaand that's what happens when i get ahead of myself. sorry @wjoel. i didn't realize they were provider specific implementations.

PR #250, opened 2018-04-04
feat(aws pubsub): adding filterPolicy to sns topics

emjburns:
- Adding filter policies to sns subscriptions as described https://docs.aws.amazon.com/sns/latest/dg/message-filtering.html so that we have more control over sns subscriptions.
- Adding some try/catch to properly show and recover from amazon exceptions (usually as a result of incorrect permissions) which were impossible to detect without logging the exception. Not all errors are fatal, but it might be better to throw new a new RuntimeException instead of just logging?
- Fixing the spelling of marshal and other typos.

### DIFF
--- a/echo-pubsub-aws/src/main/java/com/netflix/spinnaker/echo/pubsub/aws/SQSSubscriber.java
+++ b/echo-pubsub-aws/src/main/java/com/netflix/spinnaker/echo/pubsub/aws/SQSSubscriber.java
@@ -19,6 +19,8 @@ package com.netflix.spinnaker.echo.pubsub.aws;
 import com.amazonaws.auth.policy.*;
 import com.amazonaws.auth.policy.actions.SQSActions;
 import com.amazonaws.services.sns.AmazonSNS;
+import com.amazonaws.services.sns.model.SetSubscriptionAttributesResult;
+import com.amazonaws.services.sns.model.SubscribeResult;
 import com.amazonaws.services.sqs.AmazonSQS;
 import com.amazonaws.services.sqs.model.Message;
 import com.amazonaws.services.sqs.model.QueueDoesNotExistException;
@@ -140,7 +142,7 @@ public class SQSSubscriber implements Runnable, PubsubSubscriber {
     this.queueId = ensureQueueExists(
       amazonSQS, queueARN, topicARN, subscription.getSqsMessageRetentionPeriodSeconds()
     );
-    subscribeToTopic(amazonSNS, topicARN, queueARN);
+    setUpSubscription(amazonSNS, topicARN, queueARN, subscription);
   }
 
   private void listenForMessages() {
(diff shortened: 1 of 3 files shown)


echo-pubsub-aws/src/main/java/com/netflix/spinnaker/echo/pubsub/aws/SQSSubscriber.java
```diff
@@ -243,12 +245,44 @@ private static String ensureQueueExists(AmazonSQS amazonSQS,
...
                                          ARN topicARN,
                                          ARN queueARN) {
-    amazonSNS.subscribe(topicARN.getArn(), "sqs", queueARN.getArn());
-    return topicARN.getArn();
+      SubscribeResult subscribeResult = amazonSNS.subscribe(topicARN.getArn(), "sqs", queueARN.getArn());
+      return subscribeResult.getSubscriptionArn();
+  }
+
+  private static void applyFilterPolicyToSubscription(AmazonSNS amazonSNS, ARN topicARN, ARN queueARN, String subscriptionArn, String filterPolicy) {
+    if (filterPolicy == null || filterPolicy.equals("")) {
```
robzienert: What about `Strings.isEmptyOrNull`?

echo-pubsub-aws/src/main/java/com/netflix/spinnaker/echo/pubsub/aws/SQSSubscriber.java
```diff
@@ -243,12 +245,44 @@ private static String ensureQueueExists(AmazonSQS amazonSQS,
...
+  }
+
+  private static String refreshSubscription(AmazonSNS amazonSNS, ARN topicARN, ARN queueARN, String subscriptionArn){
+    log.debug("Refreshing subscription for topic {} and queue {}", topicARN, queueARN);
+    try {
+      amazonSNS.unsubscribe(subscriptionArn);
+    } catch (Exception e) {
+      log.error("Unable to unsubscribe from queue {} from topic {}.", queueARN, topicARN, e);
+    }
+    return subscribeToTopic(amazonSNS, topicARN, queueARN);
```
robzienert: It's a real bummer that AWS forces us into doing something like this.

I could see potential for losing notifications with this pattern if we wind up changing our filters while servicing production notifications. The only work around I can think of off the top of my head... Any time we want to refresh subscriptions, we create a new versioned SQS queue with the updated subscription, unsubscribe the old queue and let it drain, then delete the old SQS queue once we've received empty responses for a few cycles.

My presumption is that Rocket (and other publishers) will be operating on a fire-and-forget design, which leaves us holding the bag on making sure we receive.

ajordens: I feel like our tolerance for missed messages should be extremely low/zero.

I'd be inclined to do as @robzienert has called out ... maintain 1..* queues and monitor draining as needed.

emjburns:
Closing this while I follow up with AWS to see if they can make this API better. We should not miss messages. 

If AWS won't support this I'll revisit this PR with the solution proposed above.

PR #260, opened 2018-05-08
fix(pubsub): Change /pubsub/subscriptions API to return a list of a datatype containing pubsubSystem and subscriptionName fields.

joonlim:
* Part of a bug fix to display only Pub/Sub subscription names of subscriptions associated with a given Pub/Sub system type (e.g., 'google').
* This fix also touches gate and deck.


joonlim:
@jtk54 @lwander Could you please review?

echo-pubsub-core/src/main/java/com/netflix/spinnaker/echo/pubsub/PubsubSubscriptionController.java
```diff
@@ -34,7 +36,18 @@
...
+  List<PubsubSubscriptionBySystem> getSubscriptions() {
+    return pubsubSubscribers
+        .getAll()
+        .stream()
+        .map(s -> new PubsubSubscriptionBySystem(s.pubsubSystem().toString(), s.getName()))
+        .collect(Collectors.toList());
+  }
+
+  @AllArgsConstructor
+  @Getter
```
jtk54: I think you can omit the `@Getter` here.

lwander: `@Data` is the right call. Also, I would recommend using a `@Builder` and to construct this class instead of the constructor above. You will need `@AllArgsConstructor` and `@NoArgsConstructor` to get serialization/deserialization to work.

PR #263, opened 2018-05-22
fix(pubsub): Restart Google Pubsub subscriber on failures

wjoel:
Creates a new `Subscriber` on failures, as according to https://github.com/GoogleCloudPlatform/google-cloud-java/issues/1579#issuecomment-282603572 we can `"restart" the Subscriber by creating a new Subscriber object`.

I've also removed `.setMaxAckExtensionPeriod(Duration.ofSeconds(0))` as https://github.com/GoogleCloudPlatform/google-cloud-java/issues/2465#issuecomment-351941181 recommends against it.

Potentially solves https://github.com/spinnaker/spinnaker/issues/2762 but needs more testing. I've verified that echo resumes consuming pubsub messages after I removed its `Pub/Sub Subscriber` permission (took a little while before it noticed) and then added it back again.

### DIFF
--- a/echo-pubsub-google/src/main/java/com/netflix/spinnaker/echo/pubsub/google/GooglePubsubSubscriber.java
+++ b/echo-pubsub-google/src/main/java/com/netflix/spinnaker/echo/pubsub/google/GooglePubsubSubscriber.java
@@ -26,8 +26,8 @@ import com.google.cloud.pubsub.v1.AckReplyConsumer;
 import com.google.cloud.pubsub.v1.MessageReceiver;
 import com.google.cloud.pubsub.v1.Subscriber;
 import com.google.common.util.concurrent.MoreExecutors;
+import com.google.pubsub.v1.ProjectSubscriptionName;
 import com.google.pubsub.v1.PubsubMessage;
-import com.google.pubsub.v1.SubscriptionName;
 import com.netflix.spinnaker.echo.artifacts.MessageArtifactTranslator;
 import com.netflix.spinnaker.echo.config.GooglePubsubProperties;
 import com.netflix.spinnaker.echo.model.pubsub.MessageDescription;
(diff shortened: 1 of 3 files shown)


echo-pubsub-google/src/main/java/com/netflix/spinnaker/echo/pubsub/google/GooglePubsubSubscriber.java
```diff
@@ -115,17 +118,33 @@ public static GooglePubsubSubscriber buildSubscriber(GooglePubsubProperties.Goog
...
+  public void start() {
+    this.subscriber = Subscriber
+      .newBuilder(ProjectSubscriptionName.of(project, subscriptionName), messageReceiver)
+      .setCredentialsProvider(FixedCredentialsProvider.create(credentials))
+      .build();
 
-    return new GooglePubsubSubscriber(subscription.getName(), subscriptionName, project, subscriber);
+    subscriber.addListener(new GooglePubsubFailureHandler(this, formatSubscriptionName(project, subscriptionName)), MoreExecutors.directExecutor());
+    subscriber.startAsync().awaitRunning();
+    log.info("Google Pubsub subscriber started");
```
lwander: Can we add the subscription & project name to this log message?

echo-pubsub-google/src/main/java/com/netflix/spinnaker/echo/pubsub/google/GooglePubsubSubscriber.java
```diff
@@ -115,17 +118,33 @@ public static GooglePubsubSubscriber buildSubscriber(GooglePubsubProperties.Goog
...
   }
 
+  public void stop() {
+    subscriber.stopAsync();
+  }
+
+  private void restart() {
+    log.info("Waiting to restart Google Pubsub subscriber");
```
jtk54: Can we add more context to this log statement as well?

echo-pubsub-google/src/main/java/com/netflix/spinnaker/echo/pubsub/google/GooglePubsubSubscriber.java
```diff
@@ -81,7 +85,7 @@ public String subscriptionName() {
 
   @Override
   public String getName() {
-    return name;
+    return subscriptionName;
```
jtk54: We made a distinction between the `subscriptionName`, which is the actual locator in GCE and the logical `name` which is used in the UI when suggesting pub/sub subscriptions to use. Is there a reason to change that here?

wjoel:
@jtk54 That makes sense, I'll leave `name` as it was, and I've added the subscription name to the log messages.

jtk54:
Great, thanks for the contribution.

PR #267, opened 2018-05-28
fix(metrics): replace Spring Boot metrics with Spectator

dreynaud:
Also:
* pick up new spinnaker-dependencies for spectator 0.68.0 update
* update jacoco, which otherwise conflicts with spectator
* add explicit dependency on spectatorApi when we use it
* add exception tag on error metrics
* don't use a mock Registry when unnecessary
* make triggers.cronMisfires a timer (so that we can estimate trigger lag)
* adjust some log levels and messages
* add a pipelineConfigsPollingAgent.executionTimeMillis timer

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -32,11 +32,12 @@ buildscript {
 allprojects {
     apply plugin: "spinnaker.project"
     apply plugin: "groovy"
+    apply plugin: 'java'
     apply plugin: 'jacoco'
     group = "com.netflix.spinnaker.echo"
 
     ext {
-        spinnakerDependenciesVersion = project.hasProperty('spinnakerDependenciesVersion') ? project.property('spinnakerDependenciesVersion') : '0.156.0'
+        spinnakerDependenciesVersion = project.hasProperty('spinnakerDependenciesVersion') ? project.property('spinnakerDependenciesVersion') : '0.156.1'
     }
 
     def checkLocalVersions = [spinnakerDependenciesVersion: spinnakerDependenciesVersion]
@@ -78,7 +79,7 @@ allprojects {
 
 subprojects {
     jacoco {
-        toolVersion = '0.7.0.201403182114'
+        toolVersion = '0.8.1'
     }
 
     dependencies {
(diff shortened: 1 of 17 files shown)


build.gradle
```diff
@@ -32,11 +32,12 @@ buildscript {
 allprojects {
     apply plugin: "spinnaker.project"
     apply plugin: "groovy"
+    apply plugin: 'java'
```
robzienert: The java plugin is implicit with the groovy plugin.

echo-pipelinetriggers/src/test/groovy/com/netflix/spinnaker/echo/pipelinetriggers/orca/PipelineInitiatorSpec.groovy
```diff
@@ -1,43 +1,33 @@
...
 import com.netflix.spinnaker.echo.model.Pipeline
-import org.springframework.boot.actuate.metrics.CounterService
+import spock.lang.Unroll
+
 import static rx.Observable.empty
 
 class PipelineInitiatorSpec extends Specification {
 
-  def counter = Stub(CounterService)
+  def registry = new DefaultRegistry()
```
robzienert: You can use `NoopRegistry` instead for tests.

build.gradle
dreynaud: ah good to know, that was a remnant of me trying to debug an issue with jacoco

echo-pipelinetriggers/src/test/groovy/com/netflix/spinnaker/echo/pipelinetriggers/orca/PipelineInitiatorSpec.groovy
dreynaud: yeah I'll do that, I happened to chat with Brian H about that. I went with a DefaultRegistry because that's what the spectator testing docs show: https://github.com/Netflix/spectator/blob/master/docs/intro/testing.md

PR #269, opened 2018-06-01
fix(artifacts): Artifacts with usePrior flag do not need to match a trigger

PerGon:
When echo receives a trigger and is matching the trigger event to the pipeline triggers, there might happen that a pipeline has 2 triggers and some expected artifacts - some expected artifacts can be flagged as "If not present, use from prior execution".
When that's the case, those artifacts do not need to be matched for the pipeline to be triggered.

Example:
Pipeline FooBar expects 3 Artifacts
- Artifact A (flagged as "If not present, use from prior execution")
- Artifact B (flagged as "If not present, use from prior execution")
- Artifact C (flagged as "If not present, use from prior execution")

The pipeline has 2 triggers:
- Trigger 1 - Provides artifacts A, B and C
- Trigger 2 - Do not provides artifacts

Currently, when a "Trigger 2" event is received it is does not start the pipeline because it doesn't match the expected artifacts - it should and just get the artifacts from a previous execution. This PR fixes that.

### DIFF
--- a/echo-pipelinetriggers/src/main/java/com/netflix/spinnaker/echo/pipelinetriggers/artifacts/ArtifactMatcher.java
+++ b/echo-pipelinetriggers/src/main/java/com/netflix/spinnaker/echo/pipelinetriggers/artifacts/ArtifactMatcher.java
@@ -43,7 +43,13 @@ public class ArtifactMatcher {
     List<ExpectedArtifact> expectedArtifacts = pipelineExpectedArtifacts == null ? new ArrayList<>() : pipelineExpectedArtifacts
         .stream()
         .filter(e -> expectedArtifactIds.contains(e.getId()))
+        .filter(e -> !e.isUsePriorArtifact())
         .collect(Collectors.toList());
+
+    if ( expectedArtifacts == null || expectedArtifacts.isEmpty()) {
+      log.info("All expected artifacts are marked as UsePrior hence no need for matching.");
+      return true;
+    }
 
     if (messageArtifacts.size() > expectedArtifactIds.size()) {
       log.warn("Parsed message artifacts (size {}) greater than expected artifacts (size {}), continuing trigger anyway", messageArtifacts.size(), expectedArtifactIds.size());
--- a/echo-pipelinetriggers/src/main/java/com/netflix/spinnaker/echo/pipelinetriggers/artifacts/ArtifactMatcher.java
+++ b/echo-pipelinetriggers/src/main/java/com/netflix/spinnaker/echo/pipelinetriggers/artifacts/ArtifactMatcher.java
@@ -39,17 +39,20 @@ public class ArtifactMatcher {
       return true;
     }
 
-    List<ExpectedArtifact> pipelineExpectedArtifacts = pipeline.getExpectedArtifacts();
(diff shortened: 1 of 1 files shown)


echo-pipelinetriggers/src/main/java/com/netflix/spinnaker/echo/pipelinetriggers/artifacts/ArtifactMatcher.java
```diff
@@ -39,7 +39,16 @@ public static Boolean anyArtifactsMatchExpected(List<Artifact> messageArtifacts,
...
+        .stream()
+        .filter(e -> !e.isUsePriorArtifact())
+        .collect(Collectors.toList());
+
+    if ( pipelineExpectedArtifacts == null || pipelineExpectedArtifacts.isEmpty()) {
+      log.info("All pipeline expected artifacts are marked as UsePrior hence no need for matching.");
+      return true;
+    }
+
     List<ExpectedArtifact> expectedArtifacts = pipelineExpectedArtifacts == null ? new ArrayList<>() : pipelineExpectedArtifacts
```
lwander: I think the real fix is to return 'true' when this is empty. The reason being, the "anyMatch" on the below line here returns "false" when the list of expected artifacts to check is empty: https://github.com/spinnaker/echo/pull/269/files#diff-414234a528208013567f43f82f05dc76R64

PerGon: Unless someone ends up messing with the expected artifacts ids manually, `expectedArtifacts` should not ever be empty.
`expectedArtifacts` should always be a subset of `pipelineExpectedArtifacts` and it's "emptiness" is already checked in here: https://github.com/PerGon/echo/blob/bda18a6d1d59e2e39568040fdd39da549e391df3/echo-pipelinetriggers/src/main/java/com/netflix/spinnaker/echo/pipelinetriggers/artifacts/ArtifactMatcher.java#L38
(it might happen that due to some other bug, the `expectedArtifacts` will be empty, so that check should be added. But that not the case I'm trying to build :) will try to explain better in a comment)

lwander: Well if that's the case, and trigger 2 does not expect any artifacts -- isn't this already working as intended?

PerGon:
Oh well, nevermind! I see now where I did messed up 🙈 I'm embarassed 😊 
I missinterpreted the use of expected artifacts with use prior - I thought I still had to add them to a trigger even tho I knew that specific trigger would not have the artifact... 🤦‍♂️ 

I'm closing this!

echo-pipelinetriggers/src/main/java/com/netflix/spinnaker/echo/pipelinetriggers/artifacts/ArtifactMatcher.java
PerGon: After realizing I was messing up, yes, all works fine now! Sorry :D

PR #270, opened 2018-06-04
fix(cron): fallback for cron triggers with null id

dreynaud:
If pipeline configurations with null cron trigger ids somehow went past
front50, we would have "flagging" registrations: new triggers would be found,
then on the next pass of the PipelineConfigsPollingAgent we would find
scheduled actions with no corresponding triggers (since they don't have)
a matching id. We would then deregister then, only to register them again
right away, and so on.

With this change, we no longer need a persisted id for cron triggers in
the pipeline configurations (which is bound to be messed with, deleted,
copied, mangled...)

### DIFF
--- a/echo-model/src/main/java/com/netflix/spinnaker/echo/model/Trigger.java
+++ b/echo-model/src/main/java/com/netflix/spinnaker/echo/model/Trigger.java
@@ -21,15 +21,19 @@ import com.fasterxml.jackson.databind.annotation.JsonPOJOBuilder;
 import lombok.Builder;
 import lombok.ToString;
 import lombok.Value;
+import lombok.experimental.NonFinal;
 import lombok.experimental.Wither;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import java.util.List;
 import java.util.Map;
+import java.util.UUID;
 
 @JsonDeserialize(builder = Trigger.TriggerBuilder.class)
 @Builder(toBuilder = true)
 @Wither
-@ToString(of = {"type", "master", "job", "cronExpression", "source", "project", "slug", "account", "repository", "tag", "parameters", "payloadConstraints", "attributeConstraints", "branch", "runAsUser", "subscriptionName", "pubsubSystem", "expectedArtifactIds", "payload"}, includeFieldNames = false)
+@ToString(of = {"id", "type", "master", "job", "cronExpression", "source", "project", "slug", "account", "repository", "tag", "parameters", "payloadConstraints", "attributeConstraints", "branch", "runAsUser", "subscriptionName", "pubsubSystem", "expectedArtifactIds", "payload"}, includeFieldNames = false)
 @Value
 public class Trigger {
   public enum Type {
@@ -52,6 +56,8 @@ public class Trigger {
       return type;
     }
   }
+
+  private static final Logger log = LoggerFactory.getLogger(Trigger.class);
 
   boolean enabled;
   String id;
@@ -85,6 +91,23 @@ public class Trigger {
   String pubsubSystem;
   List<String> expectedArtifactIds;
   Map<String, ?> lastSuccessfulExecution;
+
+  // this is set after deserialization, not in the json representation
+  @NonFinal
+  Pipeline parent;
+
+  public String getIdWithFallback() {
+    if (id != null) {
+      return id;
+    }
+
+    if (parent == null) {
+      log.warn("Trigger with unknown parent: {}", this);
+      return UUID.nameUUIDFromBytes(this.toString().getBytes()).toString();
+    }
+
+    return UUID.nameUUIDFromBytes((parent.getId() + "/" + this.toString()).getBytes()).toString();
+  }
 
   public Trigger atBuildNumber(final int buildNumber) {
     return this.toBuilder()
--- a/echo-model/src/main/java/com/netflix/spinnaker/echo/model/Trigger.java
+++ b/echo-model/src/main/java/com/netflix/spinnaker/echo/model/Trigger.java
@@ -30,10 +30,13 @@ import java.util.List;
 import java.util.Map;
 import java.util.UUID;
 
+/**
+ * the values we include in toString are meaningful, they are hashed and become part of generateFallbackId()
+ */
 @JsonDeserialize(builder = Trigger.TriggerBuilder.class)
 @Builder(toBuilder = true)
 @Wither
-@ToString(of = {"id", "type", "master", "job", "cronExpression", "source", "project", "slug", "account", "repository", "tag", "parameters", "payloadConstraints", "attributeConstraints", "branch", "runAsUser", "subscriptionName", "pubsubSystem", "expectedArtifactIds", "payload"}, includeFieldNames = false)
+@ToString(of = {"id", "parent", "type", "master", "job", "cronExpression", "source", "project", "slug", "account", "repository", "tag", "parameters", "payloadConstraints", "attributeConstraints", "branch", "runAsUser", "subscriptionName", "pubsubSystem", "expectedArtifactIds", "payload"}, includeFieldNames = false)
 @Value
 public class Trigger {
   public enum Type {
@@ -93,20 +96,10 @@ public class Trigger {
   Map<String, ?> lastSuccessfulExecution;
 
   // this is set after deserialization, not in the json representation
-  @NonFinal
   Pipeline parent;
 
-  public String getIdWithFallback() {
-    if (id != null) {
-      return id;
-    }
-
-    if (parent == null) {
-      log.warn("Trigger with unknown parent: {}", this);
-      return UUID.nameUUIDFromBytes(this.toString().getBytes()).toString();
-    }
-
-    return UUID.nameUUIDFromBytes((parent.getId() + "/" + this.toString()).getBytes()).toString();
+  public String generateFallbackId() {
+    return UUID.nameUUIDFromBytes(this.toString().getBytes()).toString();
   }
 
   public Trigger atBuildNumber(final int buildNumber) {
--- a/echo-pipelinetriggers/src/main/java/com/netflix/spinnaker/echo/pipelinetriggers/PipelineCache.java
+++ b/echo-pipelinetriggers/src/main/java/com/netflix/spinnaker/echo/pipelinetriggers/PipelineCache.java
@@ -18,6 +18,7 @@ package com.netflix.spinnaker.echo.pipelinetriggers;
 
 import com.netflix.spectator.api.Registry;
 import com.netflix.spinnaker.echo.model.Pipeline;
+import com.netflix.spinnaker.echo.model.Trigger;
 import com.netflix.spinnaker.echo.services.Front50Service;
 import lombok.NonNull;
 import lombok.extern.slf4j.Slf4j;
@@ -30,9 +31,11 @@ import rx.Subscription;
 import javax.annotation.PostConstruct;
 import javax.annotation.PreDestroy;
 import java.time.Instant;
+import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
 import java.util.concurrent.atomic.AtomicReference;
+import java.util.stream.Collectors;
 
 import static java.time.Instant.now;
 import static java.util.concurrent.TimeUnit.SECONDS;
@@ -101,7 +104,7 @@ public class PipelineCache implements MonitoredPoller {
 
   private void cachePipelines(final List<Pipeline> pipelines) {
     log.info("Refreshing pipelines");
-    this.pipelines.set(pipelines);
+    this.pipelines.set(decorateTriggers(pipelines));
   }
 
   private void onFront50Request(final long tick) {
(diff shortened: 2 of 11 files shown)


echo-model/src/main/java/com/netflix/spinnaker/echo/model/Trigger.java
```diff
@@ -86,6 +92,23 @@ public String toString() {
...
+    if (id != null) {
+      return id;
+    }
+
+    if (parent == null) {
+      log.warn("Trigger with unknown parent: {}", this);
+      return UUID.nameUUIDFromBytes(this.toString().getBytes()).toString();
+    }
+
+    return UUID.nameUUIDFromBytes((parent.getId() + "/" + this.toString()).getBytes()).toString();
```
dreynaud: core piece: generating a fallback id if there isn't a hardcoded one, based on the parent pipeline plus value of this trigger

echo-scheduler/src/main/groovy/com/netflix/spinnaker/echo/scheduler/actions/pipeline/impl/PipelineConfigsPollingAgent.groovy
```diff
@@ -78,20 +78,20 @@ class PipelineConfigsPollingAgent extends AbstractPollingAgent {
...
 
-      /**
-       * Only interested in pipelines that have triggers and that too scheduled triggers
-       */
+      // Only interested in pipelines that have triggers and that too scheduled triggers
       def pipelines = pipelineCache.getPipelines().findAll {
         it.triggers && it.triggers.any { TRIGGER_TYPE.equalsIgnoreCase(it.type) }
       }
 
+      def triggerRepo = new TriggerRepository(pipelines)
```
dreynaud: as far as the `PipelineConfigsPollingAgent` is concerned, this is where the magic happens

echo-scheduler/src/main/groovy/com/netflix/spinnaker/echo/scheduler/actions/pipeline/impl/PipelineConfigsPollingAgent.groovy
```diff
@@ -173,48 +170,46 @@ class PipelineConfigsPollingAgent extends AbstractPollingAgent {
...
+      newTriggers.each { trigger ->
+        if (!Trigger.Type.CRON.toString().equalsIgnoreCase(trigger.type) || !trigger.enabled) {
+//          log.debug("Skipping disabled or non-cron trigger ${trigger}")
+          return
         }
-        if (triggers) {
-          log.info("Found '${triggers.size()}' new cron trigger(s) for ${pipeline}...")
+
+        Pipeline pipeline = trigger.parent
+        if (pipeline.disabled) {
```
dreynaud: we used to miss this case, we would scheduled triggers for disabled pipelines

echo-scheduler/src/main/groovy/com/netflix/spinnaker/echo/scheduler/actions/pipeline/impl/TriggerRepository.groovy
```diff
@@ -0,0 +1,63 @@
...
+@Slf4j
+class TriggerRepository {
+  private final Map<String, Trigger> triggersById;
+
+  TriggerRepository(List<Pipeline> pipelines) {
+    triggersById = new HashMap<>()
+    for (Pipeline pipeline : pipelines) {
+      for (Trigger trigger : pipeline.triggers) {
+        if (trigger.parent == null) {
+          trigger.parent = pipeline
```
dreynaud: looking for feedback: I am not entirely happy with this. It feels gross to sort of hotpatch triggers after the fact like this, but I couldn't think of a better way to accomplish this. Note that once the dirty deed is done, it is quite nice to be able to look up `trigger.parent`

ajordens: I suppose you could also do something like the below inside of `PipelineCache` (aside from getting more complicated with json deserializers).

```
private void cachePipelines(List<Pipeline> pipelines) {
    log.info("Refreshing pipelines");

    pipelines = pipelines
      .stream()
      .map(p -> {
        List<Trigger> triggers = p.getTriggers()
          .stream()
          .map(t -> {
            t = t.withPipeline(p);
            t = t.withId("my_new_id");
            return t;
          })
          .collect(Collectors.toList());

        return p.withTriggers(triggers);
      })
      .collect(Collectors.toList());

    this.pipelines.set(pipelines);
  }
```

ajordens: My thinking would be along the lines of always generating a new id that is scoped within the `pipelineId` ... it could be a simple concatenation most likely, I'm not sure it _has_ to be a standard UUID.

Moving away from a persisted `id` in `front50`, I'm not sure there's much value in detecting where previously someone may have inadvertently re-used a trigger id across different pipelines?

Doing it within `PipelineCache` would keep things reasonable colocated and the rest of the logic would presumably not need to change much/at all?

echo-model/src/main/java/com/netflix/spinnaker/echo/model/Trigger.java
ajordens: Not sure I'm a huge fan of having both `id` and `getIdWithFallback` ... see below for a possible workaround.

Presumably you never actually want to call `id` directly?

echo-scheduler/src/main/groovy/com/netflix/spinnaker/echo/scheduler/actions/pipeline/impl/PipelineConfigsPollingAgent.groovy
ajordens: Good catch!

echo-scheduler/src/main/groovy/com/netflix/spinnaker/echo/scheduler/actions/pipeline/impl/TriggerRepository.groovy
dreynaud: Doing it in `PipelineCache` seems sensible, it does provide a more consistent interface 👌

echo-model/src/main/java/com/netflix/spinnaker/echo/model/Trigger.java
dreynaud: Indeed, we never want to call `id` directly. Thanks for the suggestion, let me massage it in!

dreynaud: Massaging complete in da7d74f. I had to rework quite a few tests, since we are now generating copies of pipelines and triggers, but I think it is much cleaner now!

echo-model/src/main/java/com/netflix/spinnaker/echo/model/Trigger.java
```diff
@@ -86,6 +95,13 @@ public String toString() {
   List<String> expectedArtifactIds;
   Map<String, ?> lastSuccessfulExecution;
 
+  // this is set after deserialization, not in the json representation
+  Pipeline parent;
+
+  public String generateFallbackId() {
+    return UUID.nameUUIDFromBytes(this.toString().getBytes()).toString();
```
dreynaud: This is how we generate new ids. Any thoughts on using specific fields vs relying on `toString()`?

echo-pipelinetriggers/src/main/java/com/netflix/spinnaker/echo/pipelinetriggers/PipelineCache.java
```diff
@@ -114,4 +117,28 @@ private void onFront50Error(Throwable e) {
...
+    return pipelines
+      .stream()
+      .map(p -> {
+        List<Trigger> triggers = p.getTriggers();
+        if (triggers == null) {
+          return p;
+        }
+
+        List<Trigger> newTriggers = new ArrayList<>(triggers.size());
+        Pipeline newPipe = p.withTriggers(newTriggers);
```
ajordens: I've never been a fan of truncating variable names ... `newPipeline` reads better than `newPipe` to me.

could even just do ...

```
.map(pipeline -> {
...

pipeline = pipeline.withTriggers(
  triggers.stream().map { ... }.collect(Collectors.toList())
)

return pipeline
}
```

echo-model/src/main/java/com/netflix/spinnaker/echo/model/Trigger.java
ajordens: does seem a bit strange to me ... I've never done it myself.

The `expression` itself combined with parent pipeline id should be sufficiently unique? (I doubt we actually stop someone from registering the same cron expression multiple times tho?)

echo-scheduler/src/main/groovy/com/netflix/spinnaker/echo/scheduler/actions/pipeline/impl/PipelineConfigsPollingAgent.groovy
ajordens: `s/triggerRepo/triggerRepository` for consistentcy

echo-scheduler/src/main/groovy/com/netflix/spinnaker/echo/scheduler/actions/pipeline/impl/TriggerRepository.groovy
```diff
@@ -0,0 +1,62 @@
...
+class TriggerRepository {
+  private final Map<String, Trigger> triggersById
+
+  TriggerRepository(List<Pipeline> pipelines) {
+    triggersById = new HashMap<>()
+    for (Pipeline pipeline : pipelines) {
+      for (Trigger trigger : pipeline.triggers) {
+        if (trigger.id == null) {
+          // see PipelineCache::decorateTriggers
+          throw new IllegalStateException("Trigger with null id ${trigger}")
```
ajordens: this may be too broad of an exception ... could you just throw a `NullPointerException` with the same message

ajordens: you've asserted elsewhere that triggers aren't going to have a null id so is this subsequent check really doing much?

echo-scheduler/src/main/groovy/com/netflix/spinnaker/echo/scheduler/actions/pipeline/impl/TriggerRepository.groovy
```diff
@@ -0,0 +1,62 @@
...
+        }
+
+        Trigger previous = triggersById.put(trigger.id, trigger)
+        if (previous) {
+          log.warn("Duplicate trigger ids found: ${previous} with parent ${previous.parent} and ${trigger} with parent ${trigger.parent}")
```
ajordens: similarly here ... you're regenerating trigger ids so how could you end up with a previous one?

(having a check like this implies that there could be some reason for a duplicate trigger which may prove confusing months/years down the line)

echo-scheduler/src/main/groovy/com/netflix/spinnaker/echo/scheduler/actions/pipeline/impl/TriggerRepository.groovy
dreynaud: They shouldn't be null anymore coming out of `PipelineCache`, but if we there is ever a change that circumvents `decorateTriggers` and we are back to having null ids, it will break assumptions here very hard

echo-scheduler/src/main/groovy/com/netflix/spinnaker/echo/scheduler/actions/pipeline/impl/TriggerRepository.groovy
dreynaud: This is just defensive coding: it's not supposed to happen, but it doesn't cost much to check and I'd rather have a log than fail silently

echo-pipelinetriggers/src/main/java/com/netflix/spinnaker/echo/pipelinetriggers/PipelineCache.java
dreynaud: There's a snag though: the new triggers have a reference to the new pipeline, but we don't have a reference to the new pipeline until we make a copy of it with the new list of triggers. So I end up creating the new pipeline with a placeholder triggers object, and then backfilling the new triggers with the reference to the new pipeline.

echo-scheduler/src/main/groovy/com/netflix/spinnaker/echo/scheduler/actions/pipeline/impl/TriggerRepository.groovy
ajordens: throw an exception ... emit a metric ... log is just going to get lost in the noise if assumptions ever break.

PR #271, opened 2018-06-06
fix(json): prevent infinite recursion when serializing pipelines

dreynaud:
Break the cycle by ignoring the Trigger.parent field in the output.


ajordens:
:thumbs_up:

PR #274, opened 2018-06-14
fix(pubsub): Wait until pipelines are loaded to process pubsub messages

ezimanyi:
This change causes the pubsub event processor to wait until pipelines have been polled at least once before processing pubsub messages.  This fixes a race condition whereby pubsub messages that were received immediately after startup were not triggering any pipelines.


ezimanyi:
@lwander : Yes, this can definitely be generalized to other trigger types (which should also allow deprecation/removal of the old accessor for pipelines).  I think the race condition is worst for PubSub because it tends to process a backlog at startup (vs. webhook for example) but it could certainly happy to any trigger.

I wanted to make sure this approach was reasonable before jumping in and changing it for all triggers---but if this looks good to you, I'll go ahead and generalize.

lwander:
Awesome -- yes this looks great

ezimanyi:
I'll merge this, then have a second PR I'm working on to generalize.

PR #275, opened 2018-06-15
fix(pipelines): Generalize fix for pipeline race condition

ezimanyi:


ezimanyi:
I'm not overly excited about the two places I had to use `toBlocking()` are not my favorite things, but I think that using it at the interface of ReactiveX and imperative code is the one place it isn't a mis-use (and we really have no choice there---either block until we get pipelines, or return an incorrectly-empty list.  We could refactor that code to use callbacks, but that seemed a bit more than I wanted to tackle in this PR.

PR #284, opened 2018-06-23
fix(cron): retain and pass the rebake field to orca

dreynaud:
Otherwise cron triggers will always get rebake=false by the time they
get to orca.


echo-scheduler/src/main/groovy/com/netflix/spinnaker/echo/scheduler/actions/pipeline/impl/PipelineTriggerConverter.groovy
```diff
@@ -34,7 +34,8 @@ class PipelineTriggerConverter {
       triggerType          : trigger.type,
       triggerCronExpression: trigger.cronExpression,
       triggerTimeZoneId    : timeZoneId,
-      triggerEnabled       : Boolean.toString(trigger.enabled)
+      triggerEnabled       : Boolean.toString(trigger.enabled),
+      triggerRebake        : Objects.toString(trigger.rebake) // because trigger.rebake can be null
```
ajordens: it can be null ... but could we just `boolean rebake` and let it default to false?

dreynaud:
good point, changed in

0d21a4f9ae7cddf2ae4e0b6cdaf09a97a2e4f147


thanks!


Le jeu. 19 juil. 2018 à 16:01, Adam Jordens <notifications@github.com> a
écrit :

PR #286, opened 2018-06-25
fix(fiat): Set `@EnableFiatAutoConfig` to avoid unnecessary auth requirements

ajordens:
Original intention was to not require full blown `fiat` to use
`FiatStatus` but unfortunately spring security is brought in as a
transitive dependency.

This PR is a simpler alternative to moving `FiatStatus` and the
related configuration properties classes to a new artifact that
would not bring in any unnecessary dependencies.

fixes spinnaker/spinnaker#2949


ajordens:
FYI @ezimanyi

ezimanyi:
Thanks---just tested and this looks good!

PR #287, opened 2018-06-26
chore(logs+tests): removing excessive logging, adding tests

emjburns:
The logging around pubsub is a little excessive. I'd like to trim down the log messages produced and rely on emitting metrics for number of messages processed etc.

Also adding tests for ArtifactMatcher and adding clarifying text around usage of `isConstraintInPayload`

### DIFF
--- a/echo-pubsub-core/src/main/java/com/netflix/spinnaker/echo/pubsub/PubsubMessageHandler.java
+++ b/echo-pubsub-core/src/main/java/com/netflix/spinnaker/echo/pubsub/PubsubMessageHandler.java
@@ -113,8 +113,7 @@ public class PubsubMessageHandler {
 
 
   private void processEvent(MessageDescription description) {
-    log.info("Processed Pubsub event with payload {}", description.getMessagePayload());
-
+    log.debug("Processing pubsub event with payload {}", description.getMessagePayload());
     Event event = new Event();
     Map<String, Object> content = new HashMap<>();
     Metadata details = new Metadata();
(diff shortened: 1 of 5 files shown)


echo-pubsub-core/src/main/java/com/netflix/spinnaker/echo/pubsub/PubsubMessageHandler.java
```diff
@@ -113,8 +113,6 @@ private String makeKey(MessageDescription description, String messageId) {
 
 
   private void processEvent(MessageDescription description) {
-    log.info("Processed Pubsub event with payload {}", description.getMessagePayload());
```
jtk54: What's left in terms of logging when we process a message? This message specifically has been helpful at least for debugging, can we just s/info/debug here?

ezimanyi: Agree that this particular message has been helpful to me in the past.

emjburns: Sure! I'll put it on debug. I am trying in general to avoid the overload of log messages for very busy pubsub situations.

jtk54: SGTM

PR #293, opened 2018-06-30
fix(healthcheck): pipelines might have no triggers

kevinawoo:
same as https://github.com/spinnaker/echo/pull/292, but for `release-1.8.x`


ezimanyi:
Looks good; @ewiseblatt as you're handling release 1.8 I'll defer merging this to you.

PR #294, opened 2018-07-03
fix(notifications): Support email notifications with ' ' in them

ajordens:
Handle the situation where someone has configured a notification with
' '-separated email addresses (vs. presumably comma separation).


jrsquared:
🥉

PR #299, opened 2018-07-15
feat(artifacts/quay): add native quay parser

ethanfrogers:
adds native quay artifact parser

fixes spinnaker/spinnaker#2322

### DIFF
--- a/echo-webhooks/src/main/groovy/com/netflix/spinnaker/echo/artifacts/QuayArtifactExtractor.java
+++ b/echo-webhooks/src/main/groovy/com/netflix/spinnaker/echo/artifacts/QuayArtifactExtractor.java
@@ -1,0 +1,75 @@
+/*
+ * Copyright 2018 Armory, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License")
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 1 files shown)


ethanfrogers:
@lwander PTAL. there's one odd thing that i'd like your input on.

the quay api will send push notifications for _multiple_ tags (seen [here](http://docs.quay.io/guides/notifications.html#webhook_repo_push)). if the webhook does send 2 updated tags, echo will warn with `Parsed message artifacts (size 2) greater than expected artifacts (size 1), continuing trigger anyway`, but any pipelines listening to the webhook immediately fail. thoughts?

echo-webhooks/src/main/groovy/com/netflix/spinnaker/echo/artifacts/QuayArtifactExtractor.java
```diff
@@ -0,0 +1,78 @@
...
+    private String namespace;
+    private String name;
+    private String homepage;
+    @JsonProperty("docker_url")
+    private String dockerUrl;
+    @JsonProperty("updated_tags")
+    private List<String> updatedTags;
+  }
+}
+
```
lwander: Extra newlines?

ethanfrogers: will remove!

ethanfrogers:
RTM!

ethanfrogers:
@lwander this is ready to go!

PR #301, opened 2018-07-22
feat(artifacts/gitlab): gitlab artifact parser

ethanfrogers:
add a gitlab artifact parser for V4 of their API. i specified V4
specifically because there may be some assumptions made that dont
necessarily work with older versions of gitlab. worked based off of the
payload found here: https://docs.gitlab.com/ee/user/project/integrations/webhooks.html#push-events

fixes spinnaker/spinnaker#3072


echo-webhooks/src/main/groovy/com/netflix/spinnaker/echo/controllers/WebhooksController.groovy
```diff
@@ -102,6 +102,11 @@ class WebhooksController {
           sendEvent = false
         }
         log.info('Webhook event received {} {} {} {} {} {}', kv('type', type), kv('event_type', event.content.event_type), kv('hook_id', event.content.hook_id), kv('repository', event.content.repository.full_name), kv('request_id', event.content.request_id), kv('branch', event.content.branch))
+      } else if (source == 'gitlab') {
```
lwander: Could you add a TODO & file an issue to replace this large `if else` block?

ethanfrogers: ha! yeah, no problem!

ethanfrogers: added!

ethanfrogers:
ready to merge!

PR #302, opened 2018-07-26
fix(github): Fail github authentication when header is absent

ezimanyi:
The logic to check the authentication of github triggers is incorrect; if a trigger is configured with a secret, but the github webhook does not contain the X-Hub-Signature header, we incorrectly allow the
trigger to fire. We should instead fail the authentication and not fire the trigger.


echo-pipelinetriggers/src/main/java/com/netflix/spinnaker/echo/pipelinetriggers/monitor/GitEventMonitor.java
```diff
@@ -131,23 +131,36 @@ protected void emitMetricsOnMatchingPipeline(Pipeline pipeline) {
...
-    }
+  private boolean passesGithubAuthenticationCheck(TriggerEvent event, Trigger trigger) {
+    boolean triggerHasSecret = StringUtils.isNotEmpty(trigger.getSecret());
+    boolean eventHasSignature = event.getDetails().getRequestHeaders().containsKey(GITHUB_SECURE_SIGNATURE_HEADER);
 
-    String header = headers.getFirst(GITHUB_SECURE_SIGNATURE_HEADER);
-    log.debug("GitHub Signature detected. " + GITHUB_SECURE_SIGNATURE_HEADER + ": " + header);
-    String signature = StringUtils.removeStart(header, "sha1=");
+    if (triggerHasSecret && !eventHasSignature) {
+      log.warn("Received GitEvent from Github without secure signature, for trigger configured with a secret");
```
ttomsu: Remove comma

PR #305, opened 2018-07-26
fix(github): Fail github authentication when header is absent (#302)

ezimanyi:
The logic to check the authentication of github triggers is incorrect;
if a trigger is configured with a secret, but the github webhook does
not contain the X-Hub-Signature header, we incorrectly allow the
trigger to fire. We should instead fail the authentication and
not fire the trigger.


ezimanyi:
Cherry-pick into 1.6 as well as it's still supported.

PR #310, opened 2018-08-09
fix(scheduler): Prevent cron parse err from aborting compensation batch

asher:


echo-scheduler/src/main/groovy/com/netflix/spinnaker/echo/scheduler/actions/pipeline/impl/MissedPipelineTriggerCompensationJob.groovy
```diff
@@ -197,11 +198,16 @@ class MissedPipelineTriggerCompensationJob implements ApplicationListener<Contex
...
+        expr.timeZone = dateContext.timeZone
 
-      if (missedExecution(expr, lastExecution, dateContext.triggerWindowFloor, dateContext.now, pipeline)) {
-        pipelineInitiator.call(pipeline)
+        if (missedExecution(expr, lastExecution, dateContext.triggerWindowFloor, dateContext.now, pipeline)) {
+          pipelineInitiator.call(pipeline)
+        }
+      } catch (ParseException e) {
+        log.error("Error parsing cron expression (${trigger.cronExpression}) for pipeline ${pipeline.id}")
+        onOrcaError(e)
```
ajordens: Somewhat annoying that `onOrcaError` is going to dump a stack trace with a rather nondescript `log.error("Error retrieving latest pipeline executions", error)` message.

The metric increment makes sense but maybe we could do without that log output?

echo-scheduler/src/main/groovy/com/netflix/spinnaker/echo/scheduler/actions/pipeline/impl/MissedPipelineTriggerCompensationJob.groovy
```diff
@@ -197,11 +198,16 @@ class MissedPipelineTriggerCompensationJob implements ApplicationListener<Contex
...
+        CronExpression expr = new CronExpression(trigger.cronExpression)
+        expr.timeZone = dateContext.timeZone
 
-      if (missedExecution(expr, lastExecution, dateContext.triggerWindowFloor, dateContext.now, pipeline)) {
-        pipelineInitiator.call(pipeline)
+        if (missedExecution(expr, lastExecution, dateContext.triggerWindowFloor, dateContext.now, pipeline)) {
+          pipelineInitiator.call(pipeline)
+        }
+      } catch (ParseException e) {
+        log.error("Error parsing cron expression (${trigger.cronExpression}) for pipeline ${pipeline.id}")
```
dreynaud: log.error("Error parsing cron expression (${trigger.cronExpression}) for pipeline ${pipeline.id}", e)

asher: Logging a stacktrace here for a java.text.ParseExpression isn't helpful, no longer doing so was deliberate. Having the cron expression that triggers it on the other hand is useful.

PR #311, opened 2018-08-09
fix(cron): Adding fuzzy expression parsing support

robzienert:
Fixing the bug and also dre's comment from other PR.

### DIFF
--- a/echo-scheduler/src/main/groovy/com/netflix/spinnaker/echo/scheduler/actions/pipeline/impl/MissedPipelineTriggerCompensationJob.groovy
+++ b/echo-scheduler/src/main/groovy/com/netflix/spinnaker/echo/scheduler/actions/pipeline/impl/MissedPipelineTriggerCompensationJob.groovy
@@ -17,6 +17,7 @@
 package com.netflix.spinnaker.echo.scheduler.actions.pipeline.impl
 
 import com.google.common.collect.Lists
+import com.netflix.scheduledactions.triggers.CronExpressionFuzzer
 import com.netflix.spectator.api.Registry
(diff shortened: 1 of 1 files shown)


echo-scheduler/src/main/groovy/com/netflix/spinnaker/echo/scheduler/actions/pipeline/impl/MissedPipelineTriggerCompensationJob.groovy
```diff
@@ -199,14 +200,14 @@ class MissedPipelineTriggerCompensationJob implements ApplicationListener<Contex
...
-        CronExpression expr = new CronExpression(trigger.cronExpression)
+        CronExpression expr = getCronExpression(trigger)
         expr.timeZone = dateContext.timeZone
 
         if (missedExecution(expr, lastExecution, dateContext.triggerWindowFloor, dateContext.now, pipeline)) {
           pipelineInitiator.call(pipeline)
         }
       } catch (ParseException e) {
-        log.error("Error parsing cron expression (${trigger.cronExpression}) for pipeline ${pipeline.id}")
+        log.error("Error parsing cron expression (${trigger.cronExpression}) for pipeline ${pipeline.id}", e)
```
asher: Don't think logging a stacktrace here is worth the bits.

dreynaud: At least the error message then if not the full stack trace? I have a burning hatred for error logs that don't show the actual exception

PR #312, opened 2018-08-13
feat(trigger): provide link to triggering event and link text

emjburns:
In the Pipeline UI, we have great support for linking back to a build job. This support is not great when the event is not a build, because we hardcode "Build #X" in the UI, and scrape the url from `trigger.buildInfo.url`.

screen shot 2018-08-13 at 11 50 22 am

This PR provides two fields in the trigger:
- trigger.link = the link to the event that caused this trigger
- trigger.linkText = the text to display for a trigger. There is a UI refactor coming (@erikmunson ) that will allow triggers to provided their own logic, and this field will be removed, but for now you can build the correct text and provide it when you build the trigger, and the UI will soon display that.

This PR pulls the build URL for jenkins triggers as a first step.

@tomaslin can you think of any problems with adding a field to the `BuildEvent` object? This field is already present in the jenkins payload we have, just surfacing it in Echo.

### DIFF
--- a/echo-model/src/main/groovy/com/netflix/spinnaker/echo/model/trigger/BuildEvent.groovy
+++ b/echo-model/src/main/groovy/com/netflix/spinnaker/echo/model/trigger/BuildEvent.groovy
@@ -49,6 +49,7 @@ class BuildEvent extends TriggerEvent {
     boolean building
     int number
     Result result
+    String url
     List<Artifact> artifacts
   }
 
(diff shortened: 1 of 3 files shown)


tomaslin:
as long as there is a good enough fallback when the fields don't exist, it should be fine. I would double check the travis integration because I believe the same type is used across both jenkins and travis

erikmunson:
v nice. my guess is that we'll just stick with our existing logic as a fallback for now since it seems to work ok for most triggers. can always pick something better later.

emjburns:
@ezimanyi @gardleopard do either of you know if this will work for travis? specifically the changes to the `Build` event and `BuildEventMonitor`. Is there anything I need to do to pull that link successfully out of the travis payload?

ezimanyi:
I think that the Travis client is setting the URL [here](https://github.com/spinnaker/igor/blob/2aa1f1469e6afd236034bde4116f392c5f0ad6e9/igor-web/src/main/java/com/netflix/spinnaker/igor/travis/service/TravisBuildConverter.java#L33).  If I'm reading things right, it *should* be threaded all the way through and just work with your updates, but @gardleopard knows this code better than me.

PR #318, opened 2018-08-15
refactor(core): Convert echo-model to Java

ezimanyi:


ezimanyi:
I was getting compilation errors trying to use a Groovy class within a Java class (I believe because all Java gets compiled before all Groovy).  I know this can be fixed in the gradle config, but it also was reasonably low effort to just move this entire module to Java, which is the what we'd like for new code anyway.

The diff was really confusing when I submitted this as one commit (mixing up which files were moved onto which) so I made one commit per file converted to Java.

PR #319, opened 2018-08-16
feat(pubsub/google): GCB message format added

lwander:
image


ttomsu:
Awesome! Two interesting notes from the screenshot:

1. Should "Version" be relabeled to "Digest" ? I associate "version" with the commit hash, but maybe that's just me.
1. "1 received artifacts were not consumed" - should be "1 received artifact was not consumed", but also, why was it not consumed?

Otherwise SHIP IT! :shipit:

PR #323, opened 2018-08-17
fix(pubsub): echo starts up when pubsub not enabled

emjburns:
Make `redisClientSelector` optional so that if pubsub is not enabled echo actually starts up. 
Throw exception when redis is not enabled but pubsub is.


robzienert:
reviewed IRL

PR #325, opened 2018-08-21
feat(pubsub): add sqs msgattr parsing, optional id in msgattrs

emjburns:
When you set message attributes on an SNS message, and then an SQS queue pulls that message, the message attributes aren't translated to the SQS message attributes field. Instead, they're in a `MessageAttributes` map on the message body. 

Adding the ability to parse sns message attributes and add them to the list of message attributes.

In order to dedupe client messages as well as sqs messages, I'm allowing you to provide an id to dedupe with. This will be present in a message attribute field with key `id`, like below:
```
pubsub:
  amazon:
    enabled: true
    subscriptions:
    - name: echo_events
      topicARN: arn:aws:sns....
      queueARN: arn:aws:sqs....
      templatePath: /sqs.jinja
      idInMessageAttributes: true #this is the new, optional field
```
This will still work with at least once delivery because any message delivered more than once will have the same content.

### DIFF
--- a/echo-pubsub-aws/src/main/java/com/netflix/spinnaker/echo/config/AmazonPubsubProperties.java
+++ b/echo-pubsub-aws/src/main/java/com/netflix/spinnaker/echo/config/AmazonPubsubProperties.java
@@ -52,9 +52,14 @@ public class AmazonPubsubProperties {
 
     private MessageFormat messageFormat;
 
+    private boolean idInMessageAttributes;
+
     int visibilityTimeout = 30;
     int sqsMessageRetentionPeriodSeconds = 120;
     int waitTimeSeconds = 5;
+
+    // 1 hour default
+    private Long dedupeRetentionMillis = 3600000L;
 
     public AmazonPubsubSubscription() {
     }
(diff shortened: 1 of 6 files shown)


echo-pubsub-aws/src/main/java/com/netflix/spinnaker/echo/config/AmazonPubsubProperties.java
```diff
@@ -64,12 +69,19 @@ public AmazonPubsubSubscription(
...
+      boolean idInMessageAttributes,
+      Long dedupeRetentionMillis
+      ) {
       this.name = name;
       this.topicARN = topicARN;
       this.queueARN = queueARN;
       this.templatePath = templatePath;
       this.messageFormat = messageFormat;
+      this.idInMessageAttributes = idInMessageAttributes;
+      if (dedupeRetentionMillis != null && dedupeRetentionMillis >= 0) {
```
robzienert: I think if you're doing constraint checking, you should throw an exception, or at minimum log that the configuration is invalid, rather than silently dropping invalid config.

echo-pubsub-aws/src/main/java/com/netflix/spinnaker/echo/pubsub/aws/SQSSubscriber.java
```diff
@@ -219,8 +231,27 @@ private String unmarshallMessageBody(String messageBody) {
       // Try to unwrap a notification message; if that doesn't work,
       // we're dealing with a message we can't parse. The template or
       // the pipeline potentially knows how to deal with it.
-      log.error("Unable unmarshal NotificationMessageWrapper. Unknown message type. (body: {})", messageBody, e);
+      log.error("Unable unmarshall NotificationMessageWrapper. Unknown message type. (body: {})", messageBody, e);
```
robzienert: It was actually spelled the right way originally.

echo-pubsub-aws/src/main/java/com/netflix/spinnaker/echo/config/AmazonPubsubProperties.java
emjburns: Okie

PR #326, opened 2018-08-21
chore(metrics): add cron pipeline started metric

emjburns:
Apologies about fixing the formatting. 

Only change here is wiring in the `Registry` and then publishing a metric:

```java
Id id = registry.createId("pipelines.triggered")
          .withTag("monitor", getClass().getSimpleName())
          .withTag("application", Optional.ofNullable(pipeline.getApplication()).orElse("NULL_APPLICATION"))
        registry.counter(id).increment()
```

### DIFF
--- a/echo-scheduler/src/main/groovy/com/netflix/spinnaker/echo/scheduler/actions/pipeline/PipelineTriggerAction.groovy
+++ b/echo-scheduler/src/main/groovy/com/netflix/spinnaker/echo/scheduler/actions/pipeline/PipelineTriggerAction.groovy
@@ -18,6 +18,8 @@ package com.netflix.spinnaker.echo.scheduler.actions.pipeline
 import com.netflix.scheduledactions.ActionSupport
 import com.netflix.scheduledactions.Context
 import com.netflix.scheduledactions.Execution
+import com.netflix.spectator.api.Id
+import com.netflix.spectator.api.Registry
 import com.netflix.spinnaker.echo.pipelinetriggers.PipelineCache
 import com.netflix.spinnaker.echo.pipelinetriggers.orca.PipelineInitiator
 import com.netflix.spinnaker.echo.scheduler.actions.ActionDependencies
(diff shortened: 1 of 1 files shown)


echo-scheduler/src/main/groovy/com/netflix/spinnaker/echo/scheduler/actions/pipeline/PipelineTriggerAction.groovy
```diff
@@ -27,21 +29,33 @@ import org.slf4j.Logger
...
+      def pipelineInitiator = (PipelineInitiator) ActionDependencies.getBean(PipelineInitiator)
+      def pipelineCache = (PipelineCache) ActionDependencies.getBean(PipelineCache)
+      def pipeline = PipelineTriggerConverter.fromParameters(pipelineCache, context.parameters)
+
+      LOGGER.info("Executing PipelineTriggerAction for '${pipeline}'...")
+      pipelineInitiator.call(pipeline)
+
+      Id id = registry.createId("pipelines.triggered")
+        .withTag("monitor", getClass().getSimpleName())
+        .withTag("application", Optional.ofNullable(pipeline.getApplication()).orElse("NULL_APPLICATION"))
```
dreynaud: minor nitpick: I would go with just `"null"` instead of `"NULL_APPLICATION"`, if we want to query for it `application,null,:eq` wouldn't be too bad

echo-scheduler/src/main/groovy/com/netflix/spinnaker/echo/scheduler/actions/pipeline/PipelineTriggerAction.groovy
```diff
@@ -27,21 +29,33 @@ import org.slf4j.Logger
...
+      registry.counter(id).increment()
+
+      LOGGER.info("Successfully executed PipelineTriggerAction for '${pipeline}")
+    } catch (Exception e) {
+      LOGGER.error("Exception occurred while executing PipelineTriggerAction", e)
```
dreynaud: while we're at it, let's have a metric for pipeline trigger errors if you don't mind

echo-scheduler/src/main/groovy/com/netflix/spinnaker/echo/scheduler/actions/pipeline/PipelineTriggerAction.groovy
emjburns: I wanted to make it clear that we were actually setting it to a value, vs null, because apparently spectator/atlas barfs on null values. Seems ok to name it something explicit (even if it's excessive caps usage) to avoid any debugging confusion of looking for a null value and finding a string "null".

echo-scheduler/src/main/groovy/com/netflix/spinnaker/echo/scheduler/actions/pipeline/PipelineTriggerAction.groovy
emjburns: :D I checked no errors have been seen in the last week. You're right, though, probably doesn't hurt to have a metric

echo-scheduler/src/main/groovy/com/netflix/spinnaker/echo/scheduler/actions/pipeline/PipelineTriggerAction.groovy
dreynaud: `"null"` is a pretty good string for `null` IMO, no need to get creative

dreynaud: 🙏

PR #327, opened 2018-08-23
fix(pubsub): don't infinitely nack duplicate messages

emjburns:
For pubsub, `nack()`-ing tells the system to redeliver the message (optionally after some time period).
We store only one key in redis about whether someone is currently processing a message. Timeout is 7 days for google for this.
 
If we get a duplicate message, the current implementation will continue to `nack()` the message (and cause it to be redelivered) until either the system stops doing that, or the timeout expires and the duplicate will be processed.

This PR introduces a second redis key of 'completed' messages with the message body as the value. Before processing a new message, we check to see if the message was already processed (`messageComplete(completedKey, payload)`). If it was processed, we `ack()` the message and move on - no need to process the same message more than once.

If the message was not fully processed, we try to get the lock. If we get the lock, we process the message and then store the message in the completed map (`setMessageComplete(...)`).

If we don't get the lock, we `nack` the message. When the message is re-delivered, we try again.

Also emitting metrics for number of duplicates, for funsies.


echo-pubsub-core/src/main/java/com/netflix/spinnaker/echo/pubsub/PubsubMessageHandler.java
```diff
@@ -112,11 +122,29 @@ private void setMessageHandled(String messageKey, String identifier, Long retent
...
   }
 
+  private void setMessageComplete(String messageKey, String payload, Long retentionDeadlineMillis) {
+    redisClientDelegate.withCommandsClient(c -> {
+      c.psetex(messageKey, retentionDeadlineMillis, payload);
+    });
+  }
+
+  private Boolean messageComplete(String messageKey, String value) {
+    Boolean exists = redisClientDelegate.withCommandsClient(c -> {
```
jtk54: I think this is not safely concurrent between Echo instances. Reads are non-blocking, so multiple Echoes could attempt to write the 'completed' key here, and the behavior afterwards is not clear.

emjburns: Ah good catch. Fixed.

echo-pubsub-core/src/main/java/com/netflix/spinnaker/echo/pubsub/PubsubMessageHandler.java
```diff
@@ -79,10 +81,18 @@ public void handleMessage(MessageDescription description,
       throw new IllegalStateException("Redis not enabled, pubsub requires redis. Please enable.");
     }
 
-    String messageKey = makeKey(description, messageId);
-    if (tryAck(messageKey, description.getAckDeadlineMillis(), acknowledger, identifier)) {
+    String completeKey = makeCompletedKey(description, messageId);
+    if (messageComplete(completeKey, description.getMessagePayload())) {
```
jtk54: I don't think there is a bound on the size of message payloads. Is there a particular reason we're storing the entire payload here rather than using hash or just an operation on the key itself?

echo-pubsub-core/src/main/java/com/netflix/spinnaker/echo/pubsub/PubsubMessageHandler.java
```diff
@@ -141,4 +166,10 @@ private void processEvent(MessageDescription description) {
     event.setDetails(details);
     pubsubEventMonitor.processEvent(event);
   }
+
+  Id getDuplicateMetricId(MessageDescription messageDescription) {
```
jtk54: I'm a fan of this, would like to see more metrics for various other paths here, like number of messages nacked, etc. Would give us confidence that we're not being inefficient anywhere.

echo-pubsub-core/src/main/java/com/netflix/spinnaker/echo/pubsub/PubsubMessageHandler.java
emjburns: @jtk54 I was thinking that storing the message payload would be a good way to dedupe (if ids were the same but payload wasn't the same), but after thinking about it more it doesn't seem necessary. I'll accept that id should be unique in the retention period for popper behavior. What about storing the timestamp the message was complete, and then only checking if the key exists to see if the message was processed?

echo-pubsub-core/src/main/java/com/netflix/spinnaker/echo/pubsub/PubsubMessageHandler.java
emjburns: Ah, I have those metrics for the amazon message handlers, but I'll bump some up to this class.

echo-pubsub-core/src/main/java/com/netflix/spinnaker/echo/pubsub/PubsubMessageHandler.java
robzienert: A crc32 hash of the message would probably do just fine if you don't want to switch it up much.

PR #335, opened 2018-08-27
chore(pubsub): switch to seconds and setex redis command

emjburns:
We use Dynomite internally, and the `psetex` command is not implemented. Switching to seconds here, and `setex` (which is supported for us). Sound ok?

### DIFF
--- a/echo-pubsub-core/src/main/java/com/netflix/spinnaker/echo/pubsub/PubsubMessageHandler.java
+++ b/echo-pubsub-core/src/main/java/com/netflix/spinnaker/echo/pubsub/PubsubMessageHandler.java
@@ -51,7 +51,7 @@ public class PubsubMessageHandler {
   private final Registry registry;
 
   private static final String SET_IF_NOT_EXIST = "NX";
-  private static final String SET_EXPIRE_TIME_MILLIS = "PX";
+  private static final String SET_EXPIRE_TIME_SECONDS = "EX";
   private static final String SUCCESS = "OK";
   @Autowired
   public PubsubMessageHandler(PubsubEventMonitor pubsubEventMonitor,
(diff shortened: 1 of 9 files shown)


echo-pubsub-google/src/main/java/com/netflix/spinnaker/echo/pubsub/google/GooglePubsubSubscriber.java
```diff
@@ -187,8 +187,8 @@ public void receiveMessage(PubsubMessage message, AckReplyConsumer consumer) {
         .messagePayload(messagePayload)
         .messageAttributes(messageAttributes)
         .pubsubSystem(pubsubSystem)
-        .ackDeadlineMillis(5 * TimeUnit.SECONDS.toMillis(ackDeadlineSeconds)) // Set a high upper bound on message processing time.
-        .retentionDeadlineMillis(TimeUnit.DAYS.toMillis(7)) // Expire key after max retention time, which is 7 days.
+        .ackDeadlineSeconds(5 * ackDeadlineSeconds) // Set a high upper bound on message processing time.
+        .retentionDeadlineSeconds(7 * 24 * 60 * 60) // Expire key after max retention time, which is 7 days.
```
jtk54: Why not `DAYS.toSeconds()`?

echo-pubsub-core/src/main/java/com/netflix/spinnaker/echo/pubsub/PubsubMessageHandler.java
```diff
@@ -52,6 +52,7 @@
 
   private static final String SET_IF_NOT_EXIST = "NX";
   private static final String SET_EXPIRE_TIME_MILLIS = "PX";
```
jtk54: Constant needed now?

echo-pubsub-google/src/main/java/com/netflix/spinnaker/echo/pubsub/google/GooglePubsubSubscriber.java
emjburns: Good questions. Because that returns a long, and instead of converting the value I just want to use math. The "7 days" comment next to it should explain enough.

echo-pubsub-core/src/main/java/com/netflix/spinnaker/echo/pubsub/PubsubMessageHandler.java
emjburns: Removing!

echo-pubsub-google/src/main/java/com/netflix/spinnaker/echo/pubsub/google/GooglePubsubSubscriber.java
jtk54: SGTM.

PR #338, opened 2018-09-13
don't try to compensate for crons that should have run very very recently

robfletcher:
Sometimes the compensation job checks so close to the time the job actually starts that it ends up running a duplicate execution.


emjburns:
LGTM

PR #340, opened 2018-09-17
fix(ha): Rename echo-slave.yml to echo-replica.yml.

joonlim:


joonlim:
@jtk54 @lwander Renaming 'echo-slave' to 'echo-replica'

PR #341, opened 2018-09-19
Add echo manual trigger handler

ezimanyi:


ezimanyi:
@ajordens : Thanks for the comments!

After digging through the security code to learn more about auth, it looks the `X-SPINNAKER-USER` issue would be reasonably easy to resolve by forcing Echo to propagate the authentication context of the manual trigger along to Orca.  (For non-manual triggers, it's already looking at the value set on the trigger to pass along, so we could just force it to be whoever initiated the action here.)

It's not as clear to me how to address `X-SPINNAKER-ACCOUNTS`, though something similar would probably work (at the expense of complicating Echo a bit as it looks like it has no need to be aware of this header at all at this point).  Basically, just forward along any value of that header that was received from Gate.

I'm not sure I understand the comment about adding authorization checks against the application (though this is likely that I'm still new to how our auth logic works).  I don't see any checks in the older flow.  If you hit the gate endpoint `/{application}/{pipeline}` (handled by `invokePipelineConfig`) it will eventually call Orca's `/orchestrate` without (as far as I can tell) any read auth checks happening.  (Of course write checks would happen before clouddriver could actually do anything.)

As a more general question---is this crazy to try to make manual triggers go through Echo?  My goal here is to have a more consistent code path so that things we add to Echo that augment triggers (ex: by extracting artifacts from the triggering context) don't need to be special-cased when a pipeline is manually triggered and echo can just uniformly prepare the trigger for execution by Orca in all cases.  That being said, if you think this will overly-complicate things like auth (or otherwise lead to a non-intuitive flow) I can consider other ways of achieving the same goal of deduplicating code.

ajordens:
@ezimanyi I think the difference in the `echo` case is that the pipelines have been fetched in the background (with no auth checks) -- vs in `gate` where the pipeline being invoked is fetched directly from `front50` and subject to an authorization check. 

I think we'll be in reasonable shape as long as the `X-...` headers are passed along (by default they should be but it's worth double checking).

I'm not against centralizing some of this within `echo` _but_ we just need to be conscious of what impact that has on authorization, I don't _think_ there is anything beyond addressing the READ check against the matching pipeline's application.

ezimanyi:
I've now addressed the authentication issue:
* We now check that the user has permissions to read the application in Gate before forwarding the request on to Echo
* We explicitly set the `runAsUser` to the authenticated user in Gate, and thread it through the `ManualEventMonitor` in echo so that it gets included as `X-SPINNAKER-USER` in the request to Orca.

I have not yet in the change threaded through `X-SPINNAKER-ACCOUNTS`---it should be reasonably easy to pass it through in the same way that we're dealing with `X-SPINNAKER-USER` (read it in Gate, add it to the event, and have Echo send the value in a header to Orca).  Before doing that I wanted to check if this is something being actively used or if we can deprecate it.  It seems from the above message that Netflix moved away from it?  If so, we can check if other users are using it.

Mind taking another look, @lwander @ajordens .  Thanks!

ezimanyi:
Also https://github.com/spinnaker/gate/pull/600 updated with changes in the above comment

echo-model/src/main/java/com/netflix/spinnaker/echo/model/Trigger.java
```diff
@@ -119,6 +125,9 @@ public String toString() {
   @JsonIgnore
   Pipeline parent;
 
+  @JsonIgnore
+  boolean propagateAuth;
```
ajordens: There's no fear of anyone being able to explicitly define this in their pipeline trigger json?

echo-pipelinetriggers/src/main/java/com/netflix/spinnaker/echo/pipelinetriggers/monitor/ManualEventMonitor.java
```diff
@@ -0,0 +1,123 @@
...
+    return true;
+  }
+
+  @Override
+  protected Predicate<Trigger> matchTriggerFor(final TriggerEvent event, final Pipeline pipeline) {
+    ManualEvent manualEvent = (ManualEvent) event;
+    String application = manualEvent.getContent().getApplication();
+    String nameOrId = manualEvent.getContent().getPipelineNameOrId();
+
+    return trigger -> pipeline.getApplication().equals(application) && (
```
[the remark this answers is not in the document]
ezimanyi: I'd left it as is to match the current behavior, which is case-sensitive.  (Currently when Gate looks up the pipeline you specified from front50, it will 404 unless there's a case-sensitive match.)

echo-pipelinetriggers/src/main/java/com/netflix/spinnaker/echo/pipelinetriggers/monitor/ManualEventMonitor.java
```diff
@@ -0,0 +1,123 @@
...
+
+  @Override
+  protected Function<Trigger, Pipeline> buildTrigger(Pipeline pipeline, TriggerEvent event) {
+    ManualEvent manualEvent = (ManualEvent) event;
+    return trigger -> {
+      String runAsUser = ((ManualEvent) event).getContent().getRunAsUser();
+      List<Map<String, Object>> notifications = buildNotifications(pipeline.getNotifications(),
+        manualEvent.getContent().getTrigger().getNotifications());
+      return pipeline
+        .withTrigger(manualEvent.getContent().getTrigger().atRunAsUser(runAsUser))
```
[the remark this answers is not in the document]
ezimanyi: Great point, I forgot to take this out here, will fix.  (This doesn't do anything now that I removed `runAsUser` from gate.)

echo-model/src/main/java/com/netflix/spinnaker/echo/model/Trigger.java
ezimanyi: This is a valid concern; the `@JsonIgnore` is not actually causing the field to be ignored in deserialization.  I'm going to figure out why that's happening, and ensure that any value in the JSON gets ignored.

ajordens: 👍 

My preference would be to not have anything (externally settable) on the Trigger influence how auth is propagated.

ezimanyi: OK, configured the builder/deserializer to ignore that field, and added a test to ensure that `fropagateAuth` is always false after deserialization.

echo-core/src/main/groovy/com/netflix/spinnaker/echo/events/EventPropagator.groovy
```diff
@@ -40,11 +44,15 @@ class EventPropagator {
 
     void processEvent(Event event) {
         Observable.from(listeners)
+            .map({listener ->
```
lwander: Missing space after `{` here and below

PR #343, opened 2018-09-24
refactor(scheduler): Remove cassandra scheduler

dibyom:


dibyom:
Addresses https://github.com/spinnaker/spinnaker/issues/3378

PR #344, opened 2018-09-26
feat(echo): new notification type: Github status check

amanya:
This is a new type of notification which sets Github checks of an execution of a pipeline or stage.

image


echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/notification/GithubNotificationAgent.groovy
```diff
@@ -0,0 +1,126 @@
...
+    } else if (config.type == 'pipeline') {
+      description = """Pipeline '${pipeline}' is ${status}"""
+      context = "Pipeline"
+
+      targetUrl = """${spinnakerUrl}/#/applications/${application}/${event.content?.execution?.id}"""
+    } else {
+      return
+    }
+
+    log.info('Sending Github status check: {} {} {} {} {} {}',
```
ajordens: the structured `kv` usage here seems unnecessary ... the only thing that seems relevant to me would be `application` and it's not currently included.

echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/notification/GithubNotificationAgent.groovy
```diff
@@ -0,0 +1,126 @@
...
+    try {
+
+      githubService.updateCheck("""token ${token}""", repo, sha, new GithubStatus(
+        state: state,
+        target_url: targetUrl,
+        description: description,
+        context: context))
+
+    } catch (Exception e) {
+      log.error('failed to send github status ', e)
```
ajordens: This should include a lot more context about what failed ... at least the pipeline / application.

echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/api/Notification.groovy
```diff
@@ -41,7 +41,8 @@ class Notification {
     PAGER_DUTY,
     JIRA,
     BEARYCHAT,
-    GOOGLECHAT
+    GOOGLECHAT,
+    GITHUB
```
ajordens: Is `GITHUB` too generic of a notification type?

Would we ever want to do something other than set a status check? (add a comment?, etc.)

On the flip-side, it seems possible to use `GITHUB` and a secondary discriminator to differentiate on what action wants to be taken?

amanya:
@ajordens @lwander I've addressed your comments and converted it to Java, PTAL

echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/notification/GithubNotificationAgent.java
```diff
@@ -0,0 +1,110 @@
+package com.netflix.spinnaker.echo.notification;
```
lwander: Missing license header

echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/notification/EventContent.java
```diff
@@ -0,0 +1,133 @@
+/*
+ * Copyright 2015 Netflix, Inc.
```
lwander: Was this added by you or translated from a netflix groovy class?

echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/github/GithubStatus.java
```diff
@@ -0,0 +1,32 @@
+package com.netflix.spinnaker.echo.github;
```
lwander: Missing license header

echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/notification/EventContent.java
amanya: It was added by me to "hide" the ugly Map navigations in the Event in Java. Maybe you have some suggestion on how to do it nicer?

I'm fixing the license

amanya:
@lwander @ajordens @gardleopard 

Some CI systems (Travis) do a simulation of merging the commit in the PR into the default branch. If we trigger the pipeline using the `pull_request_master` trigger, the sha we get in Spinnaker is the one corresponding to that commit, so if we set the status check in that commit, we won't be able to see it in the pull request in GitHub.

To detect the real commit, we look at the message of the commit and see if it matches the string `Merge ${branch_sha} into ${master_sha}` and if it does we take the real commit.

This is a usecase we face but I'm not entirely sure that this is generic enough to go there, opinions @ajordens @lwander @gardleopard ?

PR #350, opened 2018-10-03
style(core): Add comments and re-order fields in Trigger.java

ezimanyi:
This change does not change anything functionally, but adds some structure to the long list of fields in Trigger.java so it's easier to understand what each field is used for.

### DIFF
--- a/echo-model/src/main/java/com/netflix/spinnaker/echo/model/Trigger.java
+++ b/echo-model/src/main/java/com/netflix/spinnaker/echo/model/Trigger.java
@@ -61,7 +61,63 @@ public class Trigger {
 
   private static final Logger log = LoggerFactory.getLogger(Trigger.class);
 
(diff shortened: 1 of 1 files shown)


ezimanyi:
I did this re-ordering and comment-adding as I was trying to understand this file, then realized it might be helpful to commit it.  At some point we may want to consider making different subclasses for different trigger types (and possibly separating out trigger configuration from what is bound at run-time)---but for now I think this will make this file easier to read.

emjburns:
LGTM - if you are ever thinking about subclassing Trigger please think about how it would play with a custom private trigger!

ezimanyi:
@emjburns : Yes, I'll definitely consider that, though don't have any immediate plans to do that subclassing.  (I've convinced myself that there are higher priority things to do with triggers for now.)

PR #353, opened 2018-10-04
feat(cron): only query execution history for eligible triggers

dreynaud:
... instead of all triggers, this should restrict the query to triggers
that have a valid fire time in the compensation window.


echo-scheduler/src/main/groovy/com/netflix/spinnaker/echo/scheduler/actions/pipeline/impl/MissedPipelineTriggerCompensationJob.groovy
```diff
@@ -234,6 +240,13 @@ class MissedPipelineTriggerCompensationJob implements ApplicationListener<Contex
     return new CronExpression(trigger.cronExpression)
   }
 
+  private static Date getLastValidTimeInWindow(String expr, DateContext dateContext) {
+    return getLastValidTimeInWindow(
+      new CronExpression(expr),
```
robzienert: You should pass the expr through `CronExpressionFuzzer` here.

PR #354, opened 2018-10-04
 fix(core): Ensure event processing occurs in the same thread

ezimanyi:
Two changes:
* Change the fix I made in https://github.com/spinnaker/echo/pull/274 so that we block waiting for the initialization of the pipeline cache instead of having RxJava defer the operation and perform it on the pipeline cache thread.  This is necessary to properly pass authentication headers through Echo.
* Remove some RxJava from `TriggerMonitor.java`

Keeping these as separate commits as I may want to pull the fix into 1.10 so that my new manual trigger workflow works there without a race condition on startup.

### DIFF
--- a/echo-pipelinetriggers/src/main/java/com/netflix/spinnaker/echo/pipelinetriggers/monitor/ManualEventMonitor.java
+++ b/echo-pipelinetriggers/src/main/java/com/netflix/spinnaker/echo/pipelinetriggers/monitor/ManualEventMonitor.java
@@ -34,7 +34,6 @@ import lombok.NonNull;
 import lombok.extern.slf4j.Slf4j;
 import org.springframework.stereotype.Component;
 import rx.functions.Action1;
-import rx.functions.Func1;
 
 /**
  * Triggers pipelines in _Orca_ when a user manually starts a pipeline.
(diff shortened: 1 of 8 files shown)


echo-pipelinetriggers/src/main/java/com/netflix/spinnaker/echo/pipelinetriggers/monitor/ManualEventMonitor.java
```diff
@@ -61,19 +60,18 @@ public ManualEventMonitor(@NonNull PipelineCache pipelineCache,
...
-        ManualEvent manualEvent = (ManualEvent) event;
-        Trigger trigger = manualEvent.getContent().getTrigger();
-        return Stream.of(trigger)
-          .filter(matchTriggerFor(event, pipeline))
-          .findFirst()
-          .map(buildTrigger(pipeline, event));
-      }
-    };
+  @Override
+  protected Optional<Pipeline> withMatchingTrigger(final TriggerEvent event, Pipeline pipeline) {
```
lwander: Nice, glad you took this out of the function

echo-pipelinetriggers/src/main/java/com/netflix/spinnaker/echo/pipelinetriggers/monitor/TriggerMonitor.java
```diff
@@ -97,41 +86,36 @@ protected void onEchoResponse(final TriggerEvent event) {
...
-          .map(withMatchingTrigger(event))
-          .filter(Optional::isPresent)
-          .map(Optional::get)
-          .doOnNext(this::onMatchingPipeline)
-          .subscribe(subscriber, this::onSubscriberError);
-      } else {
-        onEventProcessed(event);
-      }
-    };
+  protected void triggerEachMatchFrom(final TriggerEvent event, List<Pipeline> pipelines) {
```
lwander: Given that we're refactoring this -- can we rename this to something clearer? Maybe `triggerEachMatchingPipeline`?

ezimanyi: Good idea, done

PR #363, opened 2018-10-17
refactor(core): Convert echo-web to Java

ezimanyi:


ezimanyi:
Going to be making some changes to this module, and want to avoid writing more Groovy code (or mixing Java/Groovy in the module).  This change is refactor-only, new changes to come in a separate PR.

ezimanyi:
Yeah, I realize it's fine to have Java + Groovy, but in tiny modules like this I like to clean it up.

PR #368, opened 2018-10-19
refactor(core): Autowire PipelineInitiator

ezimanyi:


ezimanyi:
Small, reasonably mechanical step in removing some RxJava from `echo-pipelinetriggers`.  Rather than autowiring an interface of `Action1<Pipeline>` autowire a `PipelineInitiator` which makes it more clear what the purpose of the injected class is.  Similarly, change the interface of the class from `call` to `startPipeline` so it's more clear what the class is doing.

echo-pipelinetriggers/src/main/java/com/netflix/spinnaker/echo/pipelinetriggers/orca/PipelineInitiator.java
```diff
@@ -55,8 +55,7 @@ public void initialize() {
     }
   }
 
-  @Override
-  public void call(Pipeline pipeline) {
+  public void startPipeline(Pipeline pipeline) {
```
emjburns: Love this.

dreynaud:
Nice improvement, thanks @ezimanyi

PR #369, opened 2018-10-22
 refactor(core): Split pipeline matching and triggering logic

ezimanyi:
I've split this into three self-contained commits that should be easier to read one-by-one than all at once.  The goal here was to reduce the amount of work the `EventMonitor` classes were doing but deferring the actual pipeline matching logic away.

This is in preparation for adding post-processing logic that can inject build information by querying igor (whether or not the build was triggered by igor) and for adding the ability to extract artifacts using a Jinja template.


echo-pipelinetriggers/src/main/java/com/netflix/spinnaker/echo/pipelinetriggers/eventhandlers/BaseTriggerEventHandler.java
```diff
@@ -0,0 +1,81 @@
...
+      } catch (Exception e) {
+        onSubscriberError(e);
+        return Optional.empty();
+      }
+    }
+  }
+
+  private void onSubscriberError(Throwable error) {
+    log.error("Subscriber raised an error processing pipeline", error);
+    registry.counter("trigger.errors").increment();
```
dreynaud: Maybe add an error tag with some information about the error?

echo-pipelinetriggers/src/main/java/com/netflix/spinnaker/echo/pipelinetriggers/eventhandlers/GitEventHandler.java
```diff
@@ -79,13 +76,12 @@ protected boolean isValidTrigger(final Trigger trigger) {
...
   @Override
-  protected boolean isSuccessfulTriggerEvent(TriggerEvent event) {
+  public boolean isSuccessfulTriggerEvent(GitEvent event) {
     return true;
   }
 
   @Override
-  protected Predicate<Trigger> matchTriggerFor(final TriggerEvent event, final Pipeline pipeline) {
-    GitEvent gitEvent = (GitEvent) event;
+  protected Predicate<Trigger> matchTriggerFor(GitEvent gitEvent, final Pipeline pipeline) {
```
lwander: Why no longer `final`?

echo-pipelinetriggers/src/main/java/com/netflix/spinnaker/echo/pipelinetriggers/eventhandlers/ManualEventMonitor.java
```diff
@@ -40,48 +37,42 @@
...
     return objectMapper.convertValue(event, ManualEvent.class);
   }
 
-  public ManualEventMonitor(@NonNull PipelineCache pipelineCache,
-    @NonNull PipelineInitiator pipelineInitiator,
-    @NonNull Registry registry) {
-    super(pipelineCache, pipelineInitiator, registry);
+  public List<Pipeline> getMatchingPipelines(ManualEvent manualEvent, List<Pipeline> pipelines) {
+    return pipelines.stream()
+      .map(p -> withMatchingTrigger(manualEvent, p))
```
lwander: Just curious, why isn't this logic in a base class? It feels like this "find & return matching pipelines" will be shared by all event monitors.

echo-pipelinetriggers/src/main/java/com/netflix/spinnaker/echo/pipelinetriggers/monitor/TriggerEventListener.java
```diff
@@ -0,0 +1,55 @@
+/*
+ * Copyright 2018 Netflix, Inc.
```
lwander: Wrong header?

echo-pipelinetriggers/src/main/java/com/netflix/spinnaker/echo/pipelinetriggers/monitor/TriggerEventListener.java
```diff
@@ -0,0 +1,55 @@
...
+@Slf4j
+@Component
+public class TriggerEventListener implements EchoEventListener {
+  private final List<TriggerMonitor> triggerMonitors;
+
+  public TriggerEventListener(@NonNull PipelineCache pipelineCache,
+    @NonNull PipelineInitiator pipelineInitiator,
+    @NonNull Registry registry,
+    @NonNull List<TriggerEventHandler<?>> eventHandlers
+  ) {
```
lwander: The general rule of thumb for these parens is: "the closing paren is preceded by a newline i.f.f. the opening paren is followed by a newline".

echo-pipelinetriggers/src/main/java/com/netflix/spinnaker/echo/pipelinetriggers/eventhandlers/GitEventHandler.java
```diff
@@ -43,27 +42,25 @@
...
 @Slf4j
-public class GitEventMonitor extends TriggerMonitor {
-
-  public static final String GIT_TRIGGER_TYPE = "git";
+public class GitEventHandler extends BaseTriggerEventHandler<GitEvent> {
+  private static final String GIT_TRIGGER_TYPE = "git";
 
   private static final String GITHUB_SECURE_SIGNATURE_HEADER = "X-Hub-Signature";
 
+  private final ObjectMapper objectMapper = new ObjectMapper();
```
emjburns: Echo has a ton of object mappers everywhere. In other services we _try_ to initialize one object mapper as a bean in the config so that everyone uses the same object mapper. This tends to make it easier to debug serialization issues down the road, although we're not great at actually only having one object mapper. 

As you refactor echo, it might be good to only have one object mapper instead of initializing them everywhere. For this PR, could you just initialize the object mapper in `BaseTriggerEventHandler`?

echo-pipelinetriggers/src/main/java/com/netflix/spinnaker/echo/pipelinetriggers/eventhandlers/ManualEventMonitor.java
emjburns: +1

echo-pipelinetriggers/src/main/java/com/netflix/spinnaker/echo/pipelinetriggers/monitor/TriggerEventListener.java
emjburns: Are you moving to Los Gatos???? :D :D

echo-pipelinetriggers/src/main/java/com/netflix/spinnaker/echo/pipelinetriggers/monitor/TriggerMonitor.java
```diff
@@ -16,143 +16,80 @@
...
-import java.util.regex.Matcher;
-import java.util.regex.Pattern;
 import lombok.NonNull;
 import lombok.extern.slf4j.Slf4j;
-import org.springframework.stereotype.Component;
 
 /**
  * Triggers pipelines on _Orca_ when a trigger-enabled build completes successfully.
  */
-@Component
```
emjburns: Would love more class level comments about the responsibilities of the classes in general. I find them quite helpful.

echo-pipelinetriggers/src/main/java/com/netflix/spinnaker/echo/pipelinetriggers/eventhandlers/GitEventHandler.java
ezimanyi: No specific reason, other than it seemed reasonably low-value to enforce that the function can't re-asssign another pipeline to the local variable `pipeline`.  The code in this module was marking parameters as `final` in a lot (but not all) cases where the parameter was not re-assigned.  Given that it doesn't seem to be the general style in our codebase to mark parameters a final, I leaned towards just removing the cases where things were.  (Though I could see the argument to explicitly put `final` everywhere.)

echo-pipelinetriggers/src/main/java/com/netflix/spinnaker/echo/pipelinetriggers/monitor/TriggerEventListener.java
ezimanyi: Haha, I blame IntelliJ, I'll fix this

echo-pipelinetriggers/src/main/java/com/netflix/spinnaker/echo/pipelinetriggers/eventhandlers/ManualEventMonitor.java
ezimanyi: I think the similarity here is a vestige of when `ManualEventMonitor` had to be a subclass of `TriggerMonitor` (so that it did not need to re-implement all the triggering logic) and thus needed to unnaturally match the paradigm of the other monitors.  It's somewhat different than the other monitors, as it doesn't look at pipeline triggers at all---it just finds whatever pipeline matches the id it's supposed to execute.  I've refactored it to be much clearer now that it doesn't need to follow the pattern of the others.

That being said, it's true that every handler is going to loop over pipelines---so I've pushed the looping part to the base class and have made the implementations just handle deciding if a particular pipeline matches.

echo-pipelinetriggers/src/main/java/com/netflix/spinnaker/echo/pipelinetriggers/eventhandlers/GitEventHandler.java
ezimanyi: Good point; it was bothering me that I needed all of these.  I've done two things:
* I moved the actual object conversion to the base class (which I'd wanted to do all along but was running into type-safety issues when I was trying last week, but found a nice way of doing today)
* I added the `objectMapper` as a constructor parameter of the handlers and autowired it up so we're not creating them ad hoc anymore.

echo-pipelinetriggers/src/main/java/com/netflix/spinnaker/echo/pipelinetriggers/monitor/TriggerMonitor.java
ezimanyi: I'd meant to update the comments; went through and added/fixed class-level comments for things I changed.

emjburns: Great! Thanks.

ezimanyi:
Addressed all the comments; again, tried to separate out larger, multi-file refactors into separate commits to make it easier to read.  The last four commits are new (everything after "refactor(core): Clean up code in event handlers" 0e672c6).  (Old commits were rebased but not changed.

PR #370, opened 2018-10-23
fix(MPT): Fix templated pipeline triggers.

jtk54:
Depends on https://github.com/spinnaker/orca/pull/2480.


echo-pipelinetriggers/src/main/java/com/netflix/spinnaker/echo/pipelinetriggers/orca/PipelineInitiator.java
```diff
@@ -1,24 +1,24 @@
 package com.netflix.spinnaker.echo.pipelinetriggers.orca;
 
+import com.fasterxml.jackson.databind.ObjectMapper;
 import com.netflix.spectator.api.Registry;
 import com.netflix.spinnaker.echo.model.Pipeline;
 import com.netflix.spinnaker.echo.pipelinetriggers.orca.OrcaService.TriggerResponse;
 import com.netflix.spinnaker.fiat.shared.FiatStatus;
 import com.netflix.spinnaker.security.AuthenticatedRequest;
 import com.netflix.spinnaker.security.User;
+import java.util.Map;
```
lwander: We keep java annotations at the bottom & separate

PR #371, opened 2018-10-23
fix(MPT): Fix NPE in pipeline initiator

ezimanyi:


ezimanyi:
Introduced in https://github.com/spinnaker/echo/pull/370; going to merge now as Echo's pretty broken (failing to trigger any pipelines that don't have an explicit type, which is pretty common for non-templated pipelines).

PR #373, opened 2018-10-24
refactor(core): Clean up echo-pipelinetriggers dependencies

ezimanyi:
Reduce transitive dependencies in echo-pipelinetriggers by explicitly specifying dependencies, and remove any unused dependencies.

Explicitly include the following, which were transitively included before:
* echo-model
* korkArtifacts
* jacksonDatabind
* springContext
* commons-lang (pull in lang3, and update calls from lang)

Remove the following, which are not needed:
* guava (unused)
* okHttp (part of retrofitDefault group)
* eurekaClient (unused)
* groovy (only in test module now and included in test group)
* spock (included in test group)

Update two other modules to explicitly pull in logstashEncoder as they were transitively depending on it being in echo-pipelinetriggers

### DIFF
--- a/echo-notifications/echo-notifications.gradle
+++ b/echo-notifications/echo-notifications.gradle
@@ -21,6 +21,7 @@ dependencies {
     spinnaker.group("bootWeb")
     spinnaker.group("retrofitDefault")
     compile spinnaker.dependency("groovy")
+    compile spinnaker.dependency("logstashEncoder")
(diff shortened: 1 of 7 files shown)


ezimanyi:
This was as much an academic exercise for me to understand our dependency management better as anything...don't plan on exhaustively cleaning up all the build scripts but this should at least be beneficial for `echo-pipelinetriggers`.

PR #378, opened 2018-10-26
fix(quietperiods): remove conditional bean

skandragon:


echo-pipelinetriggers/src/main/java/com/netflix/spinnaker/echo/pipelinetriggers/QuietPeriodIndicator.java
```diff
@@ -36,7 +36,6 @@
 
 @Slf4j
 @Component
-@ConditionalOnExpression("${quietPeriod.enabled:false}")
 @Scope(value = ConfigurableBeanFactory.SCOPE_SINGLETON)
```
emjburns: Why do you need `Scope`? `@Component` _should_ do the trick.

emjburns:
LGTM - would appreciate a class level comment as well indicating that by default, it will return false to all inquiries of whether or not we're in a quiet period.

PR #383, opened 2018-10-30
fix(MPT): Fixes templated pipeline triggers from Echo.

jtk54:


jtk54:
Something very bad happened here, don't merge.

jtk54:
Small backport fix was required to fix imports.

echo-pipelinetriggers/src/main/java/com/netflix/spinnaker/echo/pipelinetriggers/orca/PipelineInitiator.java
```diff
@@ -16,9 +20,6 @@
 import rx.functions.Action1;
 import rx.functions.Func1;
 
-import javax.annotation.PostConstruct;
```
ezimanyi: I think moving these up into the rest of the imports is not our usual convention.

PR #389, opened 2018-11-06
refactor(rx): remove RxJava from scheduler and pipeline poller

dreynaud:
* the most consequential changes are in PipelineCache:
  - instead of driving the polling with Observable.interval, use a
    simple executor with a fixed delay instead of a rate, which can
    be problematic when things become latent
  - getPipelines() no longer returns an Observable<List<Pipeline>>, but
    rather exposes 2 APIs that return the cached value directly (and
    optionally blocking until a value is available)
* refocused PipelineCacheSpec on the internal logic rather than the
  scheduling/retrying
* similar change in MissedPipelineTriggerCompensationJob
* as a consequence, we no longer need the Scheduler bean from PipelineTriggerConfiguration
* more cosmetic: we can remove the rx bits in Front50Service, PipelineTriggerConverter,
  PipelineConfigsPollingAgent...


echo-scheduler/src/main/groovy/com/netflix/spinnaker/echo/scheduler/actions/pipeline/impl/PipelineConfigsPollingAgent.groovy
```diff
@@ -79,7 +79,7 @@ class PipelineConfigsPollingAgent extends AbstractPollingAgent {
       log.info("Running the pipeline configs polling agent...")
 
       // Only interested in pipelines that have triggers and that too scheduled triggers
-      def pipelines = pipelineCache.getPipelines().toBlocking().first().findAll {
+      def pipelines = pipelineCache.getPipelines().findAll {
```
ezimanyi: Given that this blocked before, should we be using `getPipelinesSync` here?  (This looks like it will throw an NPE if called before the first polling cycle.)

echo-scheduler/src/main/groovy/com/netflix/spinnaker/echo/scheduler/actions/pipeline/impl/PipelineTriggerConverter.groovy
```diff
@@ -56,7 +56,7 @@ class PipelineTriggerConverter {
       triggerBuilder.runAsUser(parameters.runAsUser)
     }
 
-    def existingPipeline = pipelineCache.getPipelines().toBlocking().first().find { it.id == parameters.id }
+    def existingPipeline = pipelineCache.getPipelines().find { it.id == parameters.id }
```
ezimanyi: Given that this blocked before, should we be using `getPipelinesSync` here?  (This looks like it will throw an NPE if called before the first polling cycle.)

echo-pipelinetriggers/src/main/java/com/netflix/spinnaker/echo/pipelinetriggers/monitor/TriggerMonitor.java
```diff
@@ -67,12 +69,16 @@ private void validateEvent(Event event) {
...
-      pipelineInitiator.startPipeline(p);
-    });
+    try {
+      List<Pipeline> allPipelines = pipelineCache.getPipelinesSync();
+      List<Pipeline> matchingPipelines = eventHandler.getMatchingPipelines(event, allPipelines);
+      matchingPipelines.forEach(p -> {
+        recordMatchingPipeline(p);
+        pipelineInitiator.startPipeline(p);
+      });
+    } catch (TimeoutException e) {
```
ezimanyi: I'm wondering if logging and continuing is the right fallback here?  Ideally we'd probably want to leave the event unprocessed rather than ignore it, but as echo doesn't really have a queuing mechanism, I suppose that failing is probably the only option.

echo-pipelinetriggers/src/main/java/com/netflix/spinnaker/echo/pipelinetriggers/PipelineCache.java
```diff
@@ -101,33 +130,37 @@ public int getPollingIntervalSeconds() {
...
+  @Nullable
+  public List<Pipeline> getPipelines() {
+    return pipelines;
   }
 
-  public List<Pipeline> getPipelinesSync() {
-    return getPipelines().toBlocking().first();
+  @Nonnull
+  public List<Pipeline> getPipelinesSync() throws TimeoutException {
+    return getPipelinesSync(TimeUnit.SECONDS.toMillis(pollingIntervalSeconds));
```
ezimanyi: I'm not sure that `pollingIntervalSeconds` is the correct maximum time to wait here.  In general, this is resolving the case where Echo starts processing events before the `PipelineCache` starts polling.  The time we'd have to wait is then:
```(time between event processor starting and pipeline cache starting) + (time to get a response back to a single poll)``` which is independent of the time between polls.

That being said, reasonable values of the polling interval (~10-60 seconds) are probably also reasonable values for this timeout, so the end result is probably fine.

File: echo-notifications/src/test/groovy/com/netflix/spinnaker/echo/notification/DryRunNotificationAgentSpec.groovy [hunk 5]
```diff
@@ -71,7 +71,7 @@ class DryRunNotificationAgentSpec extends Specification {
 
   def "triggers a pipeline run for a pipeline:complete notification"() {
     given:
-    front50.getPipelines(application) >> Observable.just([pipeline])
+    front50.getPipelines(application) >> [pipeline]
```
ezimanyi: Can now remove the import of `rx.Observable` above.  That's also the last direct reference to `rxJava` in `echo-notifications` so it can be removed from the `echo-notifications.gradle` file.

echo-pipelinetriggers/src/main/java/com/netflix/spinnaker/echo/pipelinetriggers/PipelineCache.java
```diff
@@ -24,70 +25,98 @@
...
   @Autowired
-  public PipelineCache(@NonNull Scheduler scheduler,
+  public PipelineCache(int pollingIntervalSeconds,
+                       @NonNull Front50Service front50,
+                       @NonNull Registry registry) {
+    this(Executors.newSingleThreadScheduledExecutor(), pollingIntervalSeconds, front50, registry);
+  }
+
+  // VisibleForTesting
+  public PipelineCache(ScheduledExecutorService executorService,
```
robzienert: Isn't there a real annotation for this via Guava?

echo-pipelinetriggers/src/main/java/com/netflix/spinnaker/echo/pipelinetriggers/PipelineCache.java
```diff
@@ -24,70 +25,98 @@
...
     this.pollingIntervalSeconds = pollingIntervalSeconds;
     this.front50 = front50;
     this.registry = registry;
+    this.running = false;
+    this.pipelines = null;
+  }
+
+  @PreDestroy
+  public void stop() {
+    running = false;
```
robzienert: Shouldn't we also be doing `executorService.shutdown()` here as well?

echo-pipelinetriggers/src/main/java/com/netflix/spinnaker/echo/pipelinetriggers/PipelineCache.java
```diff
@@ -24,70 +25,98 @@
...
+      return;
+    }
+
+    try {
+      log.debug("Getting pipelines from Front50...");
+      long start = System.currentTimeMillis();
+      pipelines = decorateTriggers(front50.getPipelines());
+
+      lastPollTimestamp = now();
+      registry.counter("front50.requests").increment();
```
robzienert: If we're tracking the timing of this call, why not just `registry.timer` it and change `front50.requests` to a timer?

echo-pipelinetriggers/src/main/java/com/netflix/spinnaker/echo/pipelinetriggers/PipelineCache.java
```diff
@@ -101,33 +130,37 @@ public int getPollingIntervalSeconds() {
...
+          throw new TimeoutException("Pipeline configs are still not available after " + elapsed + "ms");
+        }
 
-  private void onFront50Request(final long tick) {
-    log.debug("Getting pipelines from Front50...");
-    lastPollTimestamp = now();
-    registry.counter("front50.requests").increment();
-  }
+        log.trace("Waiting for initial load of pipeline configs (elapsed={}ms)...", elapsed);
+        Thread.sleep(100);
```
robzienert: Is this worth making configurable?

echo-scheduler/src/main/groovy/com/netflix/spinnaker/echo/scheduler/actions/pipeline/impl/PipelineConfigsPollingAgent.groovy
dreynaud: Yes indeed, good catch

echo-pipelinetriggers/src/main/java/com/netflix/spinnaker/echo/pipelinetriggers/monitor/TriggerMonitor.java
dreynaud: Yes, ideally we wouldn't run into this in the first place. I'd like to find a way to key some kind of health indicator based on the pipeline cache status. E.g. cache not initialized, echo is not "healthy" since it's not ready to do the work.

I'm afraid that queueing would compound ongoing issues and make the system more fragile overall

echo-pipelinetriggers/src/main/java/com/netflix/spinnaker/echo/pipelinetriggers/PipelineCache.java
dreynaud: Yes I just tried to use a semi-reasonable default for callers who don't care too much. The motivation for using `pollingIntervalSeconds` is that it guarantees at least one polling cycle has occurred while we're waiting

echo-notifications/src/test/groovy/com/netflix/spinnaker/echo/notification/DryRunNotificationAgentSpec.groovy
dreynaud: Sweet llamas of the Bahamas!

echo-pipelinetriggers/src/main/java/com/netflix/spinnaker/echo/pipelinetriggers/PipelineCache.java
dreynaud: there is, but I don't see the value of actually pulling a real annotation

echo-pipelinetriggers/src/main/java/com/netflix/spinnaker/echo/pipelinetriggers/PipelineCache.java
dreynaud: we probably should

dreynaud: I think we previously had tests that asserted that we could `stop()` then `start()` again. But given my understanding of `@PreDestroy`, this is more of a one-way street so probably safe to shutdown the executor service here

echo-pipelinetriggers/src/main/java/com/netflix/spinnaker/echo/pipelinetriggers/PipelineCache.java
dreynaud: I guess we could, but if I'm going to change this I'd rather remove this kind of one-off metrics and rely on the http client to publish metrics. If echo picks up the okhttp client from kork it _should_ already be getting some standard metrics for free?

dreynaud:
Thanks a bunch for the feedback @ezimanyi and @robzienert!

PR #392, opened 2018-11-08
chore(swabbie): update subject line

emjburns:
Subject now looks like: 
`[Cleanup] front50 images scheduled for deletion for <emails> in <account/region>`


gcomstock:
Love it!

PR #393, opened 2018-11-08
 feat(core): Add functionality for artifact-extracting Jinja templates

ezimanyi:


ezimanyi:
This PR add the following:
* Standard and user-configurable artifact-extracting Jinja templates
* Artifact post-processor now actually processes Jinja templates to create artifacts

The one piece that is missing to wire this all together is that the artifact post-processor still hard-codes an empty list of artifact templates to process, so is a no-op for all triggers.  The last change to wire this all together will be implementing the piece that selects an appropriate template based on the trigger.  (I'll also add some real tests to the artifact post-processor at that point.)

echo-core/src/main/java/com/netflix/spinnaker/echo/artifacts/MessageArtifactTranslator.java
```diff
@@ -56,7 +58,7 @@ public MessageArtifactTranslator(InputStream templateStream, JinjavaFactory jinj
       this.jinjaTemplate = "";
     } else {
       try {
-        this.jinjaTemplate = IOUtils.toString(templateStream);
+        this.jinjaTemplate = IOUtils.toString(templateStream, Charset.defaultCharset());
```
dreynaud: I don't feel super comfortable with `defaultCharset()`, can we explicitly declare `Charsets.UTF_8` instead?

echo-pipelinetriggers/src/main/java/com/netflix/spinnaker/echo/config/ArtifactTemplateProperties.java
```diff
@@ -0,0 +1,62 @@
...
+import java.util.List;
+import java.util.Map;
+import java.util.stream.Collectors;
+
+@Data
+@Configuration
+@ConfigurationProperties(prefix = "artifacts")
+@Validated
+public class ArtifactTemplateProperties {
+  @Valid
```
dreynaud: I'm not familiar with this annotation, what is its purpose?

echo-pipelinetriggers/src/main/java/com/netflix/spinnaker/echo/pipelinetriggers/artifacts/JinjaTemplateService.java
```diff
@@ -0,0 +1,56 @@
...
+import com.netflix.spinnaker.echo.config.ArtifactTemplateProperties;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Component;
+
+import javax.annotation.PostConstruct;
+import java.util.Map;
+
+@Component
+public class JinjaTemplateService {
+  private ArtifactTemplateProperties artifactTemplateProperties;
```
dreynaud: final?

echo-pipelinetriggers/src/main/java/com/netflix/spinnaker/echo/pipelinetriggers/postprocessors/ArtifactPostProcessor.java
```diff
@@ -16,19 +16,74 @@
...
+
 /**
  * Post-processor extracts artifacts from a pipeline using a supplied Jinja template and adds
  * these artifacts to the pipeline as received artifacts.
  * This post-processor is not implemented yet and is currently a no-op.
  */
 @Component
+@Slf4j
 public class ArtifactPostProcessor implements PipelinePostProcessor {
+  private ObjectMapper objectMapper;
```
dreynaud: final

echo-core/src/main/java/com/netflix/spinnaker/echo/artifacts/MessageArtifactTranslator.java
ezimanyi: I actually wanted to make it `Charsets.UTF_8` but the deprecated one-argument method itself calls `Charset.default()` for the missing argument so this seemed more backwards compatible.

That being said, it's probably safer and better to just enforce UTF-8 here so I can make the change.

echo-pipelinetriggers/src/main/java/com/netflix/spinnaker/echo/config/ArtifactTemplateProperties.java
ezimanyi: Per the [Spring docs](https://docs.spring.io/spring-boot/docs/current/reference/html/boot-features-external-config.html#boot-features-external-config-validation) "Although nested properties will also be validated when bound, it’s good practice to also annotate the associated field as @Valid. This ensure that validation is triggered even if no nested properties are found."

So my understanding is this ensures we validate the field in all cases...though I'll be honest I'm not sure I 100% understand the subtletly there and am just following the best practice.  (If we don't bind anything to the property, I'm not sure why we'd need to validate...)

ezimanyi:
Thanks for the comments @dreynaud , addressed all of them.

PR #397, opened 2018-11-12
fix(triggers): Fix triggered pipeline template artifact resolution.

jtk54:


jtk54:
Tests are passing locally, tag was added 3 hours ago. Not sure why this is `walk fail`ing.

PR #401, opened 2018-11-26
feat(artifacts): Add classifier to jar artifact

ezimanyi:
Adds the ability to set an optional classifier on a jar artifact. Also changes the artifact type from jar/archive to repotype/file where repotype defaults to maven.


lwander:
Is repotype needed because fetching the artifact is too ambiguous w/o knowing where it's hosted?

ezimanyi:
Yes, there are different artifact accounts for maven and ivy, as they have different repository layouts and the fetching code needs to know which repository it's fetching from.  (This currently can only be maven or ivy.)

PR #402, opened 2018-11-26
fix(core): Don't retry pipeline initiation when receiving a 400

ezimanyi:
When a request to Orca to start a pipeline fails, Echo currently retries a number of times (defaulting to 5). In cases where the failure is due to a problem with the request, we should not retry; gate the retry logic so that we don't retry when receiving a 400 - Bad Request from Orca. In particular, this will stop having pipelines missing artifacts lead to 5 failed executions in the history.

See also https://github.com/spinnaker/orca/pull/2524

### DIFF
--- a/echo-pipelinetriggers/src/main/java/com/netflix/spinnaker/echo/pipelinetriggers/orca/PipelineInitiator.java
+++ b/echo-pipelinetriggers/src/main/java/com/netflix/spinnaker/echo/pipelinetriggers/orca/PipelineInitiator.java
@@ -38,6 +38,7 @@ import org.springframework.http.HttpStatus;
 import org.springframework.stereotype.Component;
 import retrofit.RetrofitError;
 import retrofit.RetrofitError.Kind;
+import retrofit.client.Response;
 import rx.Observable;
 import rx.functions.Func1;
 
@@ -154,7 +155,7 @@ public class PipelineInitiator {
   }
 
   private static boolean isRetryable(Throwable error) {
-    if (!(error instanceof  RetrofitError)) {
+    if (!(error instanceof RetrofitError)) {
       return false;
     }
     RetrofitError retrofitError = (RetrofitError) error;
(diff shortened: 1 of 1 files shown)


echo-pipelinetriggers/src/main/java/com/netflix/spinnaker/echo/pipelinetriggers/orca/PipelineInitiator.java
```diff
@@ -153,8 +154,20 @@ private void onOrcaError(Pipeline pipeline, Throwable error) {
   }
 
   private static boolean isRetryable(Throwable error) {
-    return error instanceof RetrofitError &&
-      (((RetrofitError) error).getKind() == Kind.NETWORK || ((RetrofitError) error).getKind() == Kind.HTTP);
+    if (!(error instanceof  RetrofitError)) {
```
lwander: Extra space after `instanceof`

echo-pipelinetriggers/src/main/java/com/netflix/spinnaker/echo/pipelinetriggers/orca/PipelineInitiator.java
```diff
@@ -153,8 +154,20 @@ private void onOrcaError(Pipeline pipeline, Throwable error) {
...
+      return false;
+    }
+    RetrofitError retrofitError = (RetrofitError) error;
+
+    if (retrofitError.getKind() == Kind.NETWORK) {
+      return true;
+    }
+
+    if (retrofitError.getKind() == Kind.HTTP && retrofitError.getResponse().getStatus() != HttpStatus.BAD_REQUEST.value()) {
```
lwander: Are we sure that `.getResponse` is not `null`?

ezimanyi: Good point; it shouldn't be as `Kind.HTTP` explicitly means that it received a non-200 response, but that's not really statically enforced so I fixed it to be null-safe.

ezimanyi:
I tested that my Orca change did cause it to correctly return a 400; I'm not getting a single failure in my execution history for missing artifacts instead of the 5 failures I was getting before.

PR #403, opened 2018-11-27
Revert "fix(triggers): Fix triggered pipeline template artifact resolution."

jtk54:
Reverts spinnaker/echo#398

@jervi FYI

Reverting this in favor of https://github.com/spinnaker/orca/pull/2510 which is a more simple approach for the time being.


lwander:
should we cherry pick into 1.11?

jtk54:

PR #407, opened 2018-12-03
fix(notification/githubStatus): add retries to update github check

xavileon:
So updating the check is more reliable in the event of failures to update the check.

Signed-off-by: Xavi León <xavi.leon@schibsted.com>


xavileon:
@spinnaker-reviewers PTAL

echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/notification/GithubNotificationAgent.java
```diff
@@ -101,7 +112,12 @@ public void sendNotifications(
 
     GithubStatus githubStatus = new GithubStatus(state, targetUrl, description, context);
     try {
-      githubService.updateCheck("token " + token, content.getRepo(), branchCommit, githubStatus);
+      final String repo = content.getRepo();
+      retrySupport.retry(
+        () -> githubService.updateCheck("token " + token, repo, branchCommit, githubStatus),
+        MAX_RETRY,
+        RETRY_BACKOFF,
+        true);
```
emjburns: We tend to not do exponential backoff in other areas of the code.

emjburns:
LGTM once comments addressed!

echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/notification/GithubNotificationAgent.java
robzienert: Agree, I would recommend against using the exponential API.

xavileon:
@emjburns Thanks! Comments addressed in https://github.com/spinnaker/echo/pull/407/commits/4c77baa940fe0e0e021ab5ca0f5acf9c17fa2919

PR #414, opened 2018-12-17
fix(artifacts): Fix artifact population in manual triggers

ezimanyi:
Some workflows (ex: manual triggers specifying a docker tag) directly inject artifacts into the trigger; we need to add an artifacts field to the Trigger model so that these work properly.

### DIFF
--- a/echo-model/src/main/java/com/netflix/spinnaker/echo/model/Trigger.java
+++ b/echo-model/src/main/java/com/netflix/spinnaker/echo/model/Trigger.java
@@ -128,7 +128,7 @@ public class Trigger {
   boolean dryRun = false;
 
   List<Map<String, Object>> notifications;
-  List<Map<String, String>> artifacts;
+  List<Map<String, Object>> artifacts;
 
   /**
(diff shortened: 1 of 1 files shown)


ezimanyi:
Addresses https://github.com/spinnaker/spinnaker/issues/3737

echo-model/src/main/java/com/netflix/spinnaker/echo/model/Trigger.java
```diff
@@ -128,6 +128,7 @@ public String toString() {
   boolean dryRun = false;
 
   List<Map<String, Object>> notifications;
+  List<Map<String, String>> artifacts;
```
lwander: I think you want `<String, Object>` because of the `metadata` field.

ezimanyi: Good point, I'll fix that

ezimanyi:

ezimanyi:
@spinnakerbot cherry-pick 1.10

PR #314, opened 2018-08-15
feat(pubsub): Adds Pubsub Publisher as an EchoEventListener. The firs…

ttomsu:
…t impl pushes all events to a configured Google Pubsub topic.

This results in a subscriber on the topic getting a series of messages at various levels of task, stage, and pipeline/orchestration execution. Here's a sample message:

```
ackId: QBJMOABESVMrQwsqWBFOBCEhPjA-RVNEUAYWLF1GSFE3GQhoUQ5PXiM_NSAoRRIHA08CKF15MkUxQV8aB1ENGXJ8Z3BpDxMEAxAFK1VbEQ16bVxXOFAKGXd6ZHRiWBQFBE1ReneEx7LYmpJjZho9XhJLLD5-MzVFQV4
message:
  attributes:
    application: sekret
    name: resizeServerGroup
    rawType: orca:task:starting
    source: orca
    status: starting
    taskName: resizeServerGroup.waitForCapacityMatch
    type: task
[...]

PR #417, opened 2018-12-17
feat(pubsub): Enables (Google) Pubsub as first class notification mechanism

ttomsu:
This PR extends https://github.com/spinnaker/echo/pull/314.

Previous functionality has been moved to `GooglePusubAllContentEventListener` because that implementation listens on the raw event stream. To continue using this implementation, users must add `content: ALL` to the config:

```
pubsub:
  google:
    publishers:
    - project: ttomsu-dev-spinnaker
      topicName: spinnaker-notifications
      content: ALL
      jsonPath: /path/to/credential.json
```

The new default is the `GooglePubsubNotificationEventListener` (`content: NOTIFICATIONS`). This impl hooks into the `AbstractEventNotificationAgent`, which reads the raw event stream and does some processing before handing them down to the concrete notification subclass.

Manual judgement states are a special case: they invoke `/notifications` directly (as opposed to the raw event stream, which just hits `/`). The `GooglePubsubNotificationEventListener` also handles these notifications by implementing `NotificationService`. Why there are two ways to do notifications? I don't know.

Update Pubsub library to the latest published in maven central.

Things for reviewers to consider:
1. Attribute list on pubsub messages - these can be as little as:

```
attributes:
    rawType: build
```
for Igor notifications to

```
message:
  attributes:
    application: sekret
    name: Wait
    rawType: orca:task:complete
    source: orca
    status: complete
    taskName: wait.wait
    type: task
```

for low-level tasks.

2. I removed pubsub account names, but I'm not sure if I like it. My testing always referred to the short topic name, i.e. `spinnaker-notifications` from the full topic name `projects/ttomsu-dev-spinnaker/topics/spinnaker-notifications`. From a UI/UX perspective, should we always use the full name? A short name? What if two short names conflict, should we continue to use a completely separate display/operator given name?


ttomsu:
No - it just defaults to NOTIFICATIONS, which won't result in any notifications until there is UI support.

lwander:
My concern is that users won't know/forget to set the flag. Can the default be "ALL" instead of "NOTIFICATIONS" for this reason?

ttomsu:
Switched default to ALL

echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/notification/AbstractEventNotificationAgent.groovy
```diff
@@ -16,17 +16,23 @@
...
 import org.springframework.beans.factory.annotation.Value
 
+@Slf4j
 abstract class AbstractEventNotificationAgent implements EchoEventListener {
 
   @Autowired
   Front50Service front50Service
 
+  @Autowired
+  ObjectMapper mapper = new ObjectMapper()
```
lwander: Just wondering, why assign a default as well as autowire?

echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/notification/AbstractEventNotificationAgent.groovy
```diff
@@ -47,68 +53,71 @@ abstract class AbstractEventNotificationAgent implements EchoEventListener {
 
   @Override
   void processEvent(Event event) {
+    if (log.isDebugEnabled()) {
```
lwander: Shouldn't the logger only log if debug is enabled anyway?

echo-pubsub-google/src/main/java/com/netflix/spinnaker/echo/pubsub/google/GooglePubsubNotificationEventListener.java
```diff
@@ -0,0 +1,115 @@
...
+   * Judgement stage from Orca.
+   */
+  @Override
+  public EchoResponse.Void handle(Notification notification) {
+    if (log.isDebugEnabled()) {
+      try {
+        log.debug("Notification received: " + mapper.writerWithDefaultPrettyPrinter()
+          .writeValueAsString(notification));
+      } catch (JsonProcessingException jpe) {
+        log.warn("Error parsing notification", jpe);
```
lwander: Will this cause trouble later in this method? I'm wondering if this is validation we need to be doing in all cases & report on it.

echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/notification/AbstractEventNotificationAgent.groovy
ttomsu: Because tests failed without it being autowired. The preference is to use the autowired one if available, but otherwise just use a new one.

echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/notification/AbstractEventNotificationAgent.groovy
ttomsu: Yes, but the overhead of doing the serialization is avoided with this check because the args are evaluated if you just do `log.debug(arg1, arg2, ... )`

echo-pubsub-google/src/main/java/com/netflix/spinnaker/echo/pubsub/google/GooglePubsubNotificationEventListener.java
ttomsu: I don't ever expect this to be called, but because it's a checked exception, I have to try/catch it.

echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/notification/AbstractEventNotificationAgent.groovy
ezimanyi: I've tried to move towards constructor injection for this exact reason; makes it easier to have tests control what is being injected

echo-pubsub-google/src/main/java/com/netflix/spinnaker/echo/pubsub/google/GooglePubsubAllContentEventListener.java
```diff
@@ -0,0 +1,49 @@
...
+@Component
+@Slf4j
+@ConditionalOnExpression("${pubsub.enabled:false} && ${pubsub.google.enabled:false}")
+public class GooglePubsubAllContentEventListener implements EchoEventListener {
+
+  @Autowired
+  private PubsubPublishers publishers;
+
+  @Autowired
+  ObjectMapper mapper;
```
ezimanyi: Looks unused

echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/notification/AbstractEventNotificationAgent.groovy
ttomsu: Removed the new object creation, made it `required = false`, and checked that the object isn't null before using it.

PR #418, opened 2018-12-17
 fix(core): Fix auth propagation for manual triggers

ezimanyi:


echo-pipelinetriggers/src/main/java/com/netflix/spinnaker/echo/pipelinetriggers/orca/PipelineInitiator.java
```diff
@@ -93,10 +93,16 @@ public void startPipeline(Pipeline pipeline) {
...
         if (templatedPipelineType.equals(pipeline.getType())) { // TODO(jacobkiefer): Constantize.
+          // We need to store and re-set the propagateAuth flag, as it is ignored on deserialization
+          // TODO(ezimanyi): Find a better way to pass the propagateAuth flag than on the trigger itself
+          boolean propagateAuth = pipeline.getTrigger() != null && pipeline.getTrigger().isPropagateAuth();
           log.debug("Planning templated pipeline {} before triggering", pipeline.getId());
           pipeline = pipeline.withPlan(true);
           Map resolvedPipelineMap = orca.plan(objectMapper.convertValue(pipeline, Map.class));
           pipeline = objectMapper.convertValue(resolvedPipelineMap, Pipeline.class);
+          if (propagateAuth) {
+            pipeline = pipeline.withTrigger(pipeline.getTrigger().atPropagateAuth(true));
```
jtk54: Could just inline `propagateAuth` as the value to `atPropagateAuth`.

ezimanyi: I tried but ran into a failed test where `pipeline.getTrigger()` was null, so I'd still need to somehow guard that (as the `propagateAuth` is also guarding against a null trigger).  I think this is safer for now, but I can clean this up as part of my TODO.

ezimanyi:

PR #419, opened 2018-12-17
 fix(MPT): Fix manual triggering of templated pipelines

ezimanyi:
This is a manual cherry-pick of https://github.com/spinnaker/echo/pull/418 into 1.10.  (The first commit applied cleanly, but the second I had to manually do.)

This bug only shows up behind a flag in 1.10, but I'd like to fix the behavior so people can test the flag.


jtk54:
(If you make any).

PR #421, opened 2018-12-19
fix(jenkins): Fix encoding of Jenkins URL

ezimanyi:
Jenkins requires build names containing slashes to be sent with the slashes unencoded.


ezimanyi:
Addresses https://github.com/spinnaker/spinnaker/issues/3766

ezimanyi:

PR #424, opened 2018-12-21
feat(artifacts/bitbucket): Accept Bitbucket Server hooks

abergmeier:
Since Bitbuckets Server hooks are different from Bitbucket Cloud, now also
add support for the former, so can be triggered with on-prem.

### DIFF
--- a/echo-webhooks/src/main/groovy/com/netflix/spinnaker/echo/controllers/WebhooksController.groovy
+++ b/echo-webhooks/src/main/groovy/com/netflix/spinnaker/echo/controllers/WebhooksController.groovy
@@ -54,6 +54,10 @@ class WebhooksController {
     event.details.requestHeaders = headers
     event.rawContent = rawPayload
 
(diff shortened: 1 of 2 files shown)


abergmeier:
Not tested yet.

abergmeier:
Any guidance, how we could make this automatically testable would be appreciated.

abergmeier:
Found 3 things still left to do:

- [ ] Support Bitbucket Server testing Request
  This is tricky, since Bitbucket Server seems to send a request without any content. It fails already in _Gate_.
- [ ] Fix bug for Merges
- [ ] Introduce tests

abergmeier:
>  regarding your question on testing in #424 (comment), might be able to do something similar to the spec for handles initial github ping

Thanks for the pointers. Helped a lot. Since some problems come from interacting with _Gate_, is there a similar test in _Gate_?

ethanfrogers:
spinnaker/spinnaker#3073.

We need to clean this area up a lot :/.

PR #426, opened 2019-01-03
fix(artifacts): Allow customFormat to be a boolean

ezimanyi:
When parsing the customFormat supplied in the property file, we currently assume it is a string. If the property file is supplied as YAML, it could be a boolean.

### DIFF
--- a/echo-pipelinetriggers/src/main/java/com/netflix/spinnaker/echo/pipelinetriggers/postprocessors/ArtifactPostProcessor.java
+++ b/echo-pipelinetriggers/src/main/java/com/netflix/spinnaker/echo/pipelinetriggers/postprocessors/ArtifactPostProcessor.java
@@ -134,7 +134,7 @@ public class ArtifactPostProcessor implements PipelinePostProcessor {
       return Boolean.parseBoolean((String) customFormat);
     }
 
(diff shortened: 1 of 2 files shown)


ezimanyi:
Addresses comment 2 in https://github.com/spinnaker/spinnaker/issues/3401#issuecomment-450695043

echo-pipelinetriggers/src/main/java/com/netflix/spinnaker/echo/pipelinetriggers/postprocessors/ArtifactPostProcessor.java
```diff
@@ -112,14 +112,31 @@ private JinjaTemplate getTemplateFromProperty(Trigger trigger) {
...
+
+    if (customFormat instanceof Boolean) {
+      return (Boolean) customFormat;
+    }
+
+    if (customFormat instanceof String) {
+      return Boolean.parseBoolean((String) customFormat);
+    }
+
+    throw new RuntimeException("Unexpected customFormat in property file.");
```
lwander: I would append the `customFormat` object here so the exception reader gets an idea of what they accidentally passed.

PR #427, opened 2019-01-04
fix(pipelines): Pass resolveArtifacts = true to plan endpoint

ezimanyi:
Orca now requires callers to explicitly request artifact resolution when calling '/plan' on a templated pipeline. We do want this behavior when triggering from echo, so pass the flag.


ezimanyi:
Fixes the bug noted in https://github.com/spinnaker/spinnaker/issues/2654#issuecomment-449546544

Depends on https://github.com/spinnaker/orca/pull/2576

ezimanyi:

PR #431, opened 2019-01-08
feat(pubsub): update GCB transform template to extract more artifacts

sbwsg:
We're able to extract more artifacts than we currently do from Cloud Build pubsub messages.

### DIFF
--- a/echo-pubsub-google/src/main/resources/gcb.jinja
+++ b/echo-pubsub-google/src/main/resources/gcb.jinja
@@ -6,5 +6,14 @@
     "name": "{{ image.name }}",
     "version": "{{ image.digest }}",
     "type": "docker/image"
-  } {% if not loop.last %} , {% endif %} {% endfor %}
+  } {% if (not loop.last) or ((artifacts.objects.paths | length) > 0) %} , {% endif %} {% endfor %}
+
+  {% set location = artifacts.objects.location %}
+  {% for path in artifacts.objects.paths %}
+    {
+      "reference": "{{ location }}{{ path }}",
+      "name": "{{ location }}{{ path }}",
+      "version": "{{ location }}{{ path }}",
+      "type": "gcs/object"
+    } {% if not loop.last %} , {% endif %}{% endfor %}
 ]


echo-pubsub-google/src/main/resources/gcb.jinja
```diff
@@ -6,5 +6,14 @@
...
-  } {% if not loop.last %} , {% endif %} {% endfor %}
+  } {% if (not loop.last) or ((artifacts.objects.paths | length) > 1) %} , {% endif %} {% endfor %}
+
+  {% set location = artifacts.objects.location %}
+  {% for path in artifacts.objects.paths %}
+    {
+      "reference": "{{ location }}{{ path }}",
+      "name": "{{ location }}{{ path }}",
+      "version": "{{ location }}{{ path }}",
+      "type": "gcs/object"
```
sbwsg: just realised cloud build might be able to upload to non-GCS locations. Need to double check and update this accordingly.

lwander:
Just curious, is anyone asking for this?

echo-pubsub-google/src/main/resources/gcb.jinja
```diff
@@ -6,5 +6,14 @@
     "name": "{{ image.name }}",
     "version": "{{ image.digest }}",
     "type": "docker/image"
-  } {% if not loop.last %} , {% endif %} {% endfor %}
+  } {% if (not loop.last) or ((artifacts.objects.paths | length) > 1) %} , {% endif %} {% endfor %}
```
lwander: Why `> 1` instead of `>= 1`? I would think that if there are any items we need a comma

echo-pubsub-google/src/main/resources/gcb.jinja
sbwsg: Confirmed that objects in this struct are GCS-only: https://cloud.google.com/cloud-build/docs/configuring-builds/store-images-artifacts#storing_artifacts_in_gcs_name_short

echo-pubsub-google/src/main/resources/gcb.jinja
sbwsg: Oops; fixed.

PR #436, opened 2019-01-15
fix(webhooks/bitbucket): Correct repoProject for Bitbucket Server

abergmeier:


ezimanyi:
@abergmeier : I just merged your other PR, #435; looks like this one now has a merge conflict with it.

abergmeier:
Should be fine now.

PR #437, opened 2019-01-15
fix(pubsub): Reintroduces 'name' field to google pubsub publishers

ttomsu:
part of https://github.com/spinnaker/spinnaker/issues/2784

We used to have a operator-defined `name` field, but I took it out in the last revamp. Adding it back in to support the (probably?) niche use case of the same topic name in different projects.

This means that in order to trigger notifications, we match on `publisherName` instead of `topic` in the incoming notification map.

### DIFF
--- a/echo-pubsub-core/src/main/java/com/netflix/spinnaker/echo/pubsub/PubsubSubscriptionController.java
+++ b/echo-pubsub-core/src/main/java/com/netflix/spinnaker/echo/pubsub/PubsubSubscriptionController.java
@@ -16,10 +16,8 @@
 
 package com.netflix.spinnaker.echo.pubsub;
 
-import lombok.AllArgsConstructor;
 import lombok.Builder;
 import lombok.Data;
-import lombok.NoArgsConstructor;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.web.bind.annotation.RequestMapping;
 import org.springframework.web.bind.annotation.RequestMethod;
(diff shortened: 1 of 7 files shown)


echo-pubsub-core/src/main/java/com/netflix/spinnaker/echo/pubsub/PubsubSubscriptionController.java
```diff
@@ -16,6 +16,8 @@
 
 package com.netflix.spinnaker.echo.pubsub;
 
+import java.util.List;
```
ezimanyi: I think normally the `java.util.*` goes at the end in our repos

echo-pubsub-core/src/main/java/com/netflix/spinnaker/echo/pubsub/PubsubSubscriptionController.java
```diff
@@ -57,4 +60,27 @@
...
+        .pubsubSystem(p.getPubsubSystem().toString())
+        .publisherName(p.getName())
+        .topicName(p.getTopicName())
+        .build())
+      .collect(Collectors.toList());
+  }
+
+  @Data
+  @Builder
+  @NoArgsConstructor
```
ezimanyi: Minor point; I think that `@Builder` implies `@AllArgsConstructor` (which is what it uses) so we could remove that and also remove the `@NoArgsConstructor`, though I think in practice we tend to add all these annotations to data classes.

echo-pubsub-core/src/main/java/com/netflix/spinnaker/echo/pubsub/PubsubSubscriptionController.java
ttomsu: Boo intellij formatter...

echo-pubsub-core/src/main/java/com/netflix/spinnaker/echo/pubsub/PubsubSubscriptionController.java
ttomsu: Removed both.

ttomsu:
1. Nobody should be using this yet - if there were, they probably would have complained when I removed it back in December.

2. Additional changes to Orca and Deck are linked in https://github.com/spinnaker/spinnaker/issues/2784

PR #463, opened 2019-02-14
fix(pipelines): Add keepWaitingPipelines to Pipeline model

ezimanyi:
The Pipeline model is missing keepWaitingPipelines, which means it's ignored on deserialization and setting it on a pipeline has no effect.

### DIFF
--- a/echo-model/src/main/java/com/netflix/spinnaker/echo/model/Pipeline.java
+++ b/echo-model/src/main/java/com/netflix/spinnaker/echo/model/Pipeline.java
@@ -58,6 +58,9 @@ public class Pipeline {
   boolean limitConcurrent;
 
   @JsonProperty
+  boolean keepWaitingPipelines;
+
+  @JsonProperty
   boolean plan;
 
   @JsonProperty


ezimanyi:
Closes spinnaker/spinnaker#3971

ezimanyi:
@spinnakerbot cherry-pick 1.12

ajordens:
How strongly typed do we actually need the pipeline model in `echo` to be? Not the first time we've run into this and it seems like `JsonAnyGetter/Setter` might be an alternative to just passing through everything as-is (at least anything that we don't explicitly need to access in `echo`).

Just a thought.

ezimanyi:
Yeah, we could add `JsonAnyGetter/Setter`, though I've been frustrated enough times by the difference in pipeline model between `echo` and `orca` (and the presence of undocumented fields) so I'd be hesitant to loosen things up even more.  I've actually thought at some point we should try to push a common shared model between orca/echo up to `kork`.  Though I guess `JsonAnyGetter/Setter` doesn't really preclude that work as long as we migrate the superset of the two schemas to `kork` in this hypothetical plan.

So maybe if we keep seeing this it's a reasonable compromise to loosen up the model for the sake of reducing bugs until we make a shared model...

ajordens:
This is unfortunately not the first time we've been had issues with drifting models between orca and echo (and even the precursor to pipelines being in echo --- used to be in a service called `post`). 

I've come around to the opinion that if a service is largely passthrough than the addition of a strict model is likely going to be more trouble than it's worth.

Ok leaving this as-is provided the next time it comes up we give some thought to finding a "better way" :)

PR #467, opened 2019-02-21
feat(buildmasters): Permission support for build masters (CI's)

jervi:
Depends on https://github.com/spinnaker/igor/pull/374, please see that PR for a description.


ezimanyi:
As noted on spinnaker/igor#374, I think that in general Echo won't have any authorization context to pass along to Igor.  The only exception is in the event of a manual trigger, where the user's context is passed along from Gate -> Echo (and thus this would pass it along).

In most cases, Echo is responding to unauthenticated event notifications, so won't have any context to pass along.

jervi:
@ezimanyi I've addressed the auth issue now 👍

ezimanyi:
Looks good!  Makes me wonder if we should somewhat generalize the concept of having an event handler assume the role of a trigger while constructing the trigger, but we can revisit that if this pattern is needed in other event handlers.

PR #469, opened 2019-02-26
feat(artifacts): emit artifacts received in echo

emjburns:
This PR adds a an `ArtifactEmitter` that accepts artifacts from within echo and publishes them to all configured rest clients. 

Right now at Netflix we get pubsub messages for each new debian that is produced. We want to forward these on to keel so that we can react to new artifacts and bake and deploy them. @ezimanyi does `MessageArtifactTranslator` seem like a reasonable place to emit these artifacts from? 

This is using the Spring `EventListener` (unlike how other events are passed around in echo) because I don't think we need the authenticated context.


echo-core/src/main/java/com/netflix/spinnaker/echo/artifacts/MessageArtifactTranslator.java
```diff
@@ -72,7 +82,11 @@ public MessageArtifactTranslator(InputStream templateStream, JinjavaFactory jinj
     }
 
     ObjectMapper mapper = new ObjectMapper();
-    return readArtifactList(mapper, jinjaTransform(mapper, messagePayload));
+    List<Artifact> artifacts = readArtifactList(mapper, jinjaTransform(mapper, messagePayload));
+    if (!artifacts.isEmpty()) {
+      applicationEventPublisher.publishEvent(new ArtifactEvent(null, artifacts));
```
emjburns: This is where the artifacts get sent to the `ArtifactEmitter` - via spring application events.

emjburns:
Most changes in this PR are around injecting the `ApplicationEventPublisher` into the pubsub classes that create subscribers. The new functionality is in `echo-artifacts`.

echo-artifacts/src/main/java/com/netflix/spinnaker/echo/config/ArtifactEmitterConfig.java
```diff
@@ -0,0 +1,68 @@
...
+    Ok3Client ok3Client,
+    RestAdapter.LogLevel retrofitLogLevel
+  ) {
+    ArtifactEmitterUrls artifactEmitterUrls = new ArtifactEmitterUrls();
+
+    for (ArtifactEndpointConfiguration endpoint : artifactEmitterProperties.endpoints) {
+
+      RestAdapter.Builder restAdapterBuilder = new RestAdapter.Builder()
+        .setEndpoint(newFixedEndpoint(endpoint.getUrl()))
+        .setClient(endpoint.internal ? ok3Client : clientFactory.getClient(endpoint.insecure))
```
ezimanyi: Why use a different client for internal and external requests?

echo-artifacts/src/main/java/com/netflix/spinnaker/echo/config/ArtifactEmitterProperties.java
```diff
@@ -0,0 +1,38 @@
...
+
+import org.springframework.boot.context.properties.ConfigurationProperties;
+
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * Helper class to map hosts in properties file into a list
+ */
+@ConfigurationProperties(prefix = "artifact-emitter")
```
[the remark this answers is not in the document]
emjburns: The boot2 migration means that everything must switch to kebab case - so switching now event though it is awkward. (https://docs.google.com/document/d/1Qs9q-l-uRfrwoDXRfkcL_qKlhA-4bbYG4Id-MxLrFHA/edit?ts=5c704129)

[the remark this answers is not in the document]
ezimanyi: Ah, didn't realize that! Definitely makes sense to use the new casing then.

echo-artifacts/src/main/java/com/netflix/spinnaker/echo/config/ArtifactEmitterConfig.java
emjburns: Internal clients are configured with some security-ness (can you tell I'm not an expert in this? ha ha). With the generic client factory we don't have the credentials to talk to our services. The thought is "internal=true" means we're sending this artifact to something like `keel` or `front50`. External would be like a separate auditing system or some service that's not in spinnaker.

Does that make sense? @ajordens maybe you could weigh in on if this is the right strategy or if I can use just one of the clients for everything.

emjburns:
After discussing with @ezimanyi offline, we agree that there doesn't seem to be a security concern but there aren't a lot of general use cases for this behavior. Instead, I'm going to refactor this PR to just send artifacts to Keel. Any new services that the artifact should be sent to will have to be added in code OR this general config will be restored.

echo-artifacts/src/main/java/com/netflix/spinnaker/echo/config/ArtifactEmitterConfig.java
emjburns: never mind @ajordens, I changed this to create a keel service and send events to that.

PR #472, opened 2019-02-28
test(core): Basic startup test

ezimanyi:
Test that echo starts up with a very basic config, with just the baseUrl for each dependent service defined, and with beans mocking these services.


ezimanyi:
Technically we could make the URLs not required by adding a default in a few `@Value` annotations, but I'm not sure that's helpful as I don't think `echo` can reasonably function without `orca` or `front50`

PR #473, opened 2019-03-04
fix(artifacts): Move build fetching to build event handler

ezimanyi:
Artifact contstraints on build triggers currently don't work as the constraint is evaluated while determining which trigger to use but build information and artifact inflation only happens once the pipeline is ready to be sent to Orca. Move the fetching of build information to earlier in the triggering process (and scope it to only build triggers). A follow-up change will move artifact inflation earlier as well.


ezimanyi:
I added `BuildInfoPostProcessor` in #380, but in retrospect I think that having this done as a pipeline post-processor was too general, and that the logic should move earlier and be scoped to actual build events.

The one part of this change that I'm not 100% happy with is the need to re-construct a `BuildEvent` from the information on a manual trigger.  Thinking about this made me realize that it might make sense for manual triggers to send along:
* a pipeline id
* an event we want to trigger with (wrapped in a `ManualEvent` so we can distinguish)
Then the processing of these would really be the same as actual events---vs. now where it's sort of strange we're getting a fully-formed trigger.  Thinking about that end state is what made me opt to reconstruct a build event (rather than some other solution like pass the trigger directly to `BuildInfoService`)

There will be similar follow-up PR that moves artifact extraction earlier as well to fix the bug.

echo-pipelinetriggers/src/main/java/com/netflix/spinnaker/echo/pipelinetriggers/eventhandlers/ManualEventHandler.java
```diff
@@ -93,4 +102,23 @@ private Pipeline buildTrigger(Pipeline pipeline, Trigger manualTrigger) {
...
     return notifications;
   }
+
+  // Manual triggers try to replicate actual events (and in some cases build events) but rather than pass the event to
+  // echo, they add the information to the trigger. It may make sense to refactor manual triggering to pass a trigger and
+  // an event as with other triggers, but for now we'll see whether we can extract a build event from the trigger.
+  private Optional<BuildEvent> extractBuildInformation(Trigger manualTrigger) {
+    String master = manualTrigger.getMaster();
+    String job = manualTrigger.getJob();
+    if (StringUtils.isNoneEmpty(master, job)) {
```
jtk54: What is `isNoneEmpty`?

ezimanyi: It's equivalent to `StringUtils.isNotEmpty(master) && StringUtils.isNotEmpty(job)` but shorter...though maybe it's clearer just to just `isNotEmpty`?

jtk54: I think this is reasonable, just wondering.

PR #478, opened 2019-03-05
fix(artifacts): Extract artifacts before enforcing constraints

ezimanyi:
refactor(triggers): Push artifact extraction/matching to base

All of the event handlers duplicate the same logic around extracting some artifacts from an event, and checking any artifact constraints on the trigger. Push this logic to BaseTriggerEventHandler.

This commit also removes any need for the specific EventHandler classes to know about Pipelines at all; they deal only in Triggers and Events.

fix(artifacts): Extract artifacts before enforcing constraints

We currently extract artifacts using Jinja templates as a final post-processing step on the pipeline. This means that the artifacts are not available when we check artifact contstraints. Fix this bug by moving artifact extraction to occur during trigger evaulation.


echo-pipelinetriggers/src/main/java/com/netflix/spinnaker/echo/pipelinetriggers/artifacts/BuildArtifactExtractor.java
```diff
@@ -43,13 +40,13 @@
  */
 @Component
 @Slf4j
-public class ArtifactPostProcessor implements PipelinePostProcessor {
+public class BuildArtifactExtractor {
```
emjburns: So this is specifically for parsing artifacts out of property files? The naming of "BuildArtifactExtractor" confuses me a little. An updated description of the class might be nice.

emjburns:
I like this change! I like that the individual handlers don't need to know about the pipelines.

echo-pipelinetriggers/src/main/java/com/netflix/spinnaker/echo/pipelinetriggers/artifacts/BuildArtifactExtractor.java
ezimanyi: Great point, I renamed it `JinjaArtifactExtractor` and updated the description.  Right now it only knows how to look in properties for the name of the template (as that's the only place there can be) but I can definitely imagine in the future splitting this out into "where to look for the template name" and "extract the artifacts" if more places become supported.

I also realized the description still said it was an unimplemented no-op...which I should probably have updated months ago

PR #482, opened 2019-03-07
Echo to Boot2

emjburns:
Echo starts up! Wahoo


echo-web/src/test/java/com/netflix/spinnaker/echo/ApplicationSpec.java
```diff
@@ -19,13 +19,15 @@
 import com.netflix.spinnaker.echo.pipelinetriggers.orca.OrcaService;
 import com.netflix.spinnaker.echo.services.Front50Service;
 import org.junit.Test;
```
robfletcher: You need to replace this with the one from the `org.junit.jupiter` package

echo-web/src/test/java/com/netflix/spinnaker/echo/ApplicationSpec.java
```diff
@@ -19,13 +19,15 @@
...
+import org.junit.jupiter.api.extension.ExtendWith;
 import org.junit.runner.RunWith;
```
robfletcher: This shouldn't be imported any more, no?

PR #484, opened 2019-03-08
 refactor(artifacts): Call igor instead of parsing artifacts

ezimanyi:
refactor(artifacts): Remove jinja logic and import from kork

This creates some base classes for extracting artifacts from
messages using Jinja that will be pushed into kork, and updates
echo to use these. It also better encapsulates the dependencies
of the artifact extractors so that we only need to pass around
a factory bean rather than all of the internal dependencies

refactor(artifacts): Call igor instead of parsing artifacts


matthack:
@ezimanyi I think this might have effectively disabled support for custom Jinja artifact templates as documented -- https://www.spinnaker.io/reference/artifacts/from-build-triggers/

We were in the process of setting up artifacts from jenkins triggers using a custom jinja template and can confirm with the latest version of orca, custom artifact templates are no longer working. 

We might prefer to continue to use the old way instead of retrieving jenkins produced artifacts from solely from jenkins. Our jenkins can be ephemeral, so we don't trust the artifacts it stores locally to be available for historical reference and prefer them to come from something like s3.

ezimanyi:
@matthack : Where are you defining the custom jinja templates?  In Spinnaker 1.13 (or on `master` a couple of weeks ago if you're running off of `master`) the logic for processing Jinja templates moved from `echo` to `igor`, so any templates you've defined would need to be moved from your `echo` config to your `igor` config.  (Halyard 1.17+ will do this for you.)  There's a [release note](https://www.spinnaker.io/community/releases/versions/1-13-0-changelog#extracting-artifacts-from-jenkins-jobs) in the 1.13 release with this info.

Definitely let me know if either using Halyard, or moving the definition to `echo` doesn't fix it---the plan is definitely to still support custom templates!

ezimanyi:
Also realized I should update the docs referenced above; created https://github.com/spinnaker/spinnaker.github.io/pull/1282 for that.

PR #492, opened 2019-03-11
fix(trigger): Fixed configuration for manual pipeline triggers

gojenkuser:
Manually triggering a pipeline with a pipeline trigger fails because the pipeline trigger attributes are not deserialized and passed downstream to orca which needs those attributes to retrieve artifacts from upstream pipeline. This fix adds these attributes to the trigger model so they get deserialized and passed downstream.


echo-model/src/main/java/com/netflix/spinnaker/echo/model/Trigger.java
```diff
@@ -65,6 +65,10 @@ public String toString() {
   String type;
   boolean enabled;
 
+  // Configuration for pipeline triggers
+  String parentPipelineId;
+  String paraentPipelineApplication;
```
marchello2000: `parentPipelineApplication`, also I don't see it use anywhere other than a few tests. so probably not needed

gojenkuser: it looked like orca only needed parentPipelineId however i did notice that parentPipelineApplication was passed by deck as part of the trigger so i included it for completeness in case the field is needed in the future.

gojenkuser: ill fix the typo

jonsie:
Let's get that typo fixed and merge this.

jonsie:
LGTM 👍

PR #494, opened 2019-03-11
fix(scheduler): springify quartz jobs

marchello2000:
By making quartz jobs play nicely in the spring land, they can be made into components and participate in DI
This allows us to remove the hack to "request" various beans from within jobs


marchello2000:
@robzienert bumping this up :)

echo-scheduler/src/main/groovy/com/netflix/spinnaker/echo/config/SchedulerConfiguration.groovy
```diff
@@ -67,6 +68,18 @@ class SchedulerConfiguration {
     return factoryBean
   }
 
+  @Bean
+  /**
+   * Job factory used to create jobs as beans on behalf of Quartz
+   */
```
robzienert: Docs should go above method/class annotations.

echo-scheduler/src/main/groovy/com/netflix/spinnaker/echo/config/SchedulerConfiguration.groovy
```diff
@@ -67,6 +68,18 @@ class SchedulerConfiguration {
...
+  JobFactory jobFactory(ApplicationContext applicationContext)
+  {
```
robzienert: We also don't newline opening squigglies.

echo-scheduler/src/main/groovy/com/netflix/spinnaker/echo/config/SchedulerConfiguration.groovy
marchello2000: 👍

echo-scheduler/src/main/groovy/com/netflix/spinnaker/echo/config/SchedulerConfiguration.groovy
marchello2000: dunno what happened there

marchello2000:
@robzienert : fixed up, ready for mergy-merge

PR #498, opened 2019-03-13
fix(pubsub/google): Remove tag from image name

ezimanyi:
The GCB artifact extractor currently includes the tag in the image name; this conflicts with the general convention of excluding it and leads to issues matching incoming artifacts and replacing artifacts in manifests. Fix this to only include the image name.


ezimanyi:
This matches what is done for [GCR artifacts](https://github.com/spinnaker/echo/blob/9c1e03b76165067e56878c35cca1679718ad85ad/echo-pubsub-google/src/main/resources/gcr.jinja#L5).

jtk54:
👍

PR #501, opened 2019-03-15
fix(triggers): Use linear backoff for calls to igor

ezimanyi:


ezimanyi:
Also changed the backoff time to 2 seconds, given that it's linear now; max wait time is now 10 seconds.

PR #504, opened 2019-03-19
fix(MPT): Fixes triggering for template triggers.

jtk54:
Echo has its own cache of pipeline configs, but does not hydrate templated pipeline configs into the full runnable pipeline. Hence if you define a trigger in the template, Echo will not trigger it when it receives a matching trigger event.

### DIFF
--- a/echo-core/src/main/java/com/netflix/spinnaker/echo/services/Front50Service.java
+++ b/echo-core/src/main/java/com/netflix/spinnaker/echo/services/Front50Service.java
@@ -1,6 +1,7 @@
 package com.netflix.spinnaker.echo.services;
 
 import com.netflix.spinnaker.echo.model.Pipeline;
+import java.util.Map;
 import retrofit.http.*;
 
 import java.util.List;
@@ -8,7 +9,7 @@ import java.util.List;
 public interface Front50Service {
   @GET("/pipelines?restricted=false")
   @Headers("Accept: application/json")
-  List<Pipeline> getPipelines();
+  List<Map<String, Object>> getPipelines(); // Return Map here so we don't throw away MPT attributes.
 
   @GET("/pipelines/{application}?refresh=false")
   @Headers("Accept: application/json")
--- a/echo-pipelinetriggers/src/main/java/com/netflix/spinnaker/echo/pipelinetriggers/PipelineCache.java
+++ b/echo-pipelinetriggers/src/main/java/com/netflix/spinnaker/echo/pipelinetriggers/PipelineCache.java
@@ -16,32 +16,36 @@
 
 package com.netflix.spinnaker.echo.pipelinetriggers;
 
+import static java.time.Instant.now;
+
+import com.fasterxml.jackson.databind.ObjectMapper;
 import com.netflix.spectator.api.Registry;
 import com.netflix.spectator.api.patterns.PolledMeter;
 import com.netflix.spinnaker.echo.model.Pipeline;
 import com.netflix.spinnaker.echo.model.Trigger;
+import com.netflix.spinnaker.echo.pipelinetriggers.orca.OrcaService;
 import com.netflix.spinnaker.echo.services.Front50Service;
+import java.time.Duration;
+import java.time.Instant;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.Executors;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
+import java.util.function.Predicate;
+import java.util.stream.Collectors;
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+import javax.annotation.PostConstruct;
+import javax.annotation.PreDestroy;
 import lombok.NonNull;
 import lombok.extern.slf4j.Slf4j;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.beans.factory.annotation.Value;
 import org.springframework.stereotype.Component;
-
-import javax.annotation.Nonnull;
-import javax.annotation.Nullable;
-import javax.annotation.PostConstruct;
-import javax.annotation.PreDestroy;
-import java.time.Duration;
-import java.time.Instant;
-import java.util.ArrayList;
-import java.util.List;
-import java.util.concurrent.Executors;
-import java.util.concurrent.ScheduledExecutorService;
-import java.util.concurrent.TimeUnit;
-import java.util.concurrent.TimeoutException;
-import java.util.stream.Collectors;
-
-import static java.time.Instant.now;
 
 @Component
 @Slf4j
@@ -49,8 +53,10 @@ public class PipelineCache implements MonitoredPoller {
   private final int pollingIntervalMs;
   private final int pollingSleepMs;
   private final Front50Service front50;
+  private final OrcaService orca;
   private final Registry registry;
   private final ScheduledExecutorService executorService;
+  private final ObjectMapper objectMapper;
 
   private transient Boolean running;
   private transient Instant lastPollTimestamp;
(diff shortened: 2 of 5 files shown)


jtk54:
Need to fix tests.

echo-pipelinetriggers/src/main/java/com/netflix/spinnaker/echo/pipelinetriggers/PipelineCache.java
```diff
@@ -132,6 +148,40 @@ void pollPipelineConfigs() {
...
+
+    Map<Boolean, List<Pipeline>> partitionedByV2Schema = templatedPipelines.stream()
+      .filter(p -> p.getSchema() != null)
+      .collect(Collectors.partitioningBy(p -> p.getSchema().equals("v2")));
+
+    List<Pipeline> v2Templated = partitionedByV2Schema.get(Boolean.TRUE);
+    List<Pipeline> v1Templated = partitionedByV2Schema.get(Boolean.FALSE);
+
+    List<Pipeline> plannedV1Templated = v1Templated.stream().filter(Objects::nonNull)
+      .map(p -> orca.plan(objectMapper.convertValue(p, Map.class), false))
```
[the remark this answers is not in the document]
ajordens: Yes, please!

echo-pipelinetriggers/src/main/java/com/netflix/spinnaker/echo/pipelinetriggers/PipelineCache.java
```diff
@@ -132,6 +148,40 @@ void pollPipelineConfigs() {
     }
   }
 
+  private List<Pipeline> fetchHydratedPipelines() {
+    List<Pipeline> rawPipelines = front50.getPipelines();
+    if (rawPipelines == null) {
+      return Collections.EMPTY_LIST;
+    }
+
+    Map<Boolean, List<Pipeline>> partitionedByTemplated = rawPipelines.stream()
```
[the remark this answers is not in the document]
ajordens: Agreed, it may make sense to just pull the hydration into a class of its own and tests to verify that a single orca call (or problem with a pipeline) will not impact the entire pipeline cache.

The existing catch block doesn't entirely include the new behavior, there's a `front50.errors` metric but we've now introduced some calls to orca.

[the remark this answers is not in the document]
jtk54: Simplified this bad boy and added metrics for the orca calls.

echo-pipelinetriggers/src/main/java/com/netflix/spinnaker/echo/pipelinetriggers/PipelineCache.java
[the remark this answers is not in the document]
jtk54: Done.

jtk54:
Limited the MPT planning to v2 and simplified the plan endpoint to do the minimal amount of work in this PR: https://github.com/spinnaker/orca/pull/2777. V1 inherited triggers have apparently never worked so I'm hesitant to change that now since folks haven't been complaining.

echo-core/src/main/java/com/netflix/spinnaker/echo/services/Front50Service.java
```diff
@@ -1,14 +1,15 @@
...
+import java.util.Map;
 import retrofit.http.*;
 
 import java.util.List;
 
 public interface Front50Service {
   @GET("/pipelines?restricted=false")
   @Headers("Accept: application/json")
-  List<Pipeline> getPipelines();
+  List<Map> getPipelines(); // Return Map here so we don't throw away MPT attributes.
```
ezimanyi: Casting would be safer if it were `List<Map<String, Object>>`  Likewise for the code in `PipelineCache`, `Map` -> `Map<String, Object>`.

echo-pipelinetriggers/src/main/java/com/netflix/spinnaker/echo/pipelinetriggers/PipelineCache.java
```diff
@@ -132,6 +145,40 @@ void pollPipelineConfigs() {
     }
   }
 
+  private List<Pipeline> fetchHydratedPipelines() {
+    List<Map> rawPipelines = front50.getPipelines();
+    if (rawPipelines == null) {
+      return Collections.EMPTY_LIST;
```
ezimanyi: nit: `Collections.emptyList()` uses type inference to return `List<Pipeline>` whereas `Collections.EMPTY_LIST` is a raw list.  Not super important for an empty list...

echo-core/src/main/java/com/netflix/spinnaker/echo/services/Front50Service.java
jtk54: Done.

echo-pipelinetriggers/src/main/java/com/netflix/spinnaker/echo/pipelinetriggers/PipelineCache.java
jtk54: Done.

echo-core/src/main/java/com/netflix/spinnaker/echo/services/Front50Service.java
ajordens: We've previously discussed having `@JsonAnyGetter` and `@JsonAnySetter` on `Pipeline` which is arguably preferable to falling back to it just being a `Map`? 

Believe we've sorta just pushed this off until we ran into it again.

echo-pipelinetriggers/src/main/java/com/netflix/spinnaker/echo/pipelinetriggers/PipelineCache.java
```diff
@@ -132,6 +145,40 @@ void pollPipelineConfigs() {
...
+        p.getOrDefault("schema", "").equals("v2");
+    };
+
+    return rawPipelines.stream()
+      .filter(Objects::nonNull)
+      .map(p -> {
+        if (isV2Pipeline.test(p)) {
+          try {
+            Map plannedPipeline = orca.v2Plan(p); // Catch in a local variable so we can increment metrics.
+            registry.counter("orca.requests").increment();
```
ajordens: Upon closer inspection of our existing metrics ... this isn't actually necessary, we've already got `okhttp.requests` tracking inter-service success and failures. 

This isn't adding anything on top of that so could just kill it (I was otherwise going to suggest one planning-specific metric that had a tag indicating success vs failure, rather than two metrics).

jtk54: Done.

echo-core/src/main/java/com/netflix/spinnaker/echo/services/Front50Service.java
jtk54: Looking into the JsonAny* stuff.

ezimanyi:
Yeah, it's probably time to add `@JsonAnySetter` `@JsonAnyGetter` to the echo pipeline model

jtk54:
Neither the `@JsonAny*` annotations or extending `HashMap` work for making Echo's `Pipeline` class function properly with arbitrary key-value pairs coming from Front50 and Orca. I suspect it's interplay with all the annotations (e.g. `@JsonDeserialize`) making this not function properly. Since the triggers are now handled through Echo, I think it's probably high time we just kill the separate representations of `Pipeline` everywhere and explicitly define one class in Kork that is used everywhere.

That being said, I think that's beyond the scope of this PR. In the short term we can either leave the current implementation as a `Map` for flexibility or just add the set of attributes I need for MPTv2 concretely into the `Pipeline` and avoid worrying about extensibility.

jtk54:
Opened https://github.com/spinnaker/spinnaker/issues/4199 to track the Pipeline factorization into Kork.

jtk54:

PR #509, opened 2019-04-05
feat(artifacts): populate artifacts for manual trigger

emjburns:
Make config for build and artifact services explicit.
Add a function to resolve artifacts.

If a manual trigger body looks like:
```
"trigger": {
	    "type": "manual",
	    "dryRun": false,
	    "user": "emburns@netflix.com",
	    "artifacts": [{
        "name": "my-package",
        "version": "v1.1.1",
        "location": "artifactory"
      }]
		}
``` 
Then we will try to resolve the artifact. If that artifact doesn't exist, or you don't have any providers which match the location in `igor`, then the artifact won't be found and it will be removed from the manual trigger. 

@jkschneider for the manual artifactory trigger you'll need to implement lookup logic in `igor` and the ui bits that will add this half fleshed out artifact into the manual trigger.

@ezimanyi thoughts on if using the artifacts field on the trigger to provide a partially fleshed out artifact is the right strategy? I thought it was okay because it's pretty clear to the user and easy to reason about. It also gives us all the right fields to parse many artifacts from different places.

### DIFF
--- a/echo-pipelinetriggers/src/main/java/com/netflix/spinnaker/echo/pipelinetriggers/eventhandlers/ManualEventHandler.java
+++ b/echo-pipelinetriggers/src/main/java/com/netflix/spinnaker/echo/pipelinetriggers/eventhandlers/ManualEventHandler.java
@@ -17,6 +17,8 @@
 package com.netflix.spinnaker.echo.pipelinetriggers.eventhandlers;
 
 import com.fasterxml.jackson.databind.ObjectMapper;
+import com.google.common.base.Strings;
+import com.netflix.spinnaker.echo.artifacts.ArtifactInfoService;
 import com.netflix.spinnaker.echo.build.BuildInfoService;
 import com.netflix.spinnaker.echo.model.Event;
 import com.netflix.spinnaker.echo.model.Pipeline;
@@ -25,8 +27,12 @@ import com.netflix.spinnaker.echo.model.trigger.BuildEvent;
 import com.netflix.spinnaker.echo.model.trigger.ManualEvent;
 import com.netflix.spinnaker.echo.model.trigger.ManualEvent.Content;
 import com.netflix.spinnaker.kork.artifacts.model.Artifact;
+import com.netflix.spinnaker.kork.web.exceptions.NotFoundException;
 import lombok.RequiredArgsConstructor;
 import org.apache.commons.lang3.StringUtils;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Component;
 
 import java.util.*;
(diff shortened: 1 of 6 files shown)


echo-pipelinetriggers/src/main/java/com/netflix/spinnaker/echo/pipelinetriggers/eventhandlers/ManualEventHandler.java
```diff
@@ -74,25 +92,58 @@ private boolean pipelineMatches(String application, String nameOrId, Pipeline pi
...
+      Artifact artifact =  objectMapper.convertValue(a, Artifact.class);
+
+      if (Strings.isNullOrEmpty(artifact.getName()) ||
+        Strings.isNullOrEmpty(artifact.getVersion()) ||
+        Strings.isNullOrEmpty(artifact.getLocation())) {
+        log.error("Artifact does not have enough information to fetch. " +
+          "Artifact must contain name, version, and location.");
+      } else {
+        try {
+          Artifact resolvedArtifact = artifactInfoService.get()
```
ezimanyi: What happens here if I send along a fully-qualified artifact already?  It looks like we won't try to resolve it (as it won't have `location` set) but that we will remove it from the received artifacts.  Should the behavior in that case be to send it along unmodified?

emjburns: Ya, that probably isn't good. I'll change it so that if we can resolve the artifact we replace it, otherwise we leave it alone.

echo-pipelinetriggers/src/main/java/com/netflix/spinnaker/echo/pipelinetriggers/eventhandlers/ManualEventHandler.java
```diff
@@ -74,25 +92,58 @@ private boolean pipelineMatches(String application, String nameOrId, Pipeline pi
...
+            .getArtifactByVersion(artifact.getLocation(), artifact.getName(), artifact.getVersion());
+          if (resolvedArtifact != null) {
```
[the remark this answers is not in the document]
emjburns: I think this is something I just encountered when I was failing to `Mock` correctly in my tests. I'll remove

echo-pipelinetriggers/src/main/java/com/netflix/spinnaker/echo/pipelinetriggers/eventhandlers/ManualEventHandler.java
emjburns: Fixed to send along unchanged.

emjburns:
@ezimanyi to summarize our offline conversation: this behavior mimics the CI functionality: send us identifiable info, and we will populate the full artifact if possible. I think this fits in nicely now, and in the future if we want to provide different information for artifacts that can also fit in.

PR #510, opened 2019-04-08
fix(pipeline): if map -> Pipeline fails, log and move on

emjburns:
For some reason we had a malformed pipeline with a null name (old). After this PR (https://github.com/spinnaker/echo/pull/504/files#diff-123d58ccbba1de00453419d04e10722e) echo wouldn't start up with that. This adds some logging and filtering to continue ignoring that malformed pipeline, just in case...

### DIFF
--- a/echo-pipelinetriggers/src/main/java/com/netflix/spinnaker/echo/pipelinetriggers/PipelineCache.java
+++ b/echo-pipelinetriggers/src/main/java/com/netflix/spinnaker/echo/pipelinetriggers/PipelineCache.java
@@ -19,6 +19,7 @@ package com.netflix.spinnaker.echo.pipelinetriggers;
 import static java.time.Instant.now;
 
 import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.exc.InvalidDefinitionException;
 import com.netflix.spectator.api.Registry;
 import com.netflix.spectator.api.patterns.PolledMeter;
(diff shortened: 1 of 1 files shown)


echo-pipelinetriggers/src/main/java/com/netflix/spinnaker/echo/pipelinetriggers/PipelineCache.java
```diff
@@ -170,7 +168,15 @@ void pollPipelineConfigs() {
         }
       })
       .filter(m -> !m.isEmpty())
-      .map(m -> objectMapper.convertValue(m, Pipeline.class))
+      .map(m -> {
+        try {
+          return objectMapper.convertValue(m, Pipeline.class);
+        } catch (Exception e) {
```
jtk54: Since there is another try/catch in this block, does it make sense to break this out into a helper `planV2Template` or something similar and wrap all the processing in a try/catch there for readability?

emjburns: Extracting to two functions for readability. I still want to keep the functions separate with their own try/catch, though, because I like that the function names explicitly state what they're going to do and handle errors in the way they want.

jtk54: :+1:

ezimanyi:

PR #512, opened 2019-04-10
fix(mpt): fix planing of v1 pipelines for manual triggers

emjburns:
We saw that manual triggering of V1 pipelines stopped working because schema wasn't being set correctly.

More specifically, echo gets a pipeline and then calls out to orca to plan. On this conversion (https://github.com/spinnaker/orca/blob/master/orca-pipelinetemplate/src/main/java/com/netflix/spinnaker/orca/pipelinetemplate/PipelineTemplatePreprocessor.kt#L62) the `objectMapper` is not correctly pulling out the `schema` field from the `config` map (https://github.com/spinnaker/orca/blob/master/orca-pipelinetemplate/src/main/java/com/netflix/spinnaker/orca/pipelinetemplate/TemplatedPipelineRequest.java#L42). This used to not be a problem because in echo `Pipeline` did not have `schema` field (introduced in https://github.com/spinnaker/echo/pull/504/files). Adding `schema` to the echo definition of pipelines without also pulling out the `schema` from `config` seemed to lead to `schema` never getting extracted, and v1 pipelines failing to manually trigger.

This fixes that issue.

### DIFF
--- a/echo-model/src/main/java/com/netflix/spinnaker/echo/model/Pipeline.java
+++ b/echo-model/src/main/java/com/netflix/spinnaker/echo/model/Pipeline.java
@@ -100,5 +100,13 @@ public class Pipeline {
 
   @JsonPOJOBuilder(withPrefix = "")
   public static final class PipelineBuilder {
+    @JsonProperty("config")
+    private void unpackConfig(Map<String, Object> config) {
+      if (config == null) {
+        return;
+      }
+      this.config = config;
+      schema = (String) config.get("schema");
+    }
   }
 }


ajordens:
Am I correct in assuming that we couldn't just have a migrator (or front50) properly set a default v1 schema in the expected location?

This change otherwise seems fine to me.

echo-model/src/main/java/com/netflix/spinnaker/echo/model/Pipeline.java
```diff
@@ -100,5 +100,13 @@
 
   @JsonPOJOBuilder(withPrefix = "")
   public static final class PipelineBuilder {
+    @JsonProperty("config")
+    private void unpackConfig(Map<String, Object> config) {
```
marchello2000: how does this get called? is this `jackson` "magic"?

marchello2000: maybe more appropriate/idiomatic to call this method `setConfig`?

marchello2000: ah, this is symmetric to the one in `orca`, got it. still slightly prefer the `set` idiom

marchello2000:
> Am I correct in assuming that we couldn't just have a migrator (or front50) properly set a default v1 schema in the expected location?
> 
> This change otherwise seems fine to me (but I don't consider myself an expert on templates)

I believe it's true for this issue. But it's probably a good idea to run the migration since the lack of `schema=v1` has been an issue in the past

marchello2000:
@emjburns thanks for fixing this!

emjburns:
@ajordens The problem is that `schema` is defined in the `config` section. There seems to be nothing wrong with the templates themselves, so no need for a migrator. It's just the object serialization/deserialization changed with the addition of new fields so the same jackson bits need to be added. 

For reference, this is what orca receives in the `/plan` call:

```
{
    "application": "emburnstest",
    "config": {
        "schema": "1",
        "pipeline": {
            "name": "mpt",
            "application": "emburnstest",
            "pipelineConfigId": "519569ab-5b6a-424d-9c7c-b3538582a6ff",
            "template": {
                "source": "spinnaker://sample-wait"
            },
            "variables": {}
        },
        "configuration": {
            "inherit": ["parameters", "expectedArtifacts", "triggers"]
        }
    },
    "name": "mpt",
    "id": "519569ab-5b6a-424d-9c7c-b3538582a6ff",
    "parallel": false,
    "disabled": false,
    "limitConcurrent": true,
    "keepWaitingPipelines": false,
    "plan": true,
    "respectQuietPeriod": false,
    "type": "templatedPipeline",
    "stages": [],
    "notifications": [],
    "receivedArtifacts": [],
    "expectedArtifacts": [],
    "trigger": {
        "type": "manual",
        "enabled": false,
        "user": "emburns@netflix.com",
        "rebake": false,
        "dryRun": false
    }
}
```

echo-model/src/main/java/com/netflix/spinnaker/echo/model/Pipeline.java
emjburns: I'm going to keep it the same as we have in orca. Also, "unpack" to me implies that it's doing more than setting the field, it's also doing other things.

emjburns: And yes, when the object is deserialized this gets magically called.

ajordens:
Sure, my point about the migration (or something in front50) is to align the location and expectation for `schema` to avoid these sorts of issues.

We've had no end of problems with this model being duplicated in `echo` (and a couple user impacting problems related to schema this week).

brantburnett:
@emjburns 

Will this fix be cherry-picked into 1.13?

jtk54:

jtk54:
@brantburnett Thanks for the reminder. I think we missed the first patch this week but this should be available within a week or so.

PR #513, opened 2019-04-10
fix(triggers): Better error handling for trigger failures

marchello2000:
If planning a pipeline fails, make sure to log a metric


marchello2000:
@emjburns feel free to roll into your PR if you prefer them to go together

echo-pipelinetriggers/src/main/java/com/netflix/spinnaker/echo/pipelinetriggers/orca/PipelineInitiator.java
```diff
@@ -84,35 +84,40 @@ public void initialize() {
 
   public void startPipeline(Pipeline pipeline) {
     if (enabled) {
-      if (pipeline.getTrigger() != null &&
+      try {
```
marchello2000: diff with ignore white-space

marchello2000:
thanks @ezimanyi

PR #514, opened 2019-04-10
refactor(logging): Log better messages when events fail

marchello2000:
Currently we log the whole event first which can be quite large and thus overloads our log provider (truncating the message)
Split the log into parts such as the important bits are first and the rest can get truncated

### DIFF
--- a/echo-rest/src/main/groovy/com/netflix/spinnaker/echo/events/RestEventListener.groovy
+++ b/echo-rest/src/main/groovy/com/netflix/spinnaker/echo/events/RestEventListener.groovy
@@ -17,6 +17,7 @@
 package com.netflix.spinnaker.echo.events
 
 import com.fasterxml.jackson.databind.ObjectMapper
+import com.netflix.spectator.api.Registry
 import com.netflix.spinnaker.echo.config.RestUrls
 import com.netflix.spinnaker.echo.model.Event
 import groovy.util.logging.Slf4j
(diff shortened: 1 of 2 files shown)


echo-rest/src/main/groovy/com/netflix/spinnaker/echo/events/RestEventListener.groovy
```diff
@@ -36,18 +37,27 @@ class RestEventListener implements EchoEventListener {
...
 
+  Registry registry
+
   @Value('${rest.defaultEventName:spinnaker_events}')
   String eventName
 
   @Value('${rest.defaultFieldName:payload}')
   String fieldName
 
+  @Autowired
```
emjburns: <3

echo-rest/src/main/groovy/com/netflix/spinnaker/echo/events/RestEventListener.groovy
```diff
@@ -72,7 +82,12 @@ class RestEventListener implements EchoEventListener {
         }
         service.client.recordEvent(sentEvent)
       } catch (e) {
-        log.error("Could not send event ${event} to ${service.config.url}", e)
+        log.error("Could not send event source=${event?.details?.source}, type=${event?.details?.type} to ${service.config.url}.\n" +
+          "Event details: ${event}", e)
+
+        if (registry != null) {
```
emjburns: Registry should never be null here.

echo-rest/src/test/groovy/com/netflix/spinnaker/echo/events/RestEventListenerSpec.groovy
```diff
@@ -26,7 +26,7 @@ import spock.lang.Subject
 class RestEventListenerSpec extends Specification {
 
   @Subject
-  RestEventListener listener = new RestEventListener()
+  RestEventListener listener = new RestEventListener(null, null, null)
```
emjburns: Add the `NoopRegistry()` as your last argument here

marchello2000: cool, thanks

PR #515, opened 2019-04-12
fix(retrofit): make sure all retrofit clients use Jackson

marchello2000:
GSON is the default converter. It, for some reason, takes a perfectly valid JSON
with integers and parses them as doubles (sometimes). Jackson doesn't do that.
Jackson is better, use it.


marchello2000:
@robzienert and @jonsie helped fix this, thanks guys!
@ezimanyi let me know if you have any thoughts on this conversion. I think our longer-term goal is to pull out all the retrofit client into kork so that it's consistent everywhere, but until then...

emjburns:
I haven't done any testing of this but it looks amazing. Great sleuthing. LGTM

marchello2000:
> I haven't done any testing of this but it looks amazing. Great sleuthing. LGTM

I just tested that one case that was causing us issues. I will have it run against prestaging and see what happens

PR #518, opened 2019-04-13
fix(triggers): fetch latest config from front50 on impending trigger

dreynaud:
Fixes the following issue:
* save a pipeline through deck, wait for the UI to display "In Sync With
Server" (as reported by orca's MonitorFront50Task)
* trigger the pipeline manually after a short period of time
* the execution actually references the old pipeline configuration

This is because echo references the pipeline config from its in-memory
cache, not the version that was most recently saved to front50.

The window of inconsistency could be arbitrarily long, depending on echo's
PipelineCache polling configuration and front50's own caching.

This change fixes the issue by fetching the latest config from front50
just before the pipeline execution request is submitted to orca (in
TriggerMonitor::triggerMatchingPipelines)

### DIFF
--- a/echo-core/src/main/java/com/netflix/spinnaker/echo/services/Front50Service.java
+++ b/echo-core/src/main/java/com/netflix/spinnaker/echo/services/Front50Service.java
@@ -15,6 +15,10 @@ public interface Front50Service {
   @Headers("Accept: application/json")
   List<Pipeline> getPipelines(@Path("application") String application);
 
+  // either an empty list or a singleton of a raw pipeline config
+  @GET("/pipelines/{pipelineId}/history?limit=1")
+  List<Map<String, Object>> getLatestVersion(@Path("pipelineId") String pipelineId);
+
   @POST("/graphql")
   @Headers("Accept: application/json")
   GraphQLQueryResponse query(@Body GraphQLQuery body);
--- a/echo-pipelinetriggers/src/main/java/com/netflix/spinnaker/echo/pipelinetriggers/PipelineCache.java
+++ b/echo-pipelinetriggers/src/main/java/com/netflix/spinnaker/echo/pipelinetriggers/PipelineCache.java
@@ -35,6 +35,7 @@ import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
 import java.util.function.Predicate;
 import java.util.stream.Collectors;
+import java.util.stream.Stream;
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
 import javax.annotation.PostConstruct;
@@ -131,7 +132,7 @@ public class PipelineCache implements MonitoredPoller {
     try {
       log.debug("Getting pipelines from Front50...");
       long start = System.currentTimeMillis();
-      pipelines = decorateTriggers(fetchHydratedPipelines());
+      pipelines = fetchHydratedPipelines();
 
       lastPollTimestamp = now();
       registry.counter("front50.requests").increment();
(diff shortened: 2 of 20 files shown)


echo-pipelinetriggers/src/main/java/com/netflix/spinnaker/echo/pipelinetriggers/PipelineCache.java
```diff
@@ -142,25 +143,68 @@ void pollPipelineConfigs() {
...
+      .filter(Optional::isPresent)
+      .map(Optional::get)
       .collect(Collectors.toList());
   }
 
+  // looks up the latest version in front50 of a (potentially stale) pipeline config from the cache
+  // if anything fails during the refresh, we fall back to returning the cached version
+  public Pipeline refresh(Pipeline cached) {
+    try {
+      List<Map<String, Object>> latestVersion = front50.getLatestVersion(cached.getId());
```
cfieber: Think it is worth a timer or counter around this call? I'm curious about how frequently we will end up calling this... if it turns out to be a lot we might need to revisit this with something that loads in bulk for all the matchingPipelines (but maybe hopefully not because this way is definitely clearer and simpler)

echo-core/src/main/java/com/netflix/spinnaker/echo/services/Front50Service.java
```diff
@@ -15,6 +15,10 @@
   @Headers("Accept: application/json")
   List<Pipeline> getPipelines(@Path("application") String application);
 
+  // either an empty list or a singleton of a raw pipeline config
```
jtk54: Under what conditions is a pipeline config returned here? Seems to me like we'd always be making live calls instead of looking things up in the cache.

echo-pipelinetriggers/src/main/java/com/netflix/spinnaker/echo/pipelinetriggers/monitor/TriggerMonitor.java
```diff
@@ -77,6 +78,10 @@ private void triggerMatchingPipelines(T event) {
       List<Pipeline> allPipelines = pipelineCache.getPipelinesSync();
       List<Pipeline> matchingPipelines = eventHandler.getMatchingPipelines(event, allPipelines);
       matchingPipelines.stream()
+        .map(pipelineCache::refresh) // make sure we don't start an execution of a stale pipeline config
+        .map(p -> eventHandler.withMatchingTrigger(event, p)) // there is a possibility the refreshed pipeline no longer matches the trigger
```
ezimanyi: It looks like this is going to cause us to re-do all the trigger matching logic twice on every pipeline start.  In a lot of cases, there is a reasonable amount of logic in `withMatchingTrigger`.  For example, Jenkins triggers call back to `igor` to get information on the build that just completed; we'll now be asking igor for information on the completed build twice per pipeline.  Is that extra traffic/duplicated computation a concern?

ezimanyi: If it turns out we don't want all this extra computation on each trigger, one option would be to update `ManualEventHandler` to always fetch the pipeline being executed from `front50` before processing the pipeline.  This would solve the manual execution case (which I think is the main goal) without doing any re-computation.

echo-pipelinetriggers/src/main/java/com/netflix/spinnaker/echo/pipelinetriggers/PipelineCache.java
dreynaud: I was thinking about that. Ideally we would be using an instrumented okhttp client that already publishes RPC metrics, that would avoid a proliferation of handmade counters. 

EDIT: yes we do, the timer is `okhttp.requests` and it is tagged by `requestHost`, `statusCode`, etc.

echo-pipelinetriggers/src/main/java/com/netflix/spinnaker/echo/pipelinetriggers/monitor/TriggerMonitor.java
dreynaud: Ah thanks for pointing this out, I didn't realize `withMatchingTrigger` would make more external calls. 

Putting that logic back in `ManualEventHandler` could work (that's where I initially focused) but this seemed like a nice generic place to perform the refresh. Let me think some more about it and come back with something.

echo-core/src/main/java/com/netflix/spinnaker/echo/services/Front50Service.java
dreynaud: I think this always returns a pipeline config. We get a 404 if the pipeline id does not match anything.

jtk54: Is the pipeline config cache being used at all then if we're live calling first?

dreynaud: yes, the pipeline config cache is used first and we are live calling last (just when something is about to be triggered)

jtk54: I missed the first step in my readthrough, makes sense now :+1:

echo-pipelinetriggers/src/main/java/com/netflix/spinnaker/echo/pipelinetriggers/monitor/TriggerMonitor.java
dreynaud: I tried a more general rework but I find the artifacts stuff and igor lookups difficult to grok. I am temporarily admitting defeat and added the refresh on manual triggers only for now, in the interest of fixing the burning issue quickly. @ezimanyi does 9ca5c486b9729c57e5c9da10441efb065e5fd6db look alright to you?

ezimanyi: Yes, that looks great, thanks!

PR #523, opened 2019-04-17
refactor(pubsub): Encapsulate logic for creating events

ezimanyi:
* feat(pubsub): Send pubsub events through eventPropagator 

  Currently all pubsub events are sent directly to the trigger event listener, rather than to the general eventPropagator (which notifies all event listeners). This means that pubsub events can never do anything other than start pipelines. (Other incoming events, such as webhooks or build notifications are sent to the general eventPropagator.)

  In order to allow pubsub messages to do other things than start pipelines, send the resulting event to the eventPropagator so that other listeners can act on these events.

* refactor(pubsub): Encapsulate logic for creating events 

  The logic for extracting artifacts from pubsub events and for creating events from the message is embedded in the logic for handling these messages. In order to allow more general events to be created from pubsub messages, create an interface EventCreator that handles creating an event from a pubsub MessageDescription.

  The initial implementation of EventCreator will extract artifacts and create exactly the same event as prior to this change, allowing us to remove the code from the pubsub subscribers and handler.

  When setting up a subscription it is now possible to pass in a custom EventCreator so that messages read from that subscription create arbitrary event types.

### DIFF
--- a/echo-pubsub-core/src/main/java/com/netflix/spinnaker/echo/pubsub/PubsubEventCreator.java
+++ b/echo-pubsub-core/src/main/java/com/netflix/spinnaker/echo/pubsub/PubsubEventCreator.java
@@ -46,6 +46,8 @@ public class PubsubEventCreator implements EventCreator {
   }
 
   public Event createEvent(MessageDescription description) {
+    log.trace("Processing pubsub event with payload {}", description.getMessagePayload());
+
     try {
       description.setArtifacts(parseArtifacts(description.getMessagePayload()));
     } catch (FatalTemplateErrorsException e) {
--- a/echo-pubsub-core/src/main/java/com/netflix/spinnaker/echo/pubsub/PubsubEventCreator.java
+++ b/echo-pubsub-core/src/main/java/com/netflix/spinnaker/echo/pubsub/PubsubEventCreator.java
@@ -26,6 +26,8 @@ import com.netflix.spinnaker.echo.pipelinetriggers.eventhandlers.PubsubEventHand
 import com.netflix.spinnaker.echo.pubsub.model.EventCreator;
 import com.netflix.spinnaker.kork.artifacts.model.Artifact;
 import lombok.extern.slf4j.Slf4j;
+import org.springframework.util.CollectionUtils;
+import org.springframework.util.StringUtils;
 
 import java.io.IOException;
 import java.util.*;
@@ -44,8 +46,6 @@ public class PubsubEventCreator implements EventCreator {
   }
 
   public Event createEvent(MessageDescription description) {
-    log.debug("Processing pubsub event with payload {}", description.getMessagePayload());
-
     try {
       description.setArtifacts(parseArtifacts(description.getMessagePayload()));
     } catch (FatalTemplateErrorsException e) {
@@ -80,8 +80,7 @@ public class PubsubEventCreator implements EventCreator {
     }
     List<Artifact> artifacts = messageArtifactTranslator.get().parseArtifacts(messagePayload);
     // Artifact must have at least a reference defined.
-    if (artifacts == null || artifacts.size() == 0
-      || artifacts.get(0).getReference() == null || artifacts.get(0).getReference().equals("")) {
+    if (CollectionUtils.isEmpty(artifacts) || StringUtils.isEmpty(artifacts.get(0).getReference())) {
       return Collections.emptyList();
     }
     return artifacts;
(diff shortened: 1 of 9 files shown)


ezimanyi:
The main goal here is to allow us to do something with a pubsub event other than trigger a pipeline.  This PR doesn't actually do that, but enables subscribers to inject the logic for creating events into the callback that happens on each message.

How I plan to use this:
* When users have Google Cloud Build enabled, echo will create a pubsub subscription to listen for build notifications and will create an event of type `googleCloudBuild` for each message.  (It will continue to create a `pubsub` event as well as users may be relying on this for existing triggers; that's the reason I needed to allow a list of `EventCreators`.)
* I'll add an event listener for `googleCloudBuild` events that will forward the event along to `igor` for tracking.

I split the first commit out as it's worth highlighting---as far as I can tell there's no reason why pubsub events directly call the trigger listener (rather than the event propagator) but this will change the behavior there.

echo-pubsub-core/src/main/java/com/netflix/spinnaker/echo/pubsub/PubsubEventCreator.java
```diff
@@ -0,0 +1,89 @@
...
+public class PubsubEventCreator implements EventCreator {
+  private final ObjectMapper objectMapper = new ObjectMapper();
+  private final Optional<MessageArtifactTranslator> messageArtifactTranslator;
+
+  public PubsubEventCreator(Optional<MessageArtifactTranslator> messageArtifactTranslator) {
+    this.messageArtifactTranslator = messageArtifactTranslator;
+  }
+
+  public Event createEvent(MessageDescription description) {
+    log.debug("Processing pubsub event with payload {}", description.getMessagePayload());
```
emjburns: While you're refactoring, we should maybe hide this behind a debug log flag to get it hidden even further? This log statement generates a ton of log messages, which are not really helpful after things are working.

echo-pubsub-core/src/main/java/com/netflix/spinnaker/echo/pubsub/PubsubEventCreator.java
```diff
@@ -0,0 +1,89 @@
...
+  }
+
+  private List<Artifact> parseArtifacts(String messagePayload) {
+    if (!messageArtifactTranslator.isPresent()) {
+      return Collections.emptyList();
+    }
+    List<Artifact> artifacts = messageArtifactTranslator.get().parseArtifacts(messagePayload);
+    // Artifact must have at least a reference defined.
+    if (artifacts == null || artifacts.size() == 0
+      || artifacts.get(0).getReference() == null || artifacts.get(0).getReference().equals("")) {
```
emjburns: weird indentation

echo-pubsub-core/src/main/java/com/netflix/spinnaker/echo/pubsub/PubsubEventCreator.java
ezimanyi: You mean guard the whole statement by `log.isDebugEnabled()` to avoid evaluating it at all in most cases?  This is already at the debug level, so shouldn't be generating messages unless I'm missing something.

emjburns: I mean have some flag in the pubsub config that turns on this log statement. I think that this log proportionally overwhelms all other debug logs. We tend to run with debug logs enabled, except in echo, because of this statement. It would be nice to have to opt into this very very debug log statement. I don't know if other people experience this pain.

ezimanyi: Oh, I see---I didn't realize you ran with `debug` enabled, makes sense now.  I'm also happy to just remove the line, do you see any issue in just removing it?

emjburns: I don't! I'd prefer that I think. But I think @jtk54 might have opinions on that

echo-pubsub-core/src/main/java/com/netflix/spinnaker/echo/pubsub/PubsubEventCreator.java
ezimanyi: Fixed by changing to null-safe `isEmpty` functions so it fits on one line now.

echo-pubsub-core/src/main/java/com/netflix/spinnaker/echo/pubsub/PubsubEventCreator.java
jtk54: This debug message has certainly been useful to us for troubleshooting when we don't have direct access to the spinnaker instance (i.e. most cases). Most users don't have the capability to add logging instrumentation when they need for debugging purposes, so I'm inclined to keep this in. Should we just move it to TRACE?

ezimanyi: Sounds good, I'll make it `TRACE` instead of just removing it.

PR #524, opened 2019-04-18
fix(pubsub): Don't log all pubsub events

ezimanyi:
We're now logging all pubsub events in AbstractNotificationAgent; given the volume of these events, turn off this logging for pubsub.


echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/notification/AbstractEventNotificationAgent.groovy
```diff
@@ -53,7 +53,7 @@ abstract class AbstractEventNotificationAgent implements EchoEventListener {
 
   @Override
   void processEvent(Event event) {
-    if (log.isDebugEnabled() && mapper != null) {
+    if (log.isDebugEnabled() && mapper != null && !event.getDetails().getType().equals("pubsub")) {
```
ezimanyi: Not worried about NPE here as the rest of the function assumes non-null type already

ajordens:
Not object mappering / logging each event sounds reasonable to me ... will defer to @emjburns for the formal +1.

PR #526, opened 2019-04-19
perf(triggers): index triggers based on type

dreynaud:
Instead of iterating over all triggers of all pipelines to look for
pipelines that match an event, PipelineCache indexes triggers by type.
So event handlers only need to iterate over triggers of a supported type
to find matching pipelines.

Also make ArtifactoryEventHandlerTest a Spock test.


echo-pipelinetriggers/src/main/java/com/netflix/spinnaker/echo/pipelinetriggers/PipelineCache.java
```diff
@@ -134,6 +138,9 @@ void pollPipelineConfigs() {
       long start = System.currentTimeMillis();
       pipelines = fetchHydratedPipelines();
 
+      // refresh the triggers view every time we fetch the latest pipelines
+      triggers = extractEnabledTriggersFrom(pipelines);
```
dreynaud: we re-index triggers during every poll cycle

echo-pipelinetriggers/src/main/java/com/netflix/spinnaker/echo/pipelinetriggers/PipelineCache.java
```diff
@@ -178,6 +185,27 @@ void pollPipelineConfigs() {
...
+
+    // When getPipelinesSync returns, this means that we have populated the pipeline cache.
+    // At this point, we don't expect triggers to be null but we check anyway to avoid a
+    // potential race condition.
+    return triggers == null
+      ? extractEnabledTriggersFrom(pipelines)
+      : triggers;
+  }
+
+  private static Map<String, List<Trigger>> extractEnabledTriggersFrom(List<Pipeline> pipelines) {
```
dreynaud: the trigger indexing logic fits neatly in this function

echo-pipelinetriggers/src/main/java/com/netflix/spinnaker/echo/pipelinetriggers/eventhandlers/ManualEventHandler.java
```diff
@@ -93,6 +101,19 @@ public ManualEvent convertEvent(Event event) {
     }
   }
 
+  @Override
+  public List<Pipeline> getMatchingPipelines(ManualEvent event, PipelineCache pipelineCache) throws TimeoutException {
+    if (!isSuccessfulTriggerEvent(event)) {
+      return Collections.emptyList();
+    }
+
+    return pipelineCache.getPipelinesSync().stream()
```
dreynaud: manual events still go over the entire pipeline collection (cc @marchello2000)

echo-pipelinetriggers/src/test/groovy/com/netflix/spinnaker/echo/pipelinetriggers/eventhandlers/EventHandlerSupport.groovy
```diff
@@ -0,0 +1,18 @@
+package com.netflix.spinnaker.echo.pipelinetriggers.eventhandlers
```
ezimanyi: Missing header

echo-pipelinetriggers/src/main/java/com/netflix/spinnaker/echo/pipelinetriggers/eventhandlers/TriggerEventHandler.java
```diff
@@ -58,19 +67,22 @@
...
    */
-  default List<Pipeline> getMatchingPipelines(T event, List<Pipeline> pipelines) {
-    if (isSuccessfulTriggerEvent(event)) {
-      return pipelines.stream()
-        .map(p -> withMatchingTrigger(event, p))
-        .filter(Optional::isPresent)
-        .map(Optional::get)
-        .collect(Collectors.toList());
-    } else {
+  default List<Pipeline> getMatchingPipelines(T event, PipelineCache pipelineCache) throws TimeoutException {
```
dreynaud: this interface change (taking a `PipelineCache` instead of a `List<Pipeline>`) is what causes this PR to be so big, lots of call sites and tests had to be moved to the new interface. A bit unfortunate, but I think it is necessary: a `PipelineCache` can provide both the list of pipelines (for manual events) and the trigger index (for every other event handler)

echo-pipelinetriggers/src/test/groovy/com/netflix/spinnaker/echo/pipelinetriggers/PipelineCacheSpec.groovy
```diff
@@ -114,5 +114,50 @@ class PipelineCacheSpec extends Specification implements RetrofitStubs {
     then:
     notThrown(JsonMappingException)
   }
+
```
emjburns: maybe you should add a test for the happy path?

emjburns: I guess it's kind of in this first one.

echo-pipelinetriggers/src/test/groovy/com/netflix/spinnaker/echo/pipelinetriggers/eventhandlers/ArtifactoryEventHandlerSpec.groovy
```diff
@@ -0,0 +1,291 @@
...
+import spock.lang.Specification
+import spock.lang.Subject
+
+import java.util.Arrays
+import java.util.Collections
+import java.util.List
+
+import static org.assertj.core.api.Assertions.assertThat
+
+class ArtifactoryEventHandlerSpec extends Specification {
```
emjburns: ❤️

emjburns:
👏 👏 👏 wow this looks great.

echo-pipelinetriggers/src/main/java/com/netflix/spinnaker/echo/pipelinetriggers/PipelineCache.java
```diff
@@ -63,6 +63,9 @@
   @Nullable
   private transient List<Pipeline> pipelines;
 
+  @Nullable
+  private transient Map<String, List<Trigger>> triggers;
```
ezimanyi: Should this (and possibly `pipelines` above) be `volatile` instead of (or in addition to) `transient`?  I think that would remove the race condition you note below about `triggers` still being null after calling `getPipelinesSync()` because the update by the other thread would be guaranteed to be visible.

echo-pipelinetriggers/src/main/java/com/netflix/spinnaker/echo/pipelinetriggers/eventhandlers/ManualEventHandler.java
marchello2000: someday, maybe we will have `triggerType = [manual, git, pipeline, etc]` and `invocationType = [manual/etc]` and we will be able to avoid this iteration. For now, it makes sense

echo-pipelinetriggers/src/main/java/com/netflix/spinnaker/echo/pipelinetriggers/PipelineCache.java
marchello2000: 1. Why are these things marked `transient`, I don't think this class is ever serialized?
2. I don't know that `volatile` will fix the race - but it's probably a good thing to add anyway. I think it needs to either be a proper lock or we should change the order of how we assign pipeline/trigger. I.e. at https://github.com/spinnaker/echo/pull/526/files?file-filters%5B%5D=.java#diff-123d58ccbba1de00453419d04e10722eR139
```
tempPipelines = fetchHydratedPipelines();
triggers = extractEnabledTriggersFrom(tempPipelines);
pipelines = tempPipelined
```

not sure that I love either of these options

marchello2000: also, maybe rename to `allTriggersByType`?

dreynaud: Oops, I definitely meant `volatile` and not `transient`, good find 😅

After a good long read about the new guarantees in JSR-133... I sort of go back and forth between convincing myself that `volatile` fixes the race condition or not. In any case, I think it's safe to leave the `null` check in place and recompute `triggers` on the fly if that happens, that should be cheap enough.

cfieber: How about storing this in an AtomicReference instead? IMO that is much clearer and easier to get right than the equivalent level of `volatile` and `synchronized`

echo-pipelinetriggers/src/test/groovy/com/netflix/spinnaker/echo/pipelinetriggers/eventhandlers/EventHandlerSupport.groovy
dreynaud: I heard rumors that somebody heard from one of our lawyers that we don't need to do that anymore?

I'll put one in for now but it sure would be nice if we didn't have to do these shenanigans.

ezimanyi: Sounds good, totally agree that I'd prefer not do do these either, but hadn't heard otherwise

echo-pipelinetriggers/src/test/groovy/com/netflix/spinnaker/echo/pipelinetriggers/PipelineCacheSpec.groovy
dreynaud: yes that's how I think of the first test

echo-pipelinetriggers/src/main/java/com/netflix/spinnaker/echo/pipelinetriggers/PipelineCache.java
dreynaud: I don't think `AtomicReference` buys us anything here since we're not doing operations like `compareAndSet`. From the docs (https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/atomic/package-summary.html):

> The memory effects for accesses and updates of atomics generally follow the rules for volatiles, as stated in section 17.4 of The Java™ Language Specification.
> 
> get has the memory effects of reading a volatile variable.
> set has the memory effects of writing (assigning) a volatile variable.

PR #527, opened 2019-04-22
feat(gcb): Notify igor of GCB builds

ezimanyi:
When the configuration contains gcb accounts, create a PubSub listener for each account that forwards any build notifications to igor. (These notifications should also generate the usual PubSub events so that pipelines can be triggered from them.)


echo-notifications/src/test/groovy/com/netflix/spinnaker/echo/notification/GoogleCloudBuildNotificationSpec.groovy
```diff
@@ -0,0 +1,99 @@
...
+
+  def "retries on failure to communicate with igor"() {
+    given:
+    Event event = createEvent("googleCloudBuild")
+
+    when:
+    notificationAgent.processEvent(event)
+
+    then:
+    2 * igorService.updateBuildStatus(ACCOUNT_NAME, BUILD_ID, BUILD_STATUS, { it -> it.in().text == PAYLOAD }) >>
```
jtk54: What is the intended behavior of duplicate build statuses (statii)?

ezimanyi: You mean if we receive a notification twice?  We'll forward along both to igor.   Igor will refuse to update its cache with an older build status (to prevent issues due to out-of-order messages) but will allow a new message with the same status.

PR #529, opened 2019-04-23
fix(triggers): prevent double-triggering for pipelines with >1 triggers

dreynaud:
A pipeline with 2 triggers and an event matching one of them could
actually be triggered twice. The problem is that withMatchingTrigger
would consider the event and all triggers of the pipeline, and it would
be called once for each trigger.

The key fix is to matchTriggerFor(event) in getMatchingPipelines().
As a result, we can change the inputs of getMatchingPipelines to an event
and a matching trigger instead of an event and a pipeline (with all its
triggers).


dreynaud:
I'll merge this to fix the build on our side but please comment away

ezimanyi:
This looks good, though I think there's still a minor regression that this doesn't fix.  If a pipeline has two triggers that *both* match, the pipeline will still be triggered twice even with this change; before your refactor, it would be triggered only once.

There probably needs to be some kind of grouping by pipeline with a `.findFirst()` over triggers (as was happening before the refactor).  I'm not sure how common this edge case is, but is probably worth keeping the behavior consistent with what it was before.

dreynaud:
@ezimanyi I thought this seemed correct but after conferring in my chambers I think you're right, double-triggering might cause some issues for people. I'll update, again...

PR #530, opened 2019-04-23
feat(auth): Propagate `X-SPINNAKER-ACCOUNTS` when triggering pipelines

ajordens:
Echo and externally triggered (via `gate`) pipelines differ in that the
former does not propagate `X-SPINNAKER-ACCOUNTS`, even when a service
account is being used.

When `services.fiat.legacyFallback` is enabled, this PR will fetch
the allowed accounts for either the specified service account name
or `anonymous`.

The goal is to make fallbacks consistent regardless of how a pipeline
was triggered.

### DIFF
--- a/echo-pipelinetriggers/src/main/java/com/netflix/spinnaker/echo/pipelinetriggers/orca/PipelineInitiator.java
+++ b/echo-pipelinetriggers/src/main/java/com/netflix/spinnaker/echo/pipelinetriggers/orca/PipelineInitiator.java
@@ -21,6 +21,10 @@ import com.netflix.spectator.api.Registry;
 import com.netflix.spinnaker.echo.model.Pipeline;
 import com.netflix.spinnaker.echo.pipelinetriggers.QuietPeriodIndicator;
 import com.netflix.spinnaker.echo.pipelinetriggers.orca.OrcaService.TriggerResponse;
+import com.netflix.spinnaker.fiat.model.Authorization;
+import com.netflix.spinnaker.fiat.model.UserPermission;
+import com.netflix.spinnaker.fiat.model.resources.Account;
+import com.netflix.spinnaker.fiat.shared.FiatPermissionEvaluator;
 import com.netflix.spinnaker.fiat.shared.FiatStatus;
 import com.netflix.spinnaker.security.AuthenticatedRequest;
 import com.netflix.spinnaker.security.User;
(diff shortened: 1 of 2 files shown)


ajordens:
(WIP)Needs some better unit tests ... I've done a small bit of local smoke testing but plan to do more.

echo-pipelinetriggers/src/main/java/com/netflix/spinnaker/echo/pipelinetriggers/orca/PipelineInitiator.java
```diff
@@ -150,6 +169,33 @@ private void onOrcaError(Pipeline pipeline, Throwable error) {
...
+      log.error("Unable to fetch permission for {}", user, e);
+    }
+
+    if (userPermission == null) {
+      return Collections.emptySet();
+    }
+
+    return userPermission.getAccounts()
+      .stream()
+      .filter(v -> v.getAuthorizations().contains(Authorization.WRITE))
```
dreynaud: why are we filtering here?

ajordens: I'll toss a comment but this aligns with what `gate` does in [AllowedAccountsSupport](https://github.com/spinnaker/gate/blob/master/gate-core/src/main/groovy/com/netflix/spinnaker/gate/security/AllowedAccountsSupport.java).

dreynaud:
Looks good!

ajordens:
FYI @ezimanyi -- Not sure if there is someone better from the Google side but wanted to run this past you folks.

If `legacyFallback` is disabled (default behavior), the only expected change here will be that we'll send `anonymous` across as the `X-SPINNAKER-USER` if a service account isn't being used. 

This covers spots where we're not currently sending anything and relying on `AuthenticatedRequest.getSpinnakerUser().orElse("anonymous")` sprinkled throughout the code base. 

If `legacyFallback` is enabled, we'll actually make an attempt to fetch accounts when triggering a pipeline. Idea being that this should allow running pipelines to withstand `fiat` being unavailable (rather than the current mass failure that happens).

ezimanyi:
Looks good, retroactive LGTM.

ajordens:
Thanks @ezimanyi

PR #536, opened 2019-04-26
fix(core): Spring does not override default configuration files.

jasonjchu:
In the Spring Boot 2 upgrade, `spring.config.location` overrides all other config locations. The equivalent property is now called `spring.config.additional-location`.
This fix allows the app the to run without additional config files.


robzienert:
Will need to be applied to other services as well

eleftherias:
@robzienert We've made the change for igor as well https://github.com/spinnaker/igor/pull/430

PR #541, opened 2019-05-06
chore(*): Remove stale comments from gradle file

marchello2000:
`quartz` now comes from spring bom (pulled in by kork-bom).
Remove the unnecessary pinning of `quartz` (it doesn't work the way it is anyway)

Note: this means that CRON expressions that use step value of 60 (for mins/secs)
or 24 (for hours) will no longer work (it wasn't valid before but previous `quartz`
was less strict about the validation).

There shouldn't be many users affected (if at all) because
a) these are non standard CRON expression anyway
b) most users "write" their expressions from the UI which wouldn't generate these.

We will have a release note for Spinnaker 1.14 about this.

### DIFF
--- a/echo-scheduler/echo-scheduler.gradle
+++ b/echo-scheduler/echo-scheduler.gradle
@@ -15,9 +15,6 @@
  */
 
 dependencies {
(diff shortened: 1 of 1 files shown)


marchello2000:
(this change is effectively a no-op anyway since quartz version is taken from bom)

PR #542, opened 2019-05-07
feat(scheduler): Add sql quartz support

marchello2000:
**WORK IN PROGRESS** Feedback appreciated

We prefer small, well tested pull requests.


README.md
```diff
@@ -42,4 +40,30 @@ The JVM will then listen for a debugger to be attached on port 8189.  The JVM wi
 the debugger to be attached before starting Echo; the relevant JVM arguments can be seen and
 modified as needed in `build.gradle`.
 
-[//]: # "Only here to retrigger the echo build"
+## Configuring
```
marchello2000: @robzienert what do you think about this doc (it's incomplete for now, obv).
Wanted your opinion that this is right direction before I flesh it out

echo-scheduler/quartz_init.sql
```diff
@@ -0,0 +1,182 @@
+DROP TABLE IF EXISTS QRTZ_FIRED_TRIGGERS;
```
marchello2000: The reason we should make the tables manually is because if we opt to use `QuartzAutoConfigure` - it will blow away and recreate the tables on start which is not what we want to support multiple instances. Perhaps I am missing something and we can configure it more better to only make tables if none are found?

echo-scheduler/src/main/groovy/com/netflix/spinnaker/echo/config/SchedulerConfiguration.groovy
```diff
@@ -124,13 +90,33 @@ class SchedulerConfiguration {
...
+  ) {
+    return new SchedulerFactoryBeanCustomizer() {
+      @Override
+      void customize(SchedulerFactoryBean schedulerFactoryBean) {
+        if (dataSourceOptional.isPresent()) {
+          schedulerFactoryBean.setDataSource(dataSourceOptional.get())
+        }
+
+        Properties props = new Properties()
+        props.put("org.quartz.jobStore.isClustered", "true")
```
marchello2000: couldn't figure out a better way to do this... if I make a new instance of `QuartzProperties` the app fails to start due to multiple bean

echo-scheduler/src/main/groovy/com/netflix/spinnaker/echo/scheduler/actions/pipeline/AutowiringSpringBeanJobFactory.java
```diff
@@ -1,62 +0,0 @@
...
-import org.springframework.context.ApplicationContext;
-import org.springframework.context.ApplicationContextAware;
-import org.springframework.scheduling.quartz.SpringBeanJobFactory;
-
-/**
- * Taken from https://dzone.com/articles/spring-and-quartz-integration-that-works-together
- *
- * <p>Allows us to have Quartz jobs play nicely with Spring DI
- */
-public final class AutowiringSpringBeanJobFactory extends SpringBeanJobFactory
```
marchello2000: spring boot 2 has this already, so yay, we no longer need this

README.md
robzienert: this is good. We’ll want to have more in-depth docs about setup and operations in spinnaker.io - but I do like a crash course in the README... we should probably do the same for other services.

README.md
```diff
@@ -42,4 +40,32 @@ The JVM will then listen for a debugger to be attached on port 8189.  The JVM wi
...
+`scheduler.enabled`
+
+### Sample deployment
+Here are examples of what configurations you can deploy `echo`.
+
+|                   | Using in-memory           | Using SQL |
+|-------------------|---------------------------|-----------|
+|**Server Group 1** |3x `echo`                  | 3x `echo` with `echo-scheduler`
+|**Server Group 2** |1x `echo-scheduler`        | 1x `echo-missed-scheduler`
+|**Server Group 3** |1x `echo-missed-scheduler` | n/a
```
robzienert: I don’t think the SQL version would need the missed scheduler cluster either? Quartz automatically handles misfires! Worth testing to see if we get similar behavior (we should).

echo-scheduler/quartz_init.sql
robzienert: Ha. That doesn’t sound right. If we’re going to maintain a schema within the service, we should probably use liquibase to apply it so the operator experience is similar for other SQL services. Liquibase fortunately supports SQL migration files as well, so this shouldn’t be too much of a lift.

echo-scheduler/src/main/groovy/com/netflix/spinnaker/echo/config/SchedulerConfiguration.groovy
robzienert: I don’t follow? Why would you need a new `QuartzProperties`?

README.md
marchello2000: well, there is a caveat... Our "missfire" handling is way more elaborate... e.g it will only refire if the misfire is within a given window (which I believe we can configure with quartz). But we also actually ask orca if the execution happened, so in theory the missed scheduler will catch other random failures not just quartz misfires... 
For now, I think we keep it as is (as in, keep the `MissedPipelineTriggerCompensationJob` - since it's controlled by property anyway) After it runs in our env, we can talk through if we really need it - would certainly simplify the operation of `echo` if it's gone
@dreynaud will have opinions for sure so adding him

echo-scheduler/quartz_init.sql
marchello2000: makes sense, will do that

echo-scheduler/src/main/groovy/com/netflix/spinnaker/echo/config/SchedulerConfiguration.groovy
marchello2000: to specify customization for the quartz scheduler... 
If you think this is ok then certainly happy to keep as is

README.md
```diff
@@ -42,4 +40,32 @@ The JVM will then listen for a debugger to be attached on port 8189.  The JVM wi
 the debugger to be attached before starting Echo; the relevant JVM arguments can be seen and
 modified as needed in `build.gradle`.
 
-[//]: # "Only here to retrigger the echo build"
+## Configuring
+echo can run in two modes: in-memory and SQL.
+
+
```
dreynaud: extra newline. This is going to cause massive waste if people want to print this out.

README.md
```diff
@@ -42,4 +40,32 @@ The JVM will then listen for a debugger to be attached on port 8189.  The JVM wi
...
+**In-memory** mode keeps all CRON trigger information in RAM. While this is simpler to configure (this is the default) the in-memory mode doesn't provide for any redundancy because it requires that a single instance of echo scheduler be running. If/when that instance goes down, CRON triggers will not fire.
```
dreynaud: I would be extra explicit here: "If there are multiple instances, they will all attempt to start executions for a given CRON trigger. There is no locking, leader election or any other kind of coordination between scheduler instances using the in-memory mode."

README.md
```diff
@@ -42,4 +40,32 @@ The JVM will then listen for a debugger to be attached on port 8189.  The JVM wi
...
+
+**SQL** mode keeps all CRON trigger information in a single SQL database. This allows for multiple echo scheduler instances to run providing redundancy (only one instance will trigger a given CRON).
+
+For SQL you will need to initialize the database, an initialization script for MySql is provided [here]()
```
dreynaud: `MySQL`

README.md
dreynaud: Happy to experiment but yes, let's make sure we have parity with the current missed scheduler before pulling the plug

echo-scheduler/src/main/groovy/com/netflix/spinnaker/echo/config/SchedulerConfiguration.groovy
robzienert: I don’t think that will be necessary. Since we’re using Spring Boot for autoconfiguration, we’ll actually be exposing `spring.quartz.*` configuration options to operators automagically!

README.md
robzienert: Ok. With that in mind, we might want to disable quartz performing misfire remediation entirely. I’d be keen to see if we can provide custom logic for how quartz handles misfires. I don’t like maintaining additional stacks; one of the objectives of this quartz work is to reduce operational overhead. Let’s see if:

1. Quartz can support custom behavior for resolving misfires.
2. If not, evaluate how we can roll our custom behavior into the single stack.

robzienert: It looks like a `TriggerListener` could potentially do the trick for us.

echo-scheduler/quartz_init.sql
marchello2000: Fixed

README.md
marchello2000: yes, we already have a trigger listener - and it always instructs quartz not to "re-fire" but yes, in a subsequent PR I will address the quartz/missfire/missed pipeline issue

README.md
marchello2000: 👍

README.md
marchello2000: this is just the markdown way of specifying a new paragraph

dreynaud: you only need one blank line, not two though

marchello2000: fine, FINE. I will save some trees

marchello2000:
going to close this PR - this is the new one: https://github.com/spinnaker/echo/pull/561

PR #545, opened 2019-05-10
fix(email): fix email parser to not fail on multiple addresses

marchello2000:
If a pipeline specified two email addresses in the notification that email notification
will not be sent.
This change moves the email address splitting/parsing logic to the least common denominator -
the `send` function since it can be called either by handling an email event directly or
indirectly via pipeline/stage events

Also, adding the ability to split on semicolons in email addresses

### DIFF
--- a/echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/email/EmailNotificationService.groovy
+++ b/echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/email/EmailNotificationService.groovy
@@ -55,25 +55,11 @@ class EmailNotificationService implements NotificationService {
 
   @Override
   EchoResponse.Void handle(Notification notification) {
-    def expando = { String addresses ->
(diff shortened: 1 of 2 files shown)


echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/email/EmailNotificationService.groovy
```diff
@@ -85,12 +71,25 @@ class EmailNotificationService implements NotificationService {
     MimeMessage message = javaMailSender.createMimeMessage()
     MimeMessageHelper helper = new MimeMessageHelper(message, false, "utf-8")
 
+    def splitAddresses = { String addresses ->
+      // Split on space, comma, or semicolon
+      return addresses.split("[ ,;]").findAll { !it.isEmpty() }
```
ajordens: Ahh semicolons ... what will people think of next :)

PR #555, opened 2019-05-20
fix(logging): add fabulous logging for when pipelines fail to trigger

marchello2000:
The impetus for this change is to be able to better alert on and understand pipeline trigger failures.
We've been bit by trigger failures that went undiscovered (sometimes for days) until our users complained.
With this change we can reliable alert on `orca.trigger.errors` metric which is incremented when kicking
of a pipeline fails (unlike the previous metric `orca.error` which was sometimes triggered erroneously).

When triggering does fail, this change will log, in much more detail, why the trigger failed as well as
log the full payload submitted to `orca` for quick and easy debugging/reproducability of the issue.

Bonus 1: This change (mostly) removes the reliance on Rx and replaces it with an executor to
trigger pipelines.

Bonus 2: Fixup calls to `front50` to allow them to be anonymous


echo-notifications/echo-notifications.gradle
```diff
@@ -27,7 +27,6 @@ dependencies {
     implementation "com.netflix.spinnaker.kork:kork-core"
     implementation "com.netflix.spinnaker.kork:kork-artifacts"
     implementation "com.netflix.spinnaker.kork:kork-web"
-    implementation "io.reactivex:rxjava"
```
dreynaud: 👏

marchello2000:
Honestly, I am still not convinced that the `executorService` is needed:
1. cron triggers already happen on quartz thread
2. events are already Rx'd on IOScheduler [here](https://github.com/spinnaker/echo/blob/master/echo-core/src/main/java/com/netflix/spinnaker/echo/events/EventPropagator.java#L49)

So I don't see the point... but I am going to keep the executor for now (to maintain behavior as before)

echo-pipelinetriggers/src/main/java/com/netflix/spinnaker/echo/pipelinetriggers/orca/PipelineInitiator.java
```diff
@@ -136,62 +151,101 @@ public void startPipeline(Pipeline pipeline) {
...
-  private void onOrcaError(Pipeline pipeline, Throwable error) {
-    registry.counter("orca.errors", "exception", error.getClass().getName()).increment();
-    log.error("Error triggering pipeline: {}", pipeline, error);
-  }
+      try {
+        Thread.sleep(retryDelayMillis);
+      } catch (InterruptedException ignored) {
+      }
 
+      log.warn("Retrying pipeline trigger for {}", pipeline);
```
dreynaud: log.warn("Retrying pipeline trigger (attempt {}/{}) for {}", attempts, retryCount, pipeline);

echo-pipelinetriggers/src/main/java/com/netflix/spinnaker/echo/pipelinetriggers/orca/PipelineInitiator.java
```diff
@@ -136,62 +151,101 @@ public void startPipeline(Pipeline pipeline) {
...
+    int attempts = 0;
 
-  private void onOrcaResponse(TriggerResponse response) {
-    log.info("Triggered pipeline {}", response.getRef());
-  }
+    while (true) {
+      try {
+        attempts++;
+        return orca.trigger(pipeline);
+      } catch (RetrofitError e) {
```
dreynaud: would be nice to log `e`

echo-pipelinetriggers/src/main/java/com/netflix/spinnaker/echo/pipelinetriggers/orca/PipelineInitiator.java
```diff
@@ -93,7 +96,13 @@ public void initialize() {
     }
   }
 
-  public void startPipeline(Pipeline pipeline) {
+  public enum TriggerSource {
+    SCHEDULER,
+    MISSEDSCHEDULER,
+    EVENT
```
dreynaud: might be interesting to have a different value for the different types of events

marchello2000: as in jenkins vs. docker, etc?

echo-pipelinetriggers/src/main/java/com/netflix/spinnaker/echo/pipelinetriggers/orca/PipelineInitiator.java
marchello2000: well, we log the last one (i.e. if all retries fail). otherwise any transitory errors make it noisy... but i could use splainer style and only log it all when the last one fails?

echo-pipelinetriggers/src/main/java/com/netflix/spinnaker/echo/pipelinetriggers/orca/PipelineInitiator.java
dreynaud: yep

echo-pipelinetriggers/src/main/java/com/netflix/spinnaker/echo/pipelinetriggers/orca/PipelineInitiator.java
dreynaud: may be a matter of preference but I'd rather see all exceptions as they happen rather than just the last one. They each are their own event, me thinks.

marchello2000: i will put them in as warnings, then it's clear it's not fatal but we still see the issues. will also add a `orca.trigger.retry` metric

dreynaud: dreynaud likes

echo-pipelinetriggers/src/main/java/com/netflix/spinnaker/echo/pipelinetriggers/orca/PipelineInitiator.java
marchello2000: added to `logOrcaErrorMetric`

dreynaud:
I APPROVE THIS

echo-pipelinetriggers/src/main/java/com/netflix/spinnaker/echo/pipelinetriggers/orca/PipelineInitiator.java
```diff
@@ -115,17 +124,23 @@ public void startPipeline(Pipeline pipeline) {
...
+                  AuthenticatedRequest.allowAnonymous(() -> orca.plan(pipelineToPlan, true));
               pipeline = objectMapper.convertValue(resolvedPipelineMap, Pipeline.class);
             } catch (RetrofitError e) {
-              log.warn(
-                  "Pipeline planning failed: \n{}",
-                  new String(((TypedByteArray) e.getResponse().getBody()).getBytes()));
+              String orcaResponse = "N/A";
+
+              if (e.getResponse() != null && e.getResponse().getBody() != null) {
+                new String(((TypedByteArray) e.getResponse().getBody()).getBytes());
```
emjburns: should this be assigned to something? like `orcaResponse`?

marchello2000: um... yes. thank you! i refactored this and messed up

emjburns:
@ezimanyi you may be interested in this!

PR #560, opened 2019-05-23
fix(auth): propagate MDC across the thread boundary for pipeline execution

marchello2000:
previous change (https://github.com/spinnaker/echo/pull/555) moved orca invocation
to an executor thread without propagating MDC context.

### DIFF
--- a/echo-pipelinetriggers/src/main/java/com/netflix/spinnaker/echo/pipelinetriggers/orca/PipelineInitiator.java
+++ b/echo-pipelinetriggers/src/main/java/com/netflix/spinnaker/echo/pipelinetriggers/orca/PipelineInitiator.java
@@ -33,13 +33,13 @@ import java.util.Collections;
 import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
+import java.util.concurrent.Callable;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.RejectedExecutionException;
 import java.util.stream.Collectors;
 import javax.annotation.PostConstruct;
 import lombok.NonNull;
 import lombok.extern.slf4j.Slf4j;
-import org.slf4j.MDC;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.beans.factory.annotation.Value;
 import org.springframework.http.HttpStatus;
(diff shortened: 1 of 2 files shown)


echo-pipelinetriggers/src/main/java/com/netflix/spinnaker/echo/pipelinetriggers/orca/PipelineInitiator.java
```diff
@@ -166,7 +167,20 @@ public void startPipeline(Pipeline pipeline, TriggerSource triggerSource) {
 
   private void triggerPipeline(Pipeline pipeline, TriggerSource triggerSource)
       throws RejectedExecutionException {
-    executorService.submit(() -> triggerPipelineImpl(pipeline, triggerSource));
+    final Map<String, String> mdc = MDC.getCopyOfContextMap();
```
ajordens: Curious on your thought process behind this vs `propagate()`?

ajordens: Is there any sort of test we could add around this ...

marchello2000: I added a test, but then it actually didn't work. The problem is that the test is using a directExecutor and then it actually calls from the same thread... I will add a test but i will just need more time to figure out how to block on that thread in spock...

`propagate()` doesn't actually carry the MDC across thread boundary, annoyingly - maybe that was the intent but i don't see how it would actually do that short of what I am doing

ajordens: `propagate()` should be able to propagate across threads, it's not going to blindly propagate all MDC but that shouldn't matter?

But you'd need to call `propagate()` outside the `submit()` and pass the resulting `Callable<>` into it.

My go to way of testing something like this would to do an assert within the mock ... roughly something like:

```
1 * orca.blah() { 
  assert MDC.get("blah") == "what I expect"
}
```

marchello2000: Thanks for tip - added a test and used `propagate` - much better this way!

echo-pipelinetriggers/src/main/java/com/netflix/spinnaker/echo/pipelinetriggers/orca/PipelineInitiator.java
```diff
@@ -166,10 +167,28 @@ public void startPipeline(Pipeline pipeline, TriggerSource triggerSource) {
...
-    executorService.submit(() -> triggerPipelineImpl(pipeline, triggerSource));
+    Callable<Void> triggerWithCapturedContext =
+        AuthenticatedRequest.propagate(() -> triggerPipelineImpl(pipeline, triggerSource));
+
+    executorService.submit(
+        () -> {
+          try {
+            triggerWithCapturedContext.call();
+          } catch (Exception e) {
+            // This shouldn't happen since all exceptions are handled in triggerPipelineImpl...
```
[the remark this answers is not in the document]
ajordens: assuming the above ... then can we just pass `triggerWithCapturedContext` into `executorService.submit()`?

echo-pipelinetriggers/src/test/groovy/com/netflix/spinnaker/echo/pipelinetriggers/orca/PipelineInitiatorSpec.groovy
```diff
@@ -94,6 +98,51 @@ class PipelineInitiatorSpec extends Specification {
...
+    def account = "super-duper-account"
+
+    when:
+    MDC.put(AuthenticatedRequest.Header.USER.header, user)
+    MDC.put(AuthenticatedRequest.Header.ACCOUNTS.header, account)
+    pipelineInitiator.startPipeline(pipeline, PipelineInitiator.TriggerSource.SCHEDULER)
+    MDC.remove(AuthenticatedRequest.Header.ACCOUNTS.header)
+    MDC.remove(AuthenticatedRequest.Header.USER.header)
+
+    // Wait for orce to actually be called (happens on separate thread)
```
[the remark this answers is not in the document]
marchello2000: 👍

echo-pipelinetriggers/src/main/java/com/netflix/spinnaker/echo/pipelinetriggers/orca/PipelineInitiator.java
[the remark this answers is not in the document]
marchello2000: yeah, that's much nicer, thanks!

PR #562, opened 2019-05-29
fix(auth): Only trigger pipeline if service account has access to app

jervi:
In the current implementation, Echo will trigger pipelines in Orca even though the `runAsUser` does not have access to the application. This will trigger a round-dance of different errors when Orca tries unsuccessfully to execute the pipeline, so instead, we short circuit the execution in Echo.


jervi:
@spinnaker/reviewers PTAL

PR #580, opened 2019-06-18
fix(build): replace references to removed init-publish.gradle in Dock…

maggieneterval:
…erfile

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,8 +6,7 @@ COPY . workdir/
 
 WORKDIR workdir
 
-RUN GRADLE_USER_HOME=cache ./gradlew -I gradle/init-publish.gradle buildDeb \
-       -I gradle/init-publish.gradle \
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -6,8 +6,8 @@ COPY . workdir/
 
 WORKDIR workdir
 
-RUN GRADLE_USER_HOME=cache ./gradlew -I gradle/init-publish.gradle buildDeb \
-       -I gradle/init-publish.gradle \
+RUN GRADLE_USER_HOME=cache ./gradlew -PenablePublishing=true buildDeb \
+       -PenablePublishing=true \
```
ezimanyi: I'm not sure why `-I gradle/init-publish.gradle` was there twice, but I think it should have only been there once and that it can just be replaced with a single `-PenablePublishing=true`

maggieneterval: Removed the second one!

PR #588, opened 2019-06-28
refactor(slack): Add debug logging of Slack responses

dotdotdotpaul:
Adds debug-level logging for Slack notifications, useful for debugging Slack integration problems (authentication errors, etc).


echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/slack/SlackNotificationService.groovy
```diff
@@ -51,12 +51,15 @@ class SlackNotificationService implements NotificationService {
...
+      def response
       String address = it.startsWith('#') ? it : "#${it}"
       if (sendCompactMessages) {
-        slack.sendCompactMessage(token, new CompactSlackMessage(text), address, true)
+        response = slack.sendCompactMessage(token, new CompactSlackMessage(text), address, true)
       } else {
-        slack.sendMessage(token, new SlackAttachment("Spinnaker Notification", text), address, true)
+        response = slack.sendMessage(token, new SlackAttachment("Spinnaker Notification", text), address, true)
       }
+      log.debug("Received response from Slack: {} {} for message '{}'. {}",
```
marchello2000: Any reason to log this out on every request?
Probably not very useful for successful responses, especially since the description says this is to aid debugging of failures?

If you change it to only log when `response.status != 200` or `response.body ok==false` (see https://api.slack.com/methods/chat.postMessage) you can make it be `log.warn` to be more obvious

echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/slack/SlackNotificationService.groovy
```diff
@@ -51,12 +51,15 @@ class SlackNotificationService implements NotificationService {
   EchoResponse.Void handle(Notification notification) {
     def text = notificationTemplateEngine.build(notification, NotificationTemplateEngine.Type.BODY)
     notification.to.each {
+      def response
```
marchello2000: prefer actual type here
```suggestion
      Response response
```

echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/slack/SlackNotificationService.groovy
marchello2000: (also see https://api.slack.com/incoming-webhooks)

robzienert: I'm okay with logging in the successful case, but only if this is changed to a `trace` level.

dotdotdotpaul: This is logging the *response* from Slack, which might not give you what you expected even if the Slack API returned a 200 or otherwise "okay" response, so I wanted/needed it to tell me what was going on even for successful calls.  Changed to trace -- didn't realize "trace" was a thing, thought "debug" was the most verbose option?

echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/slack/SlackNotificationService.groovy
dotdotdotpaul: Cool 'nuff -- I'm not super familiar with Groovy, I just mimick'd the "def text" variable definition earlier in the function.

marchello2000:
@dotdotdotpaul : something weird is going on with these commits, i think you meant to rebase? Can you clean it up please so the merge is clean, otherwise i think it's good (hard to tell with all the other commit noise)

dotdotdotpaul:
Yeah, sorry, was trying to fix the merging with master and things got...  out of hand.  Now that the checks are passing, I'll rebase and probably just open a new PR.  😛

dotdotdotpaul:
Rebased/Resubmitted as https://github.com/spinnaker/echo/pull/612

PR #589, opened 2019-06-28
fix(notifications):  Need to maintain enum order

dotdotdotpaul:
HIPCHAT was removed from the Enum here, but this enum is duplicated in
other microservices (for example, in Orca,
orca-echo/src/main/groovy/com/netflix/spinnaker/orca/echo/EchoService.groovy
has a copy of the list)

This is a quick fix to repair the numbering scheme.  This should
probably be fully refactored so the enum is either a) specifying actual
values for each type so they maintain consistency going forward, b)
unify in kork somehow, or c) both.


robzienert:
Let's just change it back to `HIPCHAT` and keep the comment something along the lines that it's unused because it's been deprecated. `UNUSED` may be confusing if not looking directly at the enum.

Thanks for fixing this!

PR #590, opened 2019-07-01
fix(webhooks):  Test for empty ref on Git hooks

dotdotdotpaul:
Some Git webhooks don't have a "ref" field, which is assumed in this
handler.  Sets the branch to empty string in this case (presuming other
code is expecting to find a non-null value further downstream).

### DIFF
--- a/echo-webhooks/src/main/groovy/com/netflix/spinnaker/echo/scm/GithubWebhookEventHandler.java
+++ b/echo-webhooks/src/main/groovy/com/netflix/spinnaker/echo/scm/GithubWebhookEventHandler.java
@@ -63,7 +63,11 @@ public class GithubWebhookEventHandler implements GitWebhookHandler {
         objectMapper.convertValue(postedEvent, GithubWebhookEvent.class);
 
     event.content.put("hash", githubWebhookEvent.after);
-    event.content.put("branch", githubWebhookEvent.ref.replace("refs/heads/", ""));
+    if (githubWebhookEvent.ref == null) {
+      event.content.put("branch", "");
+    } else {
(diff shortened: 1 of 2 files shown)


echo-webhooks/src/main/groovy/com/netflix/spinnaker/echo/scm/GithubWebhookEventHandler.java
```diff
@@ -63,7 +63,11 @@ public void handle(Event event, Map postedEvent) {
         objectMapper.convertValue(postedEvent, GithubWebhookEvent.class);
 
     event.content.put("hash", githubWebhookEvent.after);
-    event.content.put("branch", githubWebhookEvent.ref.replace("refs/heads/", ""));
+    if (githubWebhookEvent.ref == null) {
+      event.content.put("branch", "");
+    } else {
+      event.content.put("branch", githubWebhookEvent.ref.replace("refs/heads/", ""));
```
ethanfrogers: @dotdotdotpaul this could probably be simplified by using `Optional.of(githubWebhookEvent.ref).orElse("").replace...`.

ethanfrogers: Could you also ensure this is covered in the test?

dotdotdotpaul: Will add a test.  Not sure the "Optional" thing is that much more "simplified".  IMHO, it kinda clutters the line, which is already pretty long.

PR #592, opened 2019-07-02
fix(cron): fix to not fire the same cron job on multiple instances

marchello2000:
it seems that `read-committed` on mysql doesn't have strong enough locking for
quartz to ensure only one clustered instance picks up a given cron trigger.
[Documentation](http://www.quartz-scheduler.org/documentation/quartz-2.1.7/configuration/ConfigJobStoreTX.html) suggests to set `acquireTriggersWithinLock` to `true`

(also see https://github.com/quartz-scheduler/quartz/issues/107)


ajordens:
```

ajordens:
Ahh just reading that linked `quartz-scheduler` issuel.

marchello2000:
yeah, I am at a bit of a loss myself.. but most signs point to this helping - hard to test locally

PR #597, opened 2019-07-05
fix(swabbie): Update swabbie email tempalte to link ASG resource

aravindmd:
- Update the email template to include link to the resource on spinnaker web ui , `additionalContext` will include `spinnakerWebUrl` which is used to construct the `resourceUrl`. Since we are not using email notifications for any other swabbie resource i just updated the template without any conditional checks.

### DIFF
--- a/echo-notifications/src/main/resources/templates/swabbie/body.ftl
+++ b/echo-notifications/src/main/resources/templates/swabbie/body.ftl
@@ -95,7 +95,7 @@
           </td>
           <td class="title" align="left" valign="middle" style="padding: 16px 0 16px 32px; height: 100%;">
             <h1 style="display: block; font-family: Helvetica, Arial, sans-serif; color: #ffffff; font-size: 28px;">Cleanup Notifications</h1>
-            <h2 style="display: block; font-family: Helvetica, Arial, sans-serif; color: #d8d8d8; font-size: 16px; font-weight: normal;">${notification.additionalContext.resources?size} image(s) not in use scheduled for cleanup</h2>
+            <h2 style="display: block; font-family: Helvetica, Arial, sans-serif; color: #d8d8d8; font-size: 16px; font-weight: normal;">${notification.additionalContext.resources?size} resources(s) not in use scheduled for cleanup</h2>
           </td>
           <td align="center" valign="top" style="padding: 16px 32px 16px 16px; width: 140px;" class="logo">
             <img alt="Spinnaker" src="https://www.spinnaker.io/assets/emails/spinnaker-logo-400.png" width="140" height="140" border="0">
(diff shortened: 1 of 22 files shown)


aravindmd:
Screen Shot 2019-07-08 at 2 45 15 PM

jeyrschabu:
@aravindmd thanks for adding that screenshot. I will defer to @gcomstock for comments on the look and feel. Also might help to see a screenshot with more than one resource.

A couple of things jump out, I notice some labels starting in Cap and others are lowercase. I think the `link` and `id` could be combined to point to the resource in Spinnaker.

echo-notifications/src/main/resources/templates/swabbie/body.ftl
```diff
@@ -138,6 +139,17 @@
               </td>
             </tr>
             </#if>
+            <tr>
+              <td class="loop__key" align="right" valign="top" style="padding: 4px 16px 4px 0; font-family: Helvetica, Arial, sans-serif; font-size: 12px; font-weight: bold;" width="30%">
+                Link
+              </td>
+              <td align="left" style="padding: 4px 0; font-family: Helvetica, Arial, sans-serif; font-size: 12px;">
```
jeyrschabu: Per our chat, lets remove the specific link. In the notification context in swabbie, lets pass in a `resourceType` and a `resourceUrl` (iirc this field should be there already) i.e `resourceUrl: https://spinnaker/search?q=`

https://github.com/spinnaker/swabbie/blob/3440531305926e9cda41cd5b9e93445941e701dc/swabbie-core/src/main/kotlin/com/netflix/spinnaker/config/SwabbieProperties.kt#L92

aravindmd: Agree . I changed the template to use the `spinnakerWebUrl` ( i think we still need this, its different from `resourceUrl` which links to the api route and not the web ui ) and use the `resourceType` we already have as part of the `notificationContext`

aravindmd:
Screen Shot 2019-07-08 at 4 11 24 PM

echo-notifications/src/main/resources/templates/swabbie/body.ftl
jeyrschabu: `spinnakerweburl` is what `resourceUrl` was intended to be. It's not meant to be for api but for visualizing the resource in spinnnaker UI

gcomstock:
+1 to Jeyrs' comment: just make the id clickable and drop the link key/val. If you could expand the button column just a bit so "Opt Out of Delete" worked on one line without wrapping, that'd tidy things up a bit too. Thanks!

aravindmd:
Updated the id to make it clickable and removed the link row , using `spinnakerLink` from notification context and also update the column width to avoid line wrapping.

aravindmd:
Screen Shot 2019-07-09 at 9 38 49 AM

gcomstock:
Thanks captain! Looks great.

aravindmd:
@jeyrschabu let me know if its good to go now

PR #602, opened 2019-07-09
fix(gcb): Allow anonymous calls to igor

ezimanyi:
Since the GCB functionality was written, we've started logging messages every time an intra-microservice call is anonymous. When calling igor in response to a PubSub message, we don't have any authentication context and need to make an anonymous call, so wrap these calls in an allowAnonymous call.


ezimanyi:
This has no functional effect, it just removes some log messages.

PR #606, opened 2019-07-17
fix(pubsub): Increase socket timeout

aravindmd:
- We noticed timeouts while running this locally on VPN , bumping the value to `500` ms to help with local runs


echo-pubsub-core/src/main/java/com/netflix/spinnaker/echo/pubsub/utils/NodeIdentity.java
```diff
@@ -73,7 +73,7 @@ private static String resolveHostname(String validationHost, int validationPort)
         try {
           socket = new Socket();
           socket.bind(new InetSocketAddress(address, 0));
-          socket.connect(new InetSocketAddress(validationHost, validationPort), 125);
+          socket.connect(new InetSocketAddress(validationHost, validationPort), 500);
```
marchello2000: this looks good, just wonder if it would be useful to have this configurable.. but doesn't look like this is a bean... so i guess never mind

PR #486, opened 2019-03-11
refactor(bitbucket): refactor bitbucket webhook

ethanfrogers:
refactors bitbucket webhook handling to eliminate all of the casting
type casting that i was doing before.

PR #609, opened 2019-07-24
fix(echo): Trigger properly off Github Pull Requests

justinrlee:
If you send Github webhooks for PRs, they don't trigger anything.

Draft cause I haven't built tests yet.


echo-webhooks/src/main/groovy/com/netflix/spinnaker/echo/scm/GithubWebhookEventHandler.java
```diff
@@ -63,12 +63,15 @@ public void handle(Event event, Map postedEvent) {
...
+    // Default to .ref, then to pull_request.head.ref, then to "" (empty string)
+    if (githubWebhookEvent.ref == null && githubWebhookEvent.pull_request.head.ref == null) {
       event.content.put("branch", "");
+    } else if (githubWebhookEvent.ref == null) {
+      event.content.put("branch", githubWebhookEvent.pull_request.head.ref);
     } else {
       event.content.put("branch", githubWebhookEvent.ref.replace("refs/heads/", ""));
     }
-    event.content.put("repoProject", githubWebhookEvent.repository.owner.name);
+    event.content.put("repoProject", githubWebhookEvent.repository.owner.login);
```
staciegravy: How does this change affect things down the line?

justinrlee:
This could probably be refactored to use `Optional` as part of https://github.com/spinnaker/echo/pull/486, but for now it should work.

ethanfrogers:
Personally, I'd like to see this cleaned up to mimic #486. It will make it a lot easier to encapsulate the logic for each case if each item is in it's own method. Also, it will read _so much_ cleaner.

justinrlee:
@ethanfrogers tada?

echo-webhooks/src/main/groovy/com/netflix/spinnaker/echo/scm/GithubWebhookEventHandler.java
```diff
@@ -59,34 +66,38 @@ public void handle(Event event, Map postedEvent) {
...
-        objectMapper.convertValue(postedEvent, GithubWebhookEvent.class);
+    GithubWebhookEvent webhookEvent = null;
+    String eventType = "";
 
-    event.content.put("hash", githubWebhookEvent.after);
-    if (githubWebhookEvent.ref == null) {
-      event.content.put("branch", "");
-    } else {
-      event.content.put("branch", githubWebhookEvent.ref.replace("refs/heads/", ""));
+    // TODO: Detect based on header rather than body key
```
ethanfrogers: You should actually be able to `TODO` this as part of this PR. `event.details.requestHeaders` should have a copy of all the headers. You can probably do something like this:
```
String eventType = event.details.requestHeaders.getOrDefault("X-Github-Event", "")
```

echo-webhooks/src/test/groovy/com/netflix/spinnaker/echo/controllers/WebhooksControllerSpec.groovy
```diff
@@ -431,7 +476,7 @@ class WebhooksControllerSpec extends Specification {
       event = it[0]
     }
 
-    event.content.hash == "0000000000000000000000000000000000000000"
+    event.content.hash == null
```
ethanfrogers: This should be set back to the original value.

justinrlee:
Adding flavor from conversation with Ethan:
Gate doesn't currently pass in the `x-github-event` header, so for now we will detect event type by looking at JSON structure / fields.

PR #614, opened 2019-08-02
fix(sql): Make database-mysql.sql idempotent

jmrodriguez-clv:
This makes the script idempotent, ideally to be run upon echo instance start to populate schema without having to check if the schema and users are already created.


marchello2000:
thanks, @jmrodriguez-clv are you intending to make this run as part of `echo` start? I think that might be a bit tricky given how we use/instantiate `liquibase` in `kork`. But this is definitely a nice addition even if run manually

robzienert:
Merging, but like @marchello2000 mentioned, this schema is not applied when Echo starts - although we should move towards doing that.

PR #622, opened 2019-08-14
fix(notifications/googlechat): Fix field visibility.

duftler:
Fixes https://github.com/spinnaker/spinnaker/issues/4754.


duftler:
The issue was that Jackson 2 won't implicitly recognize fields with package visibility.

duftler:
@spinnakerbot cherry-pick 1.15

PR #633, opened 2019-08-27
feat(slack): Make slack base url configurable

diasjorge:
Allow slack notifications to be sent to compatible implementations.

spinnaker/spinnaker#4634

### DIFF
--- a/echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/config/SlackConfig.groovy
+++ b/echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/config/SlackConfig.groovy
@@ -55,12 +55,12 @@ class SlackConfig {
     String endpoint;
 
     if (StringUtils.isNotBlank(slackBaseUrl)) {
-      log.info("Using Slack base url: {}.", slackBaseUrl)
       endpoint = slackBaseUrl
     } else {
-      log.info("Using Slack incoming webhooks: {}.", useIncomingWebHook)
       endpoint = useIncomingWebHook ? SLACK_INCOMING_WEBHOOK : SLACK_CHAT_API;
     }
+
+    log.info("Using Slack {}: {}.", useIncomingWebHook ? "incoming webhook" : "chat api", endpoint)
 
     newFixedEndpoint(endpoint)
   }
--- a/echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/config/SlackConfig.groovy
+++ b/echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/config/SlackConfig.groovy
@@ -54,7 +54,7 @@ class SlackConfig {
   Endpoint slackEndpoint(@Qualifier("useIncomingWebHook") boolean useIncomingWebHook) {
     String endpoint;
 
-    if (slackBaseUrl != null && slackBaseUrl != "") {
+    if (StringUtils.isNotBlank(slackBaseUrl)) {
       log.info("Using Slack base url: {}.", slackBaseUrl)
       endpoint = slackBaseUrl
     } else {
--- a/echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/config/SlackConfig.groovy
+++ b/echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/config/SlackConfig.groovy
@@ -47,8 +47,8 @@ class SlackConfig {
   @Value('${slack.base-url:}')
   String slackBaseUrl
 
-  @Value('${slack.compatibleAPI:false}')
-  Boolean slackCompatibleAPI
+  @Value('${slack.force-use-incoming-webhook:false}')
+  Boolean forceUseIncomingWebhook;
 
   @Bean
   Endpoint slackEndpoint(@Qualifier("useIncomingWebHook") boolean useIncomingWebHook) {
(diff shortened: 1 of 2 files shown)


diasjorge:
@spinnaker/reviewers

diasjorge:
@emjburns, I changed my test to define explicitly the behaviour when slack base url is an empty string and null.

I had not considered an empty string but I think we should not set the base url to it, so I made a change to reflect that.

emjburns:
Thanks! LGTM

emjburns:
One more question - does the slack compatible api also need a chat and webhook url?

diasjorge:
@emjburns our internal implementation as well as other popular ones I've seen like mattermost, do need a token.

Now, this token will not necessarily have the format "XXX/XXX/XXX". Checking a bit more, in this case the SlackService will not call the API using the incomingWebhookPath which is what we need.

How could we solve this?
- We could assume if you're setting the baseURL that then you'll use the webhook API.
- We could have another config option, maybe named "incomingWebhook", default to null. If it is not set we try to infer it from the token as we currently do, if it is set then we just use that value.

What do you think? 

Thanks

emjburns:
Does it make sense to have a config property that indicates if it's a slack-compatible api, or slack itself? In that case, the default would be that you're using slack and we should use the existing logic. If it's slack-compatable, then you also need to provide the base url and token.

diasjorge:
Hi @emjburns I like your idea, I'll make some changes in that direction.

diasjorge:
@emjburns Sorry for the delay, I've implemented the changes I think it's the right direction. What do you think?

echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/config/SlackConfig.groovy
```diff
@@ -44,10 +44,24 @@ class SlackConfig {
...
+  @Value('${slack.compatibleAPI:false}')
+  Boolean slackCompatibleAPI
+
   @Bean
   Endpoint slackEndpoint(@Qualifier("useIncomingWebHook") boolean useIncomingWebHook) {
-    log.info("Using Slack incoming webhooks: {}.", useIncomingWebHook)
-    String endpoint = useIncomingWebHook ? SLACK_INCOMING_WEBHOOK : SLACK_CHAT_API;
+    String endpoint;
+
+    if (slackBaseUrl != null && slackBaseUrl != "") {
```
marchello2000: since `org.apache.commons.lang3.StringUtils` is already imported, you can:
```suggestion
    if (StringUtils.isNotBlank(slackBaseUrl)) {
```

echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/config/SlackConfig.groovy
```diff
@@ -44,10 +44,24 @@ class SlackConfig {
   final static String SLACK_INCOMING_WEBHOOK = 'https://hooks.slack.com'
   final static String SLACK_CHAT_API = 'https://slack.com'
 
+  @Value('${slack.base-url:}')
+  String slackBaseUrl
+
+  @Value('${slack.compatibleAPI:false}')
+  Boolean slackCompatibleAPI
```
marchello2000: I am struggling to understand what is the meaning for `slackCompatibleAPI`?
I think it would be more appropriately named `forceUseIncomingWebhook`.
Also, probably worth having some exception to make sure end users understand that they either specify `slackBaseUrl` or `slackCompatibleAPI/forceUseIncomingWebhook` not both

marchello2000: but if it makes sense to keep it as is, out convention is:
```
@Value('${slack.compatible-api:false}')
```

diasjorge: Can you recommend a way to do that type of config validations? I searched the codebase but couldn't find any example.

diasjorge: slackCompatibleAPI comes from the previous comment from @emjburns. I'll change it to forceUseIncomingWebhook as it makes more sense in the context.
In our use case, we'd need to actually specify both, since we need the webhook behaviour (Which defaults to false) and the base url.

marchello2000: I think Spring provides some validation mechanism, but it might be easier to just `throw` inside the `slackEndoint`/`slackService` bean functions

diasjorge: I've changed the name of the config value and also extracted the token detection to its own method, and I think it's more clear now. I wasn't really convinced with the previous name. I think it's now clear that either you force the setting or it will be detected from the token.
The change of the base url then becomes something independent of the incoming webhook setting.

echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/config/SlackConfig.groovy
```diff
@@ -44,10 +44,24 @@ class SlackConfig {
...
   Endpoint slackEndpoint(@Qualifier("useIncomingWebHook") boolean useIncomingWebHook) {
-    log.info("Using Slack incoming webhooks: {}.", useIncomingWebHook)
-    String endpoint = useIncomingWebHook ? SLACK_INCOMING_WEBHOOK : SLACK_CHAT_API;
+    String endpoint;
+
+    if (StringUtils.isNotBlank(slackBaseUrl)) {
+      log.info("Using Slack base url: {}.", slackBaseUrl)
+      endpoint = slackBaseUrl
+    } else {
+      log.info("Using Slack incoming webhooks: {}.", useIncomingWebHook)
```
marchello2000: do you think there is value to logging the actual url we will be pointing towards?.. e.g. swap the two log lines
and then 
```
log.info("Using Slack {}: {}.", useIncomingWebHook ? "incoming webhook" : "chat api", slackBaseUrl)
```

diasjorge: I've implemented the logging change, we always log the url and if it's api or webhooks. Thanks for the review

PR #635, opened 2019-08-29
feat(telemetry): adds listener for telemetry metrics

staciegravy:
WIP: Pending PR for application hashing, excutionId hashing and default endpoint.

Adds telemetry event listener and corresponding protobuf definitions. 

[spinnaker/spinnaker#4738](https://github.com/spinnaker/spinnaker/issues/4738)


echo-telemetry/src/main/java/com/netflix/spinnaker/echo/config/TelemetryConfig.java
```diff
@@ -0,0 +1,65 @@
...
+import org.springframework.context.annotation.Configuration;
+import retrofit.Endpoint;
+import retrofit.RestAdapter;
+import retrofit.client.Client;
+import retrofit.converter.JacksonConverter;
+
+@Slf4j
+@Configuration
+@ConditionalOnProperty("telemetry.enabled")
+@CompileStatic
```
ttomsu: nit: I don't think we need compile static on a java file

echo-telemetry/src/main/java/com/netflix/spinnaker/echo/config/TelemetryConfig.java
```diff
@@ -0,0 +1,65 @@
...
+  public TelemetryService telemetryService(
+      Endpoint telemetryEndpoint, Client retrofitClient, RestAdapter.LogLevel retrofitLogLevel) {
+    log.info("Telemetry service loaded");
+
+    TelemetryService client =
+        new RestAdapter.Builder()
+            .setEndpoint(telemetryEndpoint)
+            .setConverter(new JacksonConverter())
+            .setClient(retrofitClient)
+            .setLogLevel(RestAdapter.LogLevel.FULL)
```
ttomsu: Maybe have an additional value for `telemetry.logLevel` ? I imagine this can and will get noisy quickly

staciegravy:
Closing in lieu of #640

PR #638, opened 2019-09-03
fix(error handling): mark executions failed

marchello2000:
We need a way to indicate failures in materializing a pipeline.
If `echo` fails to materialize a pipeline during a trigger the failure needs to be recorded
in `orca` otherwise, the failure appears "silent" to the user because the UI is unable to
show any errors unless they are associated with an execution.
This is especially bad if, say, there is a typo in the properties file in the jenkins trigger
Starting the pipeline appears to do nothing and the user is confused.

This change captures errors during artifact resolution and sends them to `orca` to capture for the user
>NOTE: depends on related `orca` PR: https://github.com/spinnaker/orca/pull/3126

### DIFF
--- a/echo-pipelinetriggers/src/main/java/com/netflix/spinnaker/echo/pipelinetriggers/monitor/TriggerMonitor.java
+++ b/echo-pipelinetriggers/src/main/java/com/netflix/spinnaker/echo/pipelinetriggers/monitor/TriggerMonitor.java
@@ -16,6 +16,7 @@
 
 package com.netflix.spinnaker.echo.pipelinetriggers.monitor;
 
+import com.google.common.base.Strings;
 import com.netflix.spectator.api.Id;
 import com.netflix.spectator.api.Registry;
 import com.netflix.spinnaker.echo.events.EchoEventListener;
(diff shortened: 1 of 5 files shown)


echo-pipelinetriggers/src/main/java/com/netflix/spinnaker/echo/pipelinetriggers/monitor/TriggerMonitor.java
```diff
@@ -74,12 +74,17 @@ private void triggerMatchingPipelines(T event) {
     try {
       List<Pipeline> matchingPipelines = eventHandler.getMatchingPipelines(event, pipelineCache);
       matchingPipelines.stream()
+          .filter(p -> p.getErrorMessage() == null)
```
ezimanyi: Any thoughts on whether we should also allow the empty string through this filter?  (And exclude it from the filter below?)

marchello2000: was contemplating that myself and don't recall why i picked one over the other :)
but it's prob safer to `isNullOrEmpty()`it

PR #640, opened 2019-09-03
Stats collection implementation

ttomsu:
This PR expands on https://github.com/spinnaker/echo/pull/635 to remove the duplicate proto file and some excess GRPC stuff that we don't need.

I'm willing to submit the above pull request, but I think this one includes those changes as well (under @Staciegravy 's GH handle)

cc @imosquera


echo-telemetry/echo-telemetry.gradle
```diff
@@ -0,0 +1,25 @@
...
+ */
+
+dependencies {
+  implementation project(':echo-model')
+  implementation project(':echo-notifications')
+  implementation project(':echo-proto')
+  implementation 'com.squareup.retrofit:retrofit'
+  implementation 'com.squareup.retrofit:converter-jackson'
+  implementation 'com.netflix.spinnaker.kork:kork-web'
+  implementation 'com.google.protobuf:protobuf-java-util:3.7.1'
```
robzienert: Would like to see this dep get extracted out to `kork/spinnaker-dependencies`

echo-telemetry/src/main/java/com/netflix/spinnaker/echo/telemetry/TelemetryService.java
```diff
@@ -0,0 +1,26 @@
...
+ */
+
+package com.netflix.spinnaker.echo.telemetry;
+
+import retrofit.client.Response;
+import retrofit.http.Body;
+import retrofit.http.POST;
+
+public interface TelemetryService {
+  @POST("/}")
```
robzienert: This looks like a typo. Is this a typo?

echo-telemetry/src/main/java/com/netflix/spinnaker/echo/telemetry/TelemetryEventListener.java
```diff
@@ -0,0 +1,102 @@
...
+              .setSpinnakerInstance(spinnakerInstance);
+
+      telemetryService.sendMessage(JsonFormat.printer().print(eventProto));
+
+    } catch (Exception e) {
+      log.error("Could not send Telemetry event {}", event, e);
+    }
+  }
+
+  private Execution.Builder getExecutionBuilder(Map execution) {
```
plumpy: nit: call this `createExecutionBuilder`

echo-telemetry/src/main/java/com/netflix/spinnaker/echo/telemetry/TelemetryEventListener.java
```diff
@@ -0,0 +1,102 @@
...
+
+      Map execution = (Map) event.content.get("execution");
+      Execution.Builder executionBuilder = getExecutionBuilder(execution);
+
+      List<Map> stages = (ArrayList<Map>) execution.get("stages");
+      for (Map stage : stages) {
+        executionBuilder.addStages(getStageBuilder(stage));
+      }
+
+      Application.Builder applicationBuilder =
```
plumpy: Generally speaking, people usually use an immutable object (by calling `build()`) everywhere _unless_ there's some reason they can't. That makes it easier for the reader by letting them know that we aren't going to further change this object later in the method.

The only object right now that needs to remain a `Builder` is the `executionBuilder`, but I'd be inclined to get rid of that too by doing this:
```
List<Stage> protoStages = stages.stream.map(this::createStage).collect(toImmutableList());
Execution execution = Execution.newBuilder()
    .setId(...)
    .addAllStages(protoStages)
    // ...
    .build();
```
```

File: echo-telemetry/src/test/groovy/com/netflix/spinnaker/echo/TelemetryEventListenerSpec.groovy [hunk 5]
```diff
@@ -0,0 +1,54 @@
...
+                                ]
+                        ]
+                ]
+        )
+
+        when:
+        listener.processEvent(event)
+
+        then:
+        1 * service.sendMessage('{\n  "spinnakerInstance": {\n    "id": "test-instance",\n    "application": {\n      "id": "some-application",\n      "execution": {\n        "id": "execution_id",\n        "type": "PIPELINE",\n        "trigger": {\n          "type": "GIT"\n        },\n        "stages": [{\n          "type": "deploy",\n          "status": "SUCCEEDED"\n        }, {\n          "type": "wait",\n          "status": "SUCCEEDED"\n        }],\n        "status": "SUCCEEDED"\n      }\n    }\n  }\n}')
```
plumpy: this test is way too fragile and impossible to understand.

Take the argument, parse it into a protobuf `Event` object, then make assertions about that object

echo-telemetry/src/main/java/com/netflix/spinnaker/echo/telemetry/TelemetryEventListener.java
```diff
@@ -0,0 +1,102 @@
...
+          Application.newBuilder()
+              .setId(event.details.getApplication())
```
plumpy: aren't we supposed to be salting and hashing the application name?

echo-telemetry/src/main/java/com/netflix/spinnaker/echo/telemetry/TelemetryEventListener.java
```diff
@@ -0,0 +1,102 @@
...
+              .setExecution(executionBuilder);
+
+      SpinnakerInstance.Builder spinnakerInstance =
+          SpinnakerInstance.newBuilder().setId(instanceId).setApplication(applicationBuilder);
+
+      com.netflix.spinnaker.proto.stats.Event.Builder eventProto =
+          com.netflix.spinnaker.proto.stats.Event.newBuilder()
+              .setSpinnakerInstance(spinnakerInstance);
+
+      telemetryService.sendMessage(JsonFormat.printer().print(eventProto));
```
ezimanyi: What's the failure mode if the back-end of the telemetry service is down or becomes slow?

If it's down and errors quickly, I suppose the worst that will happen is that we'll catch and log a lot of exceptions (which could still lead to a lot of log noise).

But what about the case where the service is slow?  (Or more likely, connections time out because people are running in environments that don't have access to the public internet.)

I don't see any timeout or circuit-breaking here.  Will this just eat up all of echo's connection pool in the event that connections time out?

echo-telemetry/src/main/java/com/netflix/spinnaker/echo/telemetry/TelemetryEventListener.java
```diff
@@ -0,0 +1,102 @@
...
+  }
+
+  @Value("${telemetry.instanceId}")
+  String instanceId;
+
+  @Override
+  public void processEvent(Event event) {
+    try {
+      if (event.getDetails() == null
+          || !event.getDetails().getType().equals("orca:pipeline:complete")) {
```
ezimanyi: Do failed pipelines also send `orca:pipeline:complete`?  I think they might send `orca:pipeline:failed`? If that's the case, none of the status information will be useful as we'll only be recording successful executions.

echo-telemetry/src/main/java/com/netflix/spinnaker/echo/config/TelemetryConfig.java
```diff
@@ -0,0 +1,66 @@
...
+import retrofit.client.Client;
+import retrofit.converter.JacksonConverter;
+
+@Slf4j
+@Configuration
+@ConditionalOnProperty("telemetry.enabled")
+class TelemetryConfig {
+
+  @Value("${telemetry.endpoint}")
+  String endpoint;
```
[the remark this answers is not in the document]
robzienert: Would also pile on and say that `@ConfigurationProperties` classes are preferrable to `@Value`.

echo-proto/echo-proto.gradle
```diff
@@ -0,0 +1,30 @@
+plugins {
+  id 'de.undercouch.download' version '4.0.0'
+}
+
+apply plugin: 'java'
+apply plugin: 'com.google.protobuf'
+
+task downloadProtos(type: Download) {
+  src 'https://raw.githubusercontent.com/spinnaker/stats/master/proto/stats/log.proto'
```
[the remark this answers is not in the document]
cfieber: Oh sure, treat the underlying problem and not just the symptom.  This is how you avoid a buildopiod epidemic I suppose!

(Blame me for this, I answered the question of how to do without of asking why)

echo-telemetry/src/main/java/com/netflix/spinnaker/echo/telemetry/TelemetryEventListener.java
duftler: When we are deriving audit log entries from these events, we look for `orca:pipeline:failed`: https://github.com/GoogleCloudPlatform/spinnaker-for-gcp/blob/master/scripts/install/spinnakerAuditLog/index_js.template#L74

ttomsu:
OK, so I'll take a slightly different approach here:

1.) Create a `kork-proto` module that will contain and export the pre-compiled (and versioned) Java library for the stats proto. This module should contain all the proto-generation related code, so that it's not being spread around the codebase.

2.) Update the github.com/spinnaker/stats repo to import the new location of the stats proto. The Go code would still need to be generated from the spinnaker/stats repo AFAICT right now (still learning about how Golang is used out in the wild)

I'll address the other comments once this is in place and/or someone suggests a better approach.

plumpy:
> 1.) Create a `kork-proto` module that will contain and export the pre-compiled (and versioned) Java library for the stats proto.

I like the approach @robzienert took in spinnaker/front50#571 where the repository that provides the service also owns the proto file. That feels like it makes more sense because it means you can merge changes to the proto in the same commit as the implementation supporting those changes.

I suppose the difference here is that the `stats` repo isn't another microservice but it's own weird thing. But IMO, it still feels better to do it that way...

ttomsu:
Chatted with @plumpy in person - agreed that since even though the `stats` repo is the service that primarily consumes the proto, there is no mechanism to publish pre-compiled JARs from that repo, and adding Gradle there just to do JAR publishing seems like overkill. Happy to entertain other approaches - otherwise PTAL at https://github.com/spinnaker/kork/pull/374

echo-telemetry/src/main/java/com/netflix/spinnaker/echo/telemetry/TelemetryService.java
ttomsu: Fixed.

echo-telemetry/echo-telemetry.gradle
ttomsu: The protobuf bom is in spinnaker-deps, and version has been removed here.

echo-proto/echo-proto.gradle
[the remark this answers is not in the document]
ttomsu: Moved the proto to `kork-proto` where the generated library is now being published.

echo-telemetry/echo-telemetry.gradle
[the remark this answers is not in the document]
ttomsu: Done.

echo-telemetry/src/main/java/com/netflix/spinnaker/echo/config/TelemetryConfig.java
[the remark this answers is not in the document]
ttomsu: Changed to an `@ConfigurationProperties`

echo-telemetry/src/main/java/com/netflix/spinnaker/echo/telemetry/TelemetryEventListener.java
```diff
@@ -0,0 +1,102 @@
...
+public class TelemetryEventListener implements EchoEventListener {
+  private final TelemetryService telemetryService;
+
+  @Autowired
+  public TelemetryEventListener(TelemetryService telemetryService) {
+    this.telemetryService = telemetryService;
+  }
+
+  @Value("${telemetry.instanceId}")
+  String instanceId;
```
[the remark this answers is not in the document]
ttomsu: Changed to config props injected via constructor.

echo-telemetry/src/main/java/com/netflix/spinnaker/echo/telemetry/TelemetryEventListener.java
```diff
@@ -0,0 +1,102 @@
...
+        .setId(execution.get("id").toString())
+        .setType(Execution.Type.valueOf(execution.get("type").toString().toUpperCase()))
+        .setTrigger(
+            Execution.Trigger.newBuilder()
+                .setType(
+                    Execution.Trigger.Type.valueOf(trigger.get("type").toString().toUpperCase())))
+        .setStatus(Status.valueOf(execution.get("status").toString().toUpperCase()));
+  }
+
+  private Stage.Builder getStageBuilder(Map stage) {
```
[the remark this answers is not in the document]
ttomsu: Refactored the proto formation a bit. PTAL.

echo-telemetry/src/main/java/com/netflix/spinnaker/echo/telemetry/TelemetryEventListener.java
ttomsu: Done.

echo-web/echo-web.gradle
```diff
@@ -35,14 +35,16 @@ dependencies {
...
+    implementation project(':echo-telemetry')
     implementation "com.netflix.spinnaker.fiat:fiat-api:$fiatVersion"
     implementation "com.netflix.spinnaker.fiat:fiat-core:$fiatVersion"
     implementation "org.springframework.boot:spring-boot-starter-web"
     implementation "org.springframework.boot:spring-boot-starter-actuator"
     implementation "com.netflix.spinnaker.kork:kork-core"
     implementation "com.netflix.spinnaker.kork:kork-config"
     implementation "com.netflix.spinnaker.kork:kork-secrets-aws"
-    implementation "com.netflix.spinnaker.kork:kork-stackdriver"
+//    implementation "com.netflix.spinnaker.kork:kork-stackdriver"
```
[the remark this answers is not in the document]
ttomsu: There are some conflicting gRPC issues stemming from `kork-stackdriver` that are incompatible with the pubsub library. I removed this and it resolved the issue I was seeing, but I'll leave this open and revisit once this PR is submittable

echo-telemetry/src/main/java/com/netflix/spinnaker/echo/telemetry/TelemetryEventListener.java
ttomsu: Outdated.

echo-telemetry/src/main/java/com/netflix/spinnaker/echo/telemetry/TelemetryEventListener.java
ttomsu: Good catch. Added the hashing, but haven't added a salt yet - hoping to use the application creation time, but I think I'll need to add that to the execution model

@louisjimenez take note if we're going to need to add the MPT versioning we can add them at the same time.

echo-telemetry/src/main/java/com/netflix/spinnaker/echo/telemetry/TelemetryEventListener.java
ttomsu: Added some connection and read timeout settings, though I think it would be nice to "autodetect" an airgapped env (say with an HTTP HEAD call to the endpoint?) and automatically disable itself if outbound traffic is disabled - but maybe I'm over thinking it and it's not that big of a deal and admins who are annoyed by it can opt out.

echo-telemetry/src/main/java/com/netflix/spinnaker/echo/telemetry/TelemetryEventListener.java
ttomsu: Good catch. Turns out orchestrations (ad hoc tasks) are also different - added the missing 3

ttomsu:
Things left to do for this PR:

[] Figure out (or revert) kork-stackdriver thing
[] Fix/add tests

Post submit TODOs:
* Add identifier for MPT v1/v2 to execution
* Add salt for application ID hashing in the form of application timestamp (without an RPC call to front 50)
* Add filter for uninteresting pipeline runs (like `savePipeline`)?
* Auto detect air-gapped env?

echo-telemetry/src/test/groovy/com/netflix/spinnaker/echo/TelemetryEventListenerSpec.groovy
ttomsu: Done - good advice.

echo-web/echo-web.gradle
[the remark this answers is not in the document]
ttomsu: Reverted.

ttomsu:
Tests have been fixed and the kork-stackdriver change reverted. PTAL

echo-telemetry/src/test/groovy/com/netflix/spinnaker/echo/telemetry/TelemetryEventListenerSpec.groovy
```diff
@@ -0,0 +1,112 @@
...
+              type  : "wait",
+              status: "TERMINAL"
+            ],
+          ]
+        ]
+      ]
+    )
+
+    when:
+    listener.processEvent(event)
```
plumpy: I'd run this through with an empty details and content, too... that will catch the NPEs hopefully

echo-telemetry/src/main/java/com/netflix/spinnaker/echo/telemetry/TelemetryEventListener.java
```diff
@@ -0,0 +1,162 @@
...
+      SpinnakerInstance spinnakerInstance =
+          SpinnakerInstance.newBuilder()
+              .setId(telemetryConfigProps.getInstanceId())
+              .setVersion(telemetryConfigProps.getSpinnakerVersion())
+              .setApplication(application)
+              .build();
+
+      com.netflix.spinnaker.kork.proto.stats.Event loggedEvent =
+          com.netflix.spinnaker.kork.proto.stats.Event.newBuilder()
+              .setSpinnakerInstance(spinnakerInstance)
```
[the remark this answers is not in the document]
ttomsu: An execution should always be in the context of an application - pipelines and ad hoc tasks run _in_ applications. Basically it mirrors the proto mirrors the hierarchy of the system - a spinnaker instance --> an application under management --> a deployment (or other pipeline run) of that application

echo-telemetry/src/main/java/com/netflix/spinnaker/echo/telemetry/TelemetryEventListener.java
```diff
@@ -0,0 +1,162 @@
...
+        return;
+      }
+
+      String eventType = event.getDetails().getType();
+      if (!LOGGABLE_DETAIL_TYPES.contains(eventType)) {
+        log.debug("Telemetry not sent: type '{}' not whitelisted ", eventType);
+        return;
+      }
+
+      Map<String, Object> execution = (Map<String, Object>) event.content.get("execution");
```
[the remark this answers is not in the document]
ttomsu: With public fields does it matter?

echo-telemetry/src/test/groovy/com/netflix/spinnaker/echo/telemetry/TelemetryEventListenerSpec.groovy
ttomsu: Added event validation test

echo-telemetry/src/main/java/com/netflix/spinnaker/echo/telemetry/TelemetryEventListener.java
```diff
@@ -0,0 +1,162 @@
...
+        log.debug("Missing execution from Event content.");
+        return;
+      }
+
+      String hashedApplicationId = hash(event.details.getApplication());
+      String hashedExecutionId = hash(execution.getOrDefault("id", "").toString());
+      Execution.Type executionType =
+          Execution.Type.valueOf(
+              // TODO(ttomsu, louisjimenez): Add MPTv1 and v2 execution type detection.
+              execution.getOrDefault("type", "").toString().toUpperCase());
```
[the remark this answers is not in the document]
ttomsu: defaulted `getorDefault()` to `UNKNOWN`

echo-telemetry/src/main/java/com/netflix/spinnaker/echo/telemetry/TelemetryEventListener.java
```diff
@@ -0,0 +1,162 @@
...
+        return;
+      }
+
+      Map<String, Object> execution = (Map<String, Object>) event.content.get("execution");
+      if (execution == null) {
+        log.debug("Missing execution from Event content.");
+        return;
+      }
+
+      String hashedApplicationId = hash(event.details.getApplication());
```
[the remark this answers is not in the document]
ttomsu: Added validation

echo-telemetry/src/main/java/com/netflix/spinnaker/echo/telemetry/TelemetryEventListener.java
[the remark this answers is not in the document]
plumpy: But that's backwards, right? If you're modeling a Java object of an event, you would have an application field on the event. I can tell that's how you'd model it because that's exactly how Event is already modeled. I can't see why the proto should be different?

echo-telemetry/src/main/java/com/netflix/spinnaker/echo/telemetry/TelemetryEventListener.java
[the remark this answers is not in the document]
plumpy: Why are they public, though? But okay, I suppose that can be cleaned up later, it just doesn't seem like we should be further encouraging use that anti-pattern.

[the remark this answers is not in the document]
ttomsu: I changed it to use the getter because ultimately it doesn't matter which is used.

We can have a chat on it's anti-pattern-ness over beverages.

echo-telemetry/src/main/java/com/netflix/spinnaker/echo/telemetry/TelemetryEventListener.java
```diff
@@ -0,0 +1,168 @@
...
+
+  private Stage toStage(Map stage) {
+    return Stage.newBuilder()
+        .setType(stage.get("type").toString())
+        .setStatus(Status.valueOf(stage.getOrDefault("status", "UNKNOWN").toString().toUpperCase()))
+        .build();
+  }
+
+  private String hash(String clearText) {
+    return Hashing.sha256().hashString(clearText, StandardCharsets.UTF_8).toString();
```
ezimanyi: There was a discussion on the RFC about wanting to include some salt in this hash to make it hard to find trivially guessable application names in the data by pre-computing the hash.  I don't see that included here---was that suggestion ultimately rejected?

echo-telemetry/src/main/java/com/netflix/spinnaker/echo/config/TelemetryConfig.java
```diff
@@ -0,0 +1,80 @@
...
+            .setLog(new Slf4jRetrofitLogger(TelemetryService.class))
+            .build()
+            .create(TelemetryService.class);
+
+    return client;
+  }
+
+  private OkClient telemetryOkClient(TelemetryConfigProps configProps) {
+    OkHttpClient httpClient = new OkHttpClient();
+    httpClient.setConnectTimeout(configProps.connectionTimeoutMillis, TimeUnit.MILLISECONDS);
```
ezimanyi: I think the timeouts here will help, and may be sufficient to have reasonable resilience against the endpoint going down.  (If we're only trying to log completed pipelines, those may be infrequent enough that the timeout is enough to save us from building up a huge backlog of outgoing connections.)  But I'd still suggest using a fault tolerance library, such as `Hystrix` (which we use extensively in gate) to circuit-break if this starts failing a lot---at the very least before we increase the scope of events that are logged.

echo-telemetry/src/main/java/com/netflix/spinnaker/echo/telemetry/TelemetryEventListener.java
ttomsu: I still intend to do that - however, the salt I would like to use (the application creation timestamp) is not available without an RPC call right now. I'd like to move it into the execution so that I could get the value without an additional RPC.

The logic behind this is that I think nearly every operation does a fetch from Front50 at some point or another, and that this data is static over the life of the application, and not easily guessable by someone on the outside. Happy to entertain other suggestions, though.

ezimanyi: It looks like there were some suggestions on the design doc around things to use.  I do think it's worth updating the design document to point out that the initial implementation won't salt these values given that it is something that a number of people commented on.

ttomsu: This PR won't salt them, but that doesn't mean we're going to flip the switch without it.

echo-telemetry/src/main/java/com/netflix/spinnaker/echo/config/TelemetryConfig.java
ttomsu: Ack - https://github.com/spinnaker/spinnaker/issues/4864

echo-telemetry/src/main/java/com/netflix/spinnaker/echo/telemetry/TelemetryEventListener.java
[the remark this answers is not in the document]
ttomsu: Flattened the proto - here's a sample textproto:

```
spinnaker_instance: <
  id: "146a74d1-19d4-4e33-8bbc-a209b3090d74"
  version: "branch:upstream/master"
>
application: <
  id: "5be1ecc7935f1dd85635d4feedaf660594030253cc97c9e9ca3819ffeac36b65"
>
execution: <
  id: "6fc5eec30babcba3def193ba92e5c6e83b49942c72dd142032dfc33e516757f3"
  type: ORCHESTRATION
  trigger: <
    type: MANUAL
  >
  stages: <
    type: "scaleManifest"
    status: SUCCEEDED
  >
  status: SUCCEEDED
>
```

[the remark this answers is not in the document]
ttomsu: https://github.com/spinnaker/kork/pull/381 for the proto change

echo-telemetry/src/main/java/com/netflix/spinnaker/echo/telemetry/TelemetryEventListener.java
```diff
@@ -0,0 +1,168 @@
...
+      }
+
+      Map<String, Object> execution = (Map<String, Object>) event.getContent().get("execution");
+      if (execution == null || execution.isEmpty()) {
+        log.debug("Missing execution from Event content.");
+        return;
+      }
+
+      String hashedApplicationId = hash(applicationId);
+      String hashedExecutionId = hash(execution.getOrDefault("id", "").toString());
```
plumpy: For a missing execution ID, we're going to the `id` field to sha256("") which seems less useful than just not setting it at all, doesn't it?

echo-telemetry/src/main/java/com/netflix/spinnaker/echo/telemetry/TelemetryEventListener.java
```diff
@@ -0,0 +1,168 @@
...
+      telemetryService.log(new TypedJsonString(content));
+      log.debug("Telemetry sent!");
+    } catch (Exception e) {
+      log.warn("Could not send Telemetry event {}", event, e);
+    }
+  }
+
+  private Stage toStage(Map stage) {
+    return Stage.newBuilder()
+        .setType(stage.get("type").toString())
```
plumpy: use getOrDefault to avoid an NPE

ttomsu: Done.

echo-telemetry/src/main/java/com/netflix/spinnaker/echo/telemetry/TelemetryEventListener.java
ttomsu: Done.

halconfig/echo.yml
```diff
@@ -21,8 +21,6 @@ orca:
   enabled: ${services.orca.enabled:false}
   baseUrl: ${services.orca.baseUrl:http://localhost:8083}
 
-endpoints.health.sensitive: false
```
ezimanyi: What was this removed?  It seems unrelated to the current change.

echo-telemetry/src/main/java/com/netflix/spinnaker/echo/telemetry/TelemetryEventListener.java
```diff
@@ -0,0 +1,170 @@
...
+
+      String hashedApplicationId = hash(applicationId);
+      Execution.Type executionType =
+          Execution.Type.valueOf(
+              // TODO(ttomsu, louisjimenez): Add MPTv1 and v2 execution type detection.
+              execution.getOrDefault("type", "UNKNOWN").toString().toUpperCase());
+      Status executionStatus =
+          Status.valueOf(execution.getOrDefault("status", "UNKNOWN").toString().toUpperCase());
+
+      Map trigger = (Map) execution.getOrDefault("trigger", new HashMap());
```
ezimanyi: Maybe make this a `Map<String, Object>` to provide some type safety given that we're assuming that the keys are all strings.  (We're actually also assuming all values are strings as well, I believe, so you could probably also do `Map<String, String>`.)

halconfig/echo.yml
ttomsu: I was having issues starting Echo, and the logs spit out this error message for me:

```
The use of configuration keys that are no longer supported was found in the environment:

Property source 'applicationConfig: [file:/usr/local/google/home/ttomsu/.spinnaker/echo.yml]':
        Key: endpoints.health.sensitive
                Reason: Endpoint sensitive flag is no longer customizable as Spring Boot no longer provides a customizable security auto-configuration .
```
Turns out my startup error was something else, but removing this line made this error go away.

echo-telemetry/src/main/java/com/netflix/spinnaker/echo/telemetry/TelemetryEventListener.java
ttomsu: We're assuming all keys are strings, but not all of the values are strings.

Added (String,Object) for type safety

PR #644, opened 2019-09-06
fix(notification/slack): fix missing logger

cfieber:
was causing slack notifications to repeat endlessly as the notification would be
sent, but the log statement afterwords would fail and return an error back to
orca which would retry.


danopia:
Spinnaker 1.16 shipped with "Add trace logging of Slack responses" but not this fix, will we see a backport?

cfieber:

PR #654, opened 2019-09-16
feat(stats): Adds circuit breaker logic via resilience4j

ttomsu:
This should prevent sending pings to the stats collector in case the environment is air gapped or the endpoint is actually down. 

Side note: resilience4j is pretty neat, and 1.0.0 was just released. Opened https://github.com/spinnaker/kork/pull/382 to add `resilience4j-circuitbreaker` to the next kork release.

Fixes https://github.com/spinnaker/spinnaker/issues/4864

### DIFF
--- a/echo-telemetry/echo-telemetry.gradle
+++ b/echo-telemetry/echo-telemetry.gradle
@@ -16,8 +16,6 @@
 
 dependencies {
   implementation project(':echo-model')
-  implementation project(':echo-notifications')
-  implementation 'com.google.guava:guava'
   implementation 'com.google.protobuf:protobuf-java-util'
   implementation "com.netflix.spinnaker.kork:kork-proto"
   implementation 'com.netflix.spinnaker.kork:kork-web'
--- a/echo-telemetry/src/main/java/com/netflix/spinnaker/echo/telemetry/TelemetryEventListener.java
+++ b/echo-telemetry/src/main/java/com/netflix/spinnaker/echo/telemetry/TelemetryEventListener.java
@@ -28,6 +28,8 @@ import com.netflix.spinnaker.kork.proto.stats.Execution;
 import com.netflix.spinnaker.kork.proto.stats.SpinnakerInstance;
 import com.netflix.spinnaker.kork.proto.stats.Stage;
 import com.netflix.spinnaker.kork.proto.stats.Status;
+import io.github.resilience4j.circuitbreaker.CallNotPermittedException;
+import io.github.resilience4j.circuitbreaker.CircuitBreakerRegistry;
 import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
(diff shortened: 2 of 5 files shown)


echo-telemetry/src/main/java/com/netflix/spinnaker/echo/config/TelemetryConfig.java
```diff
@@ -64,6 +68,22 @@ private OkClient telemetryOkClient(TelemetryConfigProps configProps) {
...
+        CircuitBreakerConfig.custom()
+            .waitDurationInOpenState(Duration.ofHours(12)) // Arbitrary.
+            .ringBufferSizeInClosedState(10)
+            .build();
+
+    return circuitBreakerRegistry.circuitBreaker("stats", config);
+  }
+
+  @Bean
+  public CircuitBreakerRegistry circuitBreakerRegistry() {
```
cfieber: would it be worth adding the resilience4j spring boot starter instead: https://resilience4j.readme.io/docs/getting-started-3

would allow the config options to get pulled from yaml

ttomsu: Oy - gotta upgrade to 1.0.0 to fix an issue with pulling in the whole spring boot starter: https://github.com/spinnaker/kork/pull/384

ttomsu: Done.

echo-telemetry/echo-telemetry.gradle
```diff
@@ -23,4 +23,5 @@ dependencies {
   implementation 'com.netflix.spinnaker.kork:kork-web'
   implementation 'com.squareup.retrofit:retrofit'
   implementation 'com.squareup.retrofit:converter-jackson'
+  implementation 'io.github.resilience4j:resilience4j-spring-boot2'
```
robzienert: This shouldn't be necessary. I added it as an `api` scope to `kork-core`.

https://github.com/spinnaker/kork/blob/master/kork-core/kork-core.gradle#L18

ttomsu: I think @ezimanyi was pushing for each module being more explicit about it's direct dependencies, no?

ezimanyi: I think if we add it explicitly as `api` scope in `kork-core` this is fine; I was more concerned with depending on things that were in `implementation` scope in transitive dependencies (that would nonetheless end up on the compile classpath in gradle 4).

ttomsu: Removed because of the transitive dep of echo-telemetry --> echo-model --> kork-core --> resilience4j-spring-boot2

PR #657, opened 2019-09-23
fix(build): whitelist build package to avoid being ignored by gcloud cloudbuild

Jammy-Louie:
fix to build images using `cloud builds`

when I attempt to build using gcloud `gcloud builds submit ....` the  archive file uploaded does not include the `build` package. the line in the `.gitignore` `*/build/"` is being used to ignore the package. 

Work around is to white list the package directory.

filed a bug with gcloud build:
https://issuetracker.google.com/issues/141435398

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 /.gradle/
 /build
 */build/
+!*/src/**/build/
 .DS_Store
 *.iml
 *.ipr


ezimanyi:
@Jammy-Louie : I just tested and the current `.gitignore` is working for me; in #473 I updated it so that it only ignores `build` folders at the top level or one level deep.  The uploaded source properly contains the source code in the `build` directory even before the change here...is it possible that you have something else overriding the `.gitignore` here?

Jammy-Louie:
@ezimanyi I run the command `gcloud builds submit` in a docker container

with the following tool versions
Google Cloud SDK 263.0.0
alpha 2019.09.13
beta 2019.09.13
bq 2.0.47
core 2019.09.13
gsutil 4.42
kubectl 2019.09.13

it has always continued to fail until I've either commented out `*/build` or If I've white listed the directory.

I have the docker image and docker file, which i can make available.

Jammy-Louie:
@ezimanyi I got response from my bug

```
Conclusions:

1. "*/build/", "build" and "build/" ignore "build" directories which are both in the project root directory and in "/src".
2. /build" and "/build/" ignore only "/build".
3. This behaviour is because lines in conclusion "1." don't specify an absolute path in the project so it takes any "build" directory it finds (I also tried with a directory "/src/main/build" and it was ignored too).
4. The solution is to use absolute paths (relative to the project's root) in ".gitignore".
```

ezimanyi:
Thanks for the details!  I somehow had the top-level `gcloud` config setting `gcloudignore=false` so it wasn't ignoring any files, which is why this was working for me.  I think this change makes sense, thanks!

Jammy-Louie:

PR #659, opened 2019-09-24
fix(auth): Fetch updated pipeline when auth is enabled

costimuraru:
Fixes https://github.com/spinnaker/spinnaker/issues/4941

depends on https://github.com/spinnaker/front50/pull/599


costimuraru:
ping @ttomsu @ajordens @ezimanyi @jtk54 
Please help with a review to unblock this.

echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/notification/DryRunNotificationAgent.java
```diff
@@ -59,7 +60,7 @@ public void sendNotifications(
     }
     log.info("Received dry run notification for {}", pipelineConfigId);
     Optional<Pipeline> match =
-        front50.getPipelines(application).stream()
+        AuthenticatedRequest.allowAnonymous(() -> front50.getPipelines(application)).stream()
```
marchello2000: I'm not sure I follow how does this end up setting `restricted = false` in front50?

ttomsu:
Closing this because the bug was closed.

PR #660, opened 2019-09-24
fix(stats): Parse Event objects into structured POJOs for Telemetry events

ttomsu:
Fixes https://github.com/spinnaker/spinnaker/issues/4885

I tried (a lot) to make this broader - to change [Event.content](https://github.com/spinnaker/echo/blob/master/echo-model/src/main/java/com/netflix/spinnaker/echo/model/Event.java#L28) from a `Map` to a proper POJO `Content` object - but I couldn't get Jackson deserialization to work as expected and _not_ have to change large amounts of both prod and test code (access like `event.content.something.somethingElse` is pervasive across this codebase).

So instead I opted to fix it locally - grabbing the Map and converting it locally to known fields and setting reasonable defaults for those fields, should they be absent.

One thing to note also is the wrapper `TEL` object - because there are a lot of duplicate names here, I wrapped the local objects in a wrapper so I wouldn't have to use long, fully qualified names everywhere.

### DIFF
--- a/echo-telemetry/src/main/java/com/netflix/spinnaker/echo/config/TelemetryConfig.java
+++ b/echo-telemetry/src/main/java/com/netflix/spinnaker/echo/config/TelemetryConfig.java
@@ -20,6 +20,7 @@ import com.netflix.spinnaker.echo.telemetry.TelemetryService;
 import com.netflix.spinnaker.retrofit.RetrofitConfigurationProperties;
 import com.netflix.spinnaker.retrofit.Slf4jRetrofitLogger;
 import com.squareup.okhttp.OkHttpClient;
+import de.huxhorn.sulky.ulid.ULID;
 import java.util.concurrent.TimeUnit;
 import lombok.Data;
 import lombok.extern.slf4j.Slf4j;
@@ -72,7 +73,7 @@ public class TelemetryConfig {
 
     boolean enabled = false;
     String endpoint = DEFAULT_TELEMETRY_ENDPOINT;
-    String instanceId;
+    String instanceId = new ULID().nextULID();
     String spinnakerVersion = "unknown";
     int connectionTimeoutMillis = 3000;
     int readTimeoutMillis = 5000;
--- a/echo-telemetry/src/main/java/com/netflix/spinnaker/echo/telemetry/TelemetryEventListener.java
+++ b/echo-telemetry/src/main/java/com/netflix/spinnaker/echo/telemetry/TelemetryEventListener.java
@@ -16,8 +16,14 @@
 
 package com.netflix.spinnaker.echo.telemetry;
 
+import com.fasterxml.jackson.databind.DeserializationFeature;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
+import com.fasterxml.jackson.databind.annotation.JsonPOJOBuilder;
 import com.google.common.collect.ImmutableSet;
 import com.google.common.hash.Hashing;
+import com.google.protobuf.Descriptors.EnumDescriptor;
+import com.google.protobuf.Descriptors.EnumValueDescriptor;
 import com.google.protobuf.util.JsonFormat;
 import com.netflix.spinnaker.echo.config.TelemetryConfig;
 import com.netflix.spinnaker.echo.events.EchoEventListener;
@@ -32,13 +38,14 @@ import io.github.resilience4j.circuitbreaker.CallNotPermittedException;
 import io.github.resilience4j.circuitbreaker.CircuitBreakerRegistry;
 import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
-import java.util.HashMap;
 import java.util.List;
-import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
 import java.util.stream.Collectors;
+import lombok.Builder;
+import lombok.Getter;
 import lombok.extern.slf4j.Slf4j;
+import org.apache.commons.lang3.StringUtils;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
 import org.springframework.stereotype.Component;
@@ -59,7 +66,10 @@ public class TelemetryEventListener implements EchoEventListener {
           "orca:pipeline:failed");
 
   private static final JsonFormat.Printer JSON_PRINTER =
-      JsonFormat.printer().omittingInsignificantWhitespace();
+      JsonFormat.printer().includingDefaultValueFields();
+
+  private static final ObjectMapper objectMapper =
+      new ObjectMapper().configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false);
 
   private final TelemetryService telemetryService;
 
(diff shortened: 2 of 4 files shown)


echo-telemetry/src/main/java/com/netflix/spinnaker/echo/telemetry/TelemetryEventListener.java
```diff
@@ -217,4 +216,61 @@ public String toString() {
       return new String(getBytes(), StandardCharsets.UTF_8);
     }
   }
+
+  // Arbitrary container for TelemetryEventListener structured data, so there aren't huge
+  // fully-qualified names for the equivalent classes in the kork.proto package all over the code.
+  public static class TEL {
```
ezimanyi: Given that this is only used 3 times above, I'd suggest giving it a more descriptive name; reading `TEL` above seems more distracting than if we just gave it a longer name.  (Also, I think that this inner class can be `private` which would be particularly important if we're giving it a non-conventional name.)

echo-telemetry/src/main/java/com/netflix/spinnaker/echo/telemetry/TelemetryEventListener.java
```diff
@@ -202,6 +193,14 @@ private String hash(String clearText) {
     return Hashing.sha256().hashString(clearText, StandardCharsets.UTF_8).toString();
   }
 
+  private static EnumValueDescriptor parseEnum(EnumDescriptor ed, String value) {
```
ezimanyi: The fact that we have to call this a few times (and also get a generic `EnumValueDescriptor` that loses type information) makes me wonder if there's a more standard way of doing this with protos?  Maybe there's a way to parse and default without needing to manually do this?  @plumpy may have thoughts here.

echo-telemetry/src/main/java/com/netflix/spinnaker/echo/telemetry/TelemetryEventListener.java
```diff
@@ -217,4 +216,61 @@ public String toString() {
...
+
+    @Data
+    public static class Stage {
+      private String status = "UNKNOWN";
+      private String type = "UNKNOWN";
+      private String syntheticStageOwner;
+      private Context context = new Context();
+
+      public boolean isSyntheticStage() {
+        return syntheticStageOwner != null && !"".equalsIgnoreCase(syntheticStageOwner);
```
ezimanyi: Really minor, but `equalsIgnoreCase` is unecessary here as we're only ever comparing to the empty string.

echo-telemetry/src/main/java/com/netflix/spinnaker/echo/telemetry/TelemetryEventListener.java
```diff
@@ -217,4 +216,61 @@ public String toString() {
       return new String(getBytes(), StandardCharsets.UTF_8);
     }
   }
+
+  // Arbitrary container for TelemetryEventListener structured data, so there aren't huge
+  // fully-qualified names for the equivalent classes in the kork.proto package all over the code.
+  public static class TEL {
+
+    @Data
```
ezimanyi: It's a bit more annoying to write these in a way that can be easily deserialized, but I've been trying write as many of my data classes as possible to be immutable.  In this case I don't think we're ever mutating instances of these classes after creation.

echo-telemetry/src/main/java/com/netflix/spinnaker/echo/telemetry/TelemetryEventListener.java
```diff
@@ -80,85 +86,79 @@ public TelemetryEventListener(
   @SuppressWarnings("unchecked")
```
ezimanyi: This can be removed now! :tada:

echo-telemetry/src/main/java/com/netflix/spinnaker/echo/config/TelemetryConfig.java
```diff
@@ -72,7 +73,7 @@ private OkClient telemetryOkClient(TelemetryConfigProps configProps) {
 
     boolean enabled = false;
     String endpoint = DEFAULT_TELEMETRY_ENDPOINT;
-    String instanceId;
+    String instanceId = new ULID().nextULID();
```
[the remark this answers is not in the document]
ttomsu: That's correct. As this is disabled by default, adding an identifier here will do nothing.

echo-telemetry/src/main/java/com/netflix/spinnaker/echo/telemetry/TelemetryEventListener.java
```diff
@@ -58,15 +62,17 @@
...
-      JsonFormat.printer().omittingInsignificantWhitespace();
+  private static final JsonFormat.Printer JSON_PRINTER = JsonFormat.printer();
 
   private final TelemetryService telemetryService;
 
   private final TelemetryConfig.TelemetryConfigProps telemetryConfigProps;
 
   private final CircuitBreakerRegistry registry;
 
+  private final ObjectMapper objectMapper =
```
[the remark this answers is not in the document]
ttomsu: Made it static

echo-telemetry/src/main/java/com/netflix/spinnaker/echo/telemetry/TelemetryEventListener.java
ttomsu: Renamed to `Holder`. Discussed in person that this needs to be `public` for deserialization.

echo-telemetry/src/main/java/com/netflix/spinnaker/echo/telemetry/TelemetryEventListener.java
ttomsu: I don't think the loss of type information matters as it's passed to the enum's `valueOf()` method that accepts a generic `EnumValueDescriptor`.

Unfortunately there aren't many examples or SO posts out there of how to better parse proto Enums, and I've never worked with Java protos internally.

echo-telemetry/src/main/java/com/netflix/spinnaker/echo/telemetry/TelemetryEventListener.java
ttomsu: Dropped in favor of StringUtils.

echo-telemetry/src/main/java/com/netflix/spinnaker/echo/telemetry/TelemetryEventListener.java
ttomsu: Copied the pattern you pointed me to from Orca - I think this ended up with more code rather than the previous null-checking. I think it would have more impact if I were able to actually put this in the `model` package and use it all over the code, rather than just in this local file.

echo-telemetry/src/main/java/com/netflix/spinnaker/echo/telemetry/TelemetryEventListener.java
ttomsu: Yay! Done.

echo-telemetry/src/main/java/com/netflix/spinnaker/echo/telemetry/TelemetryEventListener.java
```diff
@@ -217,4 +223,71 @@ public String toString() {
...
+
+  // Arbitrary container for TelemetryEventListener structured data, so there aren't huge
+  // fully-qualified names for the equivalent classes in the kork.proto package all over the code.
+  public static class Holder {
+
+    @Getter
+    @Builder
+    @JsonDeserialize(builder = Content.ContentBuilder.class)
+    public static class Content {
+      @Builder.Default private Execution execution = Execution.builder().build();
```
ezimanyi: Now that these classes are immutable :tada: :tada: , all the member variables can/should be made `final`.

PR #663, opened 2019-09-25
fix(notifications): Don't send failure notifications for cancelled pipelines and stages

luispollo:
This fixes an issue where Echo would send failure notifications for cancelled stages, and tries to account for a similar scenario with cancelled pipelines where the execution object has two different fields for cancellation status and only one was being checked.

Also added tests to cover some of the basic flows in `AbstractEventNotificationAgent`, which is where this logic resides.

### DIFF
--- a/echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/notification/AbstractEventNotificationAgent.groovy
+++ b/echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/notification/AbstractEventNotificationAgent.groovy
@@ -78,7 +78,13 @@ abstract class AbstractEventNotificationAgent implements EchoEventListener {
       return
     }
 
-    if (config.type == 'pipeline' && event.content.execution?.status == 'CANCELED') {
+    if (config.type == 'stage' && event.content.canceled == true) {
+      return
+    }
+
+    // TODO (lpollo): why do we have a 'CANCELED' status and a canceled property, which are prime for inconsistency?
+    if (config.type == 'pipeline' &&
+      (event.content.execution?.status == 'CANCELED' || event.content.execution?.canceled == true)) {
       return
     }
 
--- a/echo-notifications/src/test/groovy/com/netflix/spinnaker/echo/notification/AbstractEventNotificationAgentSpec.groovy
+++ b/echo-notifications/src/test/groovy/com/netflix/spinnaker/echo/notification/AbstractEventNotificationAgentSpec.groovy
@@ -73,10 +73,16 @@ class AbstractEventNotificationAgentSpec extends Specification {
 
     // notifications OFF, stage complete
     fakeStageEvent("orca:stage:complete", null)                                               || 0
+    // notifications OFF, stage failed
+    fakeStageEvent("orca:stage:complete", null)                                               || 0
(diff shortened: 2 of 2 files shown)


luispollo:
@emjburns @marchello2000 Please review.

echo-notifications/src/test/groovy/com/netflix/spinnaker/echo/notification/AbstractEventNotificationAgentSpec.groovy
```diff
@@ -0,0 +1,139 @@
...
+      content: [execution: [
+        id: "1",
+        name: "foo-pipeline",
+        status: 'WHATEVER',
+        canceled: true,
+        notifications: [[type: "fake", when: "pipeline.failed"]]
+      ]])                                                                                   || 0
+    // notifications ON, another check for cancelled pipeline (should skip notifications)
+    // note: this case is a bit convoluted as the event type is still set to "failed" by orca for cancelled pipelines
+    new Event(details: [type: "orca:pipeline:failed"],
```
emjburns: I think this case is the same as the one above?

echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/notification/AbstractEventNotificationAgent.groovy
```diff
@@ -78,7 +78,13 @@ abstract class AbstractEventNotificationAgent implements EchoEventListener {
       return
     }
 
-    if (config.type == 'pipeline' && event.content.execution?.status == 'CANCELED') {
+    if (config.type == 'stage' && event.content.canceled == true) {
```
marchello2000: so one repro/failure scenario is, have a "failure" notification on a stage, cancel a pipeline while that stage is running?

I don't understand the second repro case (below)?

marchello2000: oh, i see.... when we cancel the execution, we don't refetch the execution from the db hence `canceled = true` but `status` is probably still `= running`..? but I still don't understand why this doesn't happen always..?

luispollo: Yes, that’s what I’m thinking, but I couldn’t quite pin it down, so this was more of a hunch...

echo-notifications/src/test/groovy/com/netflix/spinnaker/echo/notification/AbstractEventNotificationAgentSpec.groovy
luispollo: Indeed. Copy/paste mistake. Will fix.

luispollo: Indeed. Copy/paste mistake. Will fix.

luispollo:
@emjburns and @marchello2000 Would you please look this over one more time and give it a 👍 if you think it looks good? Thanks!

luispollo:
@gal-yardeni FYI

PR #665, opened 2019-09-30
fix(emailTemplate): Updates to swabbie notification template

aravindmd:
Updates to email template based on some feedback

- added region and account information 
- links to serverGroup now takes the user to the actual serverGroup instead of the search results
view
- option to link the slack channel in the email 
- cleanup to the email body

### DIFF
--- a/echo-notifications/src/main/resources/templates/swabbie/body.ftl
+++ b/echo-notifications/src/main/resources/templates/swabbie/body.ftl
@@ -112,9 +112,10 @@
     <!-- CONTENT -->
       <table bgcolor="#ffffff" border="0" cellpadding="0" cellspacing="0" width="100%" style="max-width: 1000px;">
         <!-- To map the resourceType to web ui we need an additional 's' -->
-      <#assign rawResourceType = notification.additionalContext.configuration.resourceType+"s">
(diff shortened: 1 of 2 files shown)


aravindmd:
Screen Shot 2019-09-30 at 9 11 06 AM

emjburns:
I feel like each server group takes up a large amount of space. Is there any way we could display them as something like `server group: account/region/name`? Or do you think that would be confusing?

aravindmd:
> I feel like each server group takes up a large amount of space. Is there any way we could display them as something like `server group: account/region/name`? Or do you think that would be confusing?

That's true , @gcomstock can you provide some inputs here ?

gcomstock:
Do we expect this list to be super long? If not, I don't think this current layout is an issue.

emjburns:
I'm not sure how long we expect it to be. We can figure something out later if it ends up being a lot :)

gcomstock:
Sounds good! Thank you

aravindmd:
Sounds good , i'll merge this after i have the swabbie change merged.

jeyrschabu:
Closing this guy as these changes were ported into https://github.com/spinnaker/echo/pull/671.

PR #668, opened 2019-10-02
feat(tasks): allow notifications for tasks

emjburns:
Right now we can't send notifications for tasks. Why? Not sure. This code is 2015 vintage.  

This PR adds support for sending notifications for tasks:

```json
"trigger": {
		"notifications": [
			{
				"address": "#myslackchannel",
				"level": "pipeline",
				"type": "slack",
				"when": [
					"orchestration.starting",
					"orchestration.complete",
					"orchestration.failed"
				]
			}
		]
	},
```

It also moves some error handling from subclasses to the parent class because of 💁style.

### DIFF
--- a/echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/notification/BearychatNotificationAgent.groovy
+++ b/echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/notification/BearychatNotificationAgent.groovy
@@ -49,52 +49,47 @@ class BearychatNotificationAgent extends AbstractEventNotificationAgent {
 
   @Override
   void sendNotifications(Map preference, String application, Event event, Map config, String status) {
-    try {
-      String buildInfo = ''
+    String buildInfo = ''
 
-      if (config.type == 'pipeline' || config.type == 'stage') {
-        if (event.content?.execution?.trigger?.buildInfo?.url) {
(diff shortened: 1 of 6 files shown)


echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/notification/BearychatNotificationAgent.groovy
```diff
@@ -49,52 +49,47 @@ class BearychatNotificationAgent extends AbstractEventNotificationAgent {
...
-      log.info('Sending bearychat message {} for {} {} {} {}', kv('address', preference.address), kv('application', application), kv('type', config.type), kv('status', status), kv('executionId', event.content?.execution?.id))
-
+    String buildInfo = ''
 
-      String message = ''
-
-      if (config.type == 'stage') {
-        String stageName = event.content.name ?: event.content?.context?.stageDetails?.name
-        message = """Stage $stageName for """
+    if (config.type == 'pipeline' || config.type == 'stage') {
```
robzienert: Pulling on a thread here... I think we may want to `isExecution(config.type)` here as well. I bet the other notification agents also have similar conditions.

emjburns: how about I save those improvements for going through and cleaning up all the notification logic?

PR #673, opened 2019-10-16
chore(github): test out mergify config

marchello2000:


plumpy:
Correct me if I'm wrong, but since anyone with reviewer status can apply a label, this now means every reviewer now also has merge permission, right?

If so, this needs to be reverted until we figure something else out...

PR #675, opened 2019-10-16
chore(github): revert mergify config

plumpy:
Reverts spinnaker/echo#673

I'm fairly certain this turns every reviewer into an approver. (Reviewers have the `triage` permission, which allows them to [add labels](https://help.github.com/en/articles/repository-permission-levels-for-an-organization).)

Merging this without a review, since you did too :wink:

### DIFF
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -1,9 +1,0 @@
-pull_request_rules:
-  - name: automatic merge on CI success and review
-    conditions:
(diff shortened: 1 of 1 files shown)


marchello2000:
@plumpy thanks for reverting this and sorry about this - my understanding from @robzienert was that only approvers can add labels. I will follow up with him on that

PR #678, opened 2019-10-18
chore(github): Add mergify config

marchello2000:
Automatically merge PRs when
a) ready label is present and
b) ci passed
c) the PR has at least 1 approval by someone who is a member of oss-approvers group

Also, automatically merge kork autoBumps on CI success

### DIFF
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -1,0 +1,19 @@
+pull_request_rules:
+  - name: Automatically merge on CI success and review
+    conditions:
+      - status-success=continuous-integration/travis-ci/pr
+      - "label=ready to merge"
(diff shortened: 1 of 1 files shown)


.mergify.yml
```diff
@@ -0,0 +1,19 @@
+pull_request_rules:
+  - name: Automatically merge on CI success and review
+    conditions:
+      - status-success=continuous-integration/travis-ci/pr
+      - "label=ready"
```
plumpy: could we use @rz's `ready to merge` instead of just `ready`?

marchello2000: yes, i actually went to every repo and updated the label to `ready to merge` but forgot to update this, thank you for catching it!

PR #691, opened 2019-10-29
feat(triggers) Adds support for payloadConstraints on property files for build triggers

linjmeyer:
Implements [Spinnaker Issue #5053](https://github.com/spinnaker/spinnaker/issues/5053)

Use case is conditionally triggering pipelines after a successful Jenkins build.  For example, Jenkins may be setup for builds on all branches but Spinnaker pipelines should only be master.  More complex use cases are possible too, like triggering 'stage' pipelines on feature/* branches or when a Jenkins build has something like `deploy_target=stage`.

* Adds support for payloadConstrains: {} dictionary for Build triggers (e.g. Jenkins).  
* If payloadConstraints are present, then properties in the properties file must pass the constraints to trigger a build
* Regexp is supported (works the same as webhook and pub/sub payloadConstraints)
* Does not support Manual Pipeline triggers (ie: if build 30 fails constraints it can still be manually triggered after the fact)
* Only tested with Jenkins

I think it will work with other build systems as long as they support properties, but unfortunately I don't have any available to test.

### DIFF
--- a/echo-pipelinetriggers/src/main/java/com/netflix/spinnaker/echo/pipelinetriggers/eventhandlers/BuildEventHandler.java
+++ b/echo-pipelinetriggers/src/main/java/com/netflix/spinnaker/echo/pipelinetriggers/eventhandlers/BuildEventHandler.java
@@ -144,7 +144,7 @@ public class BuildEventHandler extends BaseTriggerEventHandler<BuildEvent> {
     return Collections.emptyList();
   }
 
-  static User getKorkUser(Trigger trigger) {
+  private User getKorkUser(Trigger trigger) {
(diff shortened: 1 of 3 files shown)


linjmeyer:
I should probably add: I'm not a Java dev and this is my first time contributing to Spinnaker outside the spin cli.  Feedback is very welcome

linjmeyer:
@cfieber done!  Let me know what you think, and if you want me to fix the incorrect commit messages.

cfieber:
Thanks for this!

emjburns:
@linjmeyer Thanks for this, this is awesome!

PR #703, opened 2019-11-20
feat(extension): Add event parser extension for echo-rest

jonsie:
Alpha PF4J extension for parsing Echo events prior to POST to configured service.  This implementation is quite simple at the moment.

Current implementation will override any additional configuration (like `wrap`).  However, as this progresses I would like to see this extension point work with existing configurations and allow multiple parsers for multiple configured URLs.

If things progress well, the more powerful extension point would be `EchoEventListener` which would allow us to decompose echo event listeners into discrete plugins.

### DIFF
--- a/echo-api/echo-api.gradle
+++ b/echo-api/echo-api.gradle
@@ -1,0 +1,6 @@
+dependencies {
+  api "com.netflix.spinnaker.kork:kork-plugins-api"
+
+  compileOnly("org.projectlombok:lombok")
+  annotationProcessor("org.projectlombok:lombok")
+}
--- a/echo-api/src/main/java/com/netflix/spinnaker/echo/extension/rest/RestEventParser.java
+++ b/echo-api/src/main/java/com/netflix/spinnaker/echo/extension/rest/RestEventParser.java
@@ -1,0 +1,61 @@
+/*
+ * Copyright 2019 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.spinnaker.echo.extension.rest;
+
+import com.netflix.spinnaker.kork.annotations.Alpha;
+import java.util.List;
+import java.util.Map;
+import lombok.AllArgsConstructor;
+import lombok.Data;
+import lombok.NoArgsConstructor;
+import org.pf4j.ExtensionPoint;
+
+@Alpha
+public interface RestEventParser extends ExtensionPoint {
+
+  /**
+   * Parse an Event prior to POST to the configured URL
+   *
+   * @param event {@link Event}
+   * @return Map, which conforms to the RestEventListener API
+   */
+  Map parse(Event event);
+
+  @Data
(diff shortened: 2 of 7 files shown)


echo-api/echo-api.gradle
```diff
@@ -0,0 +1,7 @@
+dependencies {
+  api "com.netflix.spinnaker.kork:kork-plugins-api"
+  implementation project(':echo-model')
```
cfieber: I think it would be worth moving Event out of echo-model and into echo-api to support this, or defining an equivalent Event class in API and have the extension processor handle mapping the API event to the model event.

As it stands the plugin implementor is going to end up with a dependency on echo-model as well which I think we would want to try to avoid.

jonsie: Yeah, I wasn't sure what to do here.  I was going with the idea that, at the end of the day, the plugin implementor can just use whatever dependencies they want but I do agree - moving things into `-api` makes more sense.

echo-api/src/main/java/com/netflix/spinnaker/echo/extension/rest/RestEventParser.java
```diff
@@ -0,0 +1,60 @@
...
+   *
+   * @param event {@link Event}
+   * @return Map, which conforms to the RestEventListener API
+   */
+  Map parse(Event event);
+
+  @Data
+  class Event {
+    public Metadata details;
+    public Map content;
```
cfieber: would maybe be good to get some generic types on these even if its just Map<String, Object> for the json object payload bits

echo-api/src/main/java/com/netflix/spinnaker/echo/extension/rest/RestEventParser.java
```diff
@@ -0,0 +1,60 @@
...
+  @NoArgsConstructor
+  class Metadata {
+    private String source;
+    private String type;
+    private String created;
+    private String organization;
+    private String project;
+    private String application;
+    private String _content_id;
+    private Map requestHeaders;
```
cfieber: Think this would be Map<String, List<String>> to match HttpHeaders on echo-model

echo-api/src/main/java/com/netflix/spinnaker/echo/extension/rest/RestEventParser.java
```diff
@@ -0,0 +1,60 @@
...
+    private Map attributes;
```
cfieber: This is probably Map<String, String>... there are two places that use this off the original Event object and one is untyped and the other is <String, String>. I'd lean towards narrowing it to <String, String> for the API..

echo-rest/src/main/groovy/com/netflix/spinnaker/echo/events/RestEventListener.groovy
```diff
@@ -70,7 +77,10 @@ class RestEventListener implements EchoEventListener {
           eventAsMap.details = mapper.writeValueAsString(eventAsMap.details)
         }
 
-        if (service.config.wrap) {
+        if (restEventParser.isPresent()) {
+          RestEventParser.Event e = mapper.convertValue(event, RestEventParser.Event)
+          sentEvent = restEventParser.get().parse(e)
+        } else if (service.config.wrap) {
```
cfieber: think that there would be a case to have the eventParser process the event but still want the flatten or wrap behavior to happen on the processed event?

Thinking of a half-baked example of an eventParser that augments all the events with something like an instanceId that handled the event but otherwise they fan out to all the registered services as normal.

Maybe that is less of a concern if we get to the service.config specifying specific eventParsers to apply and make the final form totally up to the eventParser.

I could honestly go either way here..

jonsie: I don't think `flatten` will work with a custom parser (since flatten requires `content` and `details` fields and a custom parser could parse the event to any schema).  However, I can make this work with `wrap`.

PR #707, opened 2019-11-22
feat(stats): Pull apart the cloudProviders passed along during applic…

ttomsu:
…ation creation

Fixes https://github.com/spinnaker/spinnaker/issues/5168

Application creation/updating annoyingly can include multiple `cloudProviders`. The easiest, non-model-breaking way to achieve this is to emit N `createApplication` stages for the stat collection - one for each cloud provider.

### DIFF
--- a/echo-telemetry/src/main/java/com/netflix/spinnaker/echo/telemetry/TelemetryEventListener.java
+++ b/echo-telemetry/src/main/java/com/netflix/spinnaker/echo/telemetry/TelemetryEventListener.java
@@ -126,9 +126,10 @@ public class TelemetryEventListener implements EchoEventListener {
 
     List<Stage> protoStages =
         execution.getStages().stream()
-            .map(this::toStage)
+            .map(this::toStages)
             .filter(Optional::isPresent)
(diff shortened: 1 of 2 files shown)


echo-telemetry/src/main/java/com/netflix/spinnaker/echo/telemetry/TelemetryEventListener.java
```diff
@@ -38,10 +38,7 @@
 import io.github.resilience4j.circuitbreaker.CallNotPermittedException;
 import io.github.resilience4j.circuitbreaker.CircuitBreakerRegistry;
 import java.nio.charset.StandardCharsets;
-import java.util.ArrayList;
-import java.util.List;
-import java.util.Optional;
-import java.util.Set;
+import java.util.*;
```
ethanfrogers: Do we really want `*` here?

ttomsu: Silly IDE... Reverted.

PR #708, opened 2019-11-25
feat(extension): Add flattenEvent parsing to RestEventParser and convert Groovy to Java in echo-rest

jonsie:
Decided to add the `flattenEvent` to `RestEventParser` (following up on @cfieber suggestion to support that feature from last week) and as I was doing that I started converting some things to Java and then I ended up converting it all to Java. 

I could break this PR up into multiple PRs if we feel that's better to review.


echo-rest/src/main/java/com/netflix/spinnaker/echo/config/RestConfig.java
```diff
@@ -0,0 +1,141 @@
...
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+import retrofit.RequestInterceptor;
+import retrofit.RestAdapter;
+import retrofit.converter.JacksonConverter;
+
+/** Rest endpoint configuration */
+@Configuration
+@ConditionalOnProperty("rest.enabled")
+@SuppressWarnings("GStringExpressionWithinString")
```
cfieber: this is probably not necessary now that this is in java

echo-rest/src/main/java/com/netflix/spinnaker/echo/events/RestEventListener.java
```diff
@@ -0,0 +1,140 @@
...
+import org.slf4j.LoggerFactory;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.beans.factory.annotation.Value;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
+import org.springframework.stereotype.Component;
+
+/** Event listener for echo events */
+@Component
+@ConditionalOnProperty("rest.enabled")
+@SuppressWarnings("GStringExpressionWithinString")
```
cfieber: this too

PR #711, opened 2019-12-02
feat(mergify): Allow OSS approvers to autosubmit

ttomsu:
https://github.com/spinnaker/spinnaker/issues/5189


.mergify.yml
```diff
@@ -10,6 +10,17 @@ pull_request_rules:
         strict: smart
       label:
         add: ["auto merged"]
+  - name: Automatically merge PRs from maintainers on CI success and review
```
marchello2000: If we do decide to keep this change, the description should be accurate
```suggestion
  - name: Automatically merge PRs from maintainers on CI success
```

PR #726, opened 2019-12-11
fix(github): Put back github endpoint configuration

ferwguerra:
I restore the endpoint in the Github configuration. If it doesn't receive any value, the default value that already exists will remain.

### DIFF
--- a/echo-notifications/src/test/groovy/com/netflix/spinnaker/echo/config/GithubConfigSpec.groovy
+++ b/echo-notifications/src/test/groovy/com/netflix/spinnaker/echo/config/GithubConfigSpec.groovy
@@ -20,16 +20,4 @@ class GithubConfigSpec extends Specification {
     endpoint.url == ownEndpoint
   }
 
-  def 'test github default endpoint is correctly setted'() {
-    given:
-    String defaultEndpoint = "https://api.github.com"
-    githubConfig.endpoint = defaultEndpoint;
-
-    when:
-    Endpoint endpoint = githubConfig.githubEndpoint()
-
-    then:
-    endpoint.url == defaultEndpoint
-  }
-
 }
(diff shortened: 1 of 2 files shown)


xavileon:
@spinnaker/oss-reviewers PTAL

echo-notifications/src/test/groovy/com/netflix/spinnaker/echo/config/GithubConfigSpec.groovy
```diff
@@ -0,0 +1,35 @@
+package com.netflix.spinnaker.echo.config
+
+import spock.lang.Specification
+import spock.lang.Subject
+import retrofit.Endpoint
+
+class GithubConfigSpec extends Specification {
```
ttomsu: These tests don't actually do anything useful, and don't actually test that Spring is injecting the right value. Please remove.

ferwguerra: Thank you!
The intention of the test is to ensure that the "endpoint" attribute is not removed, because this happened and broke our process. I think the test helps this not happen again.
Of course, if you have any suggestions, it is welcome!

ttomsu: Got it - I see in the history where it was removed.

In that case, I would still advocate for removing the second test, as it appears to be a [change detector test](https://testing.googleblog.com/2015/01/testing-on-toilet-change-detector-tests.html).

ferwguerra: Done! Thanks.

xavileon:
Thanks @ttomsu!
@spinnakerbot cherry-pick 1.17

ttomsu:
@xavileon 1.17 just got released today, and will likely be the last release of 1.17 and of 2019. 1.18 gets cut first week of January.

PR #733, opened 2019-12-16
perf(pipelines): Use new endpoint to fetch pipeline config

ezimanyi:
The only call from echo to the front50 /pipelines/id/history endpoint passes limit=1. Replace it with the new endpoint /pipelines/id/get that gets the current state of the pipeline config and does so more performantly than the history endpoint.

Requires front50 >= 2.11.0, which is when this new endpoint was added.

### DIFF
--- a/echo-core/src/main/java/com/netflix/spinnaker/echo/services/Front50Service.java
+++ b/echo-core/src/main/java/com/netflix/spinnaker/echo/services/Front50Service.java
@@ -15,9 +15,8 @@ public interface Front50Service {
   @Headers("Accept: application/json")
   List<Pipeline> getPipelines(@Path("application") String application);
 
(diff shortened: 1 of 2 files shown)


dreynaud:
If I remember correctly the idea of using the `history?limit=1` endpoint was to punch through front50's in-mem cache. Are we sure `/pipelines/id/get` does the same?

ezimanyi:
@dreynaud : Yes, I just added the new endpoint in spinnaker/front50#644 (as prior to that there was no endpoint to get the config for a single pipeline).  It just calls `findById` which calls out to `loadObject` on the underlying data store.  There *is* a `Hystrix` fallback to use the cache, but only if the circuit breaker on calling to the data store has tripped.  Let me know if this looks good based on that.  Thanks!

PR #738, opened 2020-01-08
fix(slack): Allow specifying full url for webhooks integration

diasjorge:
In a previous PR we addressed spinnaker/spinnaker#4634 by allowing to specify the baseURL for slack. Unfortunately we're always posting to BASE_URL/services/TOKEN because of the way the client is declared.

With this patch I move the /services part to the base url and we append the token to the end of the url, this way the user can provide the full url to their webhook endpoint.

E.g: https://example.com/webhooks, we'll send a post request for https://example.com/webhooks/${token} instead of https://example.com/webhooks/services/${token} as we would currently do.


diasjorge:
@marchello2000 I'm tagging you as discussed in the original issue. Thanks for your attention

PR #741, opened 2020-01-13
fix(stats): Fix NPE when pipeline explicitly sends 'source=null'

ttomsu:
It seems the explicitly set `source=null` comes across the wire to Echo often enough that it halted all testing stats for the past few days. Had `source` been not set the default object would be fine.

This will need to be cherry picked into 1.18


duftler:

PR #744, opened 2020-01-14
feat(notifications): Introduce support for interactive notifications

luispollo:
Introduces a generic framework for interactive notifications in `echo`, with an initial implementation for Slack.

The flow is as follows:
image

And this is what an [example implementation looks like with `keel` as the originating service](https://github.com/spinnaker/keel/pull/723) and notifications sent to Slack:
Manual Jugment

More comments on the design inline.


echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/api/Notification.groovy
```diff
@@ -50,4 +54,40 @@ class Notification {
     NORMAL,
     HIGH
   }
+
+  /**
+   * Allows Spinnaker services sending Notifications through Echo to specify one or more interactive actions
+   * that, when acted upon by a user, cause a callback to Echo which gets routed to that originating service.
+   */
+  static class InteractiveActions {
```
luispollo: The classes here provide the generic framework for Spinnaker services to generate interactive notifications that are not specific to any implementation (Slack, e-mail, whatever). Currently supporting only buttons, but we could add whatever we want/need to the framework.

echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/config/SlackConfig.groovy
```diff
@@ -77,7 +77,7 @@ class SlackConfig {
         .setEndpoint(slackEndpoint)
         .setConverter(new JacksonConverter())
         .setClient(retrofitClient)
-        .setLogLevel(retrofitLogLevel)
+        .setLogLevel(RestAdapter.LogLevel.FULL)
```
luispollo: Just for testing, will revert.

echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/controller/NotificationController.groovy
```diff
@@ -32,8 +38,25 @@ class NotificationController {
...
     notificationServices?.find { it.supportsType(notification.notificationType) }?.handle(notification)
   }
+
+  @RequestMapping(
+    value = '/callbacks/{source}',
+    method = RequestMethod.POST,
+    consumes = MediaType.APPLICATION_FORM_URLENCODED_VALUE,
+    produces = MediaType.APPLICATION_JSON_VALUE
+  )
+  // Required due to https://github.com/spring-projects/spring-framework/issues/22734
```
luispollo: Never mind. I initially was trying to reuse `WebhooksController`.

echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/notification/InteractiveNotificationCallbackHandler.groovy
```diff
@@ -0,0 +1,97 @@
...
+  Ok3Client spinnakerServiceClient
+
+  @Autowired
+  List<InteractiveNotificationService> notificationServices
+
+  @Autowired
+  private Environment environment
+
+  /**
+   * Processes a callback request from Slack by relaying it to the downstream service which originated the message
```
luispollo: Will fix the comment. This actually finds the right notification service to delegate to based on the `source` parameter.

echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/notification/InteractiveNotificationCallbackHandler.groovy
```diff
@@ -0,0 +1,97 @@
...
+    log.debug("Received callback response from downstream Spinnaker service: ${response}")
+    //      5,
+    //      2000,
+    //      false);
+    //  }
+
+    // Allows the notification service implementation to respond to the callback as needed
+    notificationService.respondToCallback(content)
+
+    return null
```
luispollo: For Slack, this is just an acknowledgement so there's no body to return.

echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/notification/InteractiveNotificationCallbackHandler.groovy
```diff
@@ -0,0 +1,97 @@
...
+    InteractiveActionCallback callback = notificationService.parseInteractionCallback(content)
+    String requestUrl = environment.getProperty("${callback.serviceId}.baseUrl") + "/notifications/callback"
+    SpinnakerService spinnakerService = new RestAdapter.Builder()
+      .setEndpoint(newFixedEndpoint(requestUrl))
+      .setClient(spinnakerServiceClient)
+      .setLogLevel(RestAdapter.LogLevel.BASIC)
+      .setLog(new Slf4jRetrofitLogger(SpinnakerService.class))
+      .setConverter(new JacksonConverter())
+      .build()
+      .create(SpinnakerService.class)
```
luispollo: This allows us to dynamically get a client for any of the Spinnaker services in the config, without having to hard-code names and have a `switch` statement.

echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/notification/NotificationTemplateEngine.groovy
```diff
@@ -45,12 +45,12 @@ class NotificationTemplateEngine {
     String build(Notification notification, Type type) {
         if (!notification.templateGroup) {
           if (type == Type.SUBJECT) {
-            return (FORMATTERS.find { it.type == notification.additionalContext.formatter }?: new MarkdownToHtmlFormatter())
+            return (FORMATTERS.find { it.type == notification.additionalContext.formatter }?: new MarkdownPassThruFormatter())
```
luispollo: I believe this was a bug caused by recent updates in the way Slack renders text/markdown/HTML. Instead of rendering formatted content, Slack was displaying the HTML code. Having a pass-through formatter allows markdown to render correctly.

echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/slack/SlackNotificationService.groovy
```diff
@@ -58,12 +82,95 @@ class SlackNotificationService implements NotificationService {
...
+
+  @Override
+  void respondToCallback(Map content) {
+    Map payload = parseSlackPayload(content)
+    log.debug("Responding to Slack callback via ${payload.response_url}")
+
+    // Example: https://hooks.slack.com/actions/TSJ9N5Z3J/906732166263/3zDfYbq62ei4eNELNJcAaK0C
+    URI responseUrl = new URI(payload.response_url)
+
+    SlackHookService slackHookService = new RestAdapter.Builder()
```
luispollo: Instantiating a service client dynamically here is required because the URL is sent by Slack as part of the payload, i.e. we don't know it upfront (or even if the host/format won't change).

echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/slack/SlackService.groovy
```diff
@@ -37,6 +37,10 @@ class SlackService {
       slackClient.sendMessage(token, toJson(message), channel, asUser)
   }
 
+  Response respondToMessage(String responseUrl, Map content) {
+    slackClient.respondToMessage(responseUrl, content)
+  }
+
```
luispollo: Not used. Will remove.

echo-rest/src/main/java/com/netflix/spinnaker/echo/rest/OkHttpClientFactoryImpl.java
```diff
@@ -34,8 +33,8 @@ public OkHttpClient getInsecureClient() {
 
     try {
       // Create a trust manager that does not validate certificate chains
-      final TrustManager[] trustAllCerts =
-          new TrustManager[] {
+      final X509TrustManager[] trustAllCerts =
```
luispollo: This was me trying to debug TLS errors. Will revert.

echo-webhooks/src/main/groovy/com/netflix/spinnaker/echo/controllers/WebhooksController.groovy
```diff
@@ -103,7 +106,11 @@ class WebhooksController {
       WebhookResponse.newInstance(eventProcessed: false);
   }
 
-  @RequestMapping(value = '/webhooks/{type}', method = RequestMethod.POST)
+  @RequestMapping(
```
luispollo: Will revert.

echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/controller/NotificationController.groovy
robzienert: So is this code going away?

echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/notification/InteractiveNotificationCallbackHandler.groovy
```diff
@@ -0,0 +1,97 @@
+/*
```
robzienert: Do not add new Groovy files to the main source set.

echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/controller/NotificationController.groovy
```diff
@@ -32,8 +38,25 @@ class NotificationController {
...
+  @Autowired
+  InteractiveNotificationCallbackHandler callbackHandler
+
   @RequestMapping(method = RequestMethod.POST)
   EchoResponse create(@RequestBody Notification notification) {
     notificationServices?.find { it.supportsType(notification.notificationType) }?.handle(notification)
   }
+
+  @RequestMapping(
+    value = '/callbacks/{source}',
```
luispollo: This would be a new external, unauthenticated API exposed through gate.

echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/notification/InteractiveNotificationCallbackHandler.groovy
```diff
@@ -0,0 +1,97 @@
...
+import retrofit.http.Body
+import retrofit.http.Header
+import retrofit.http.POST
+
+import static retrofit.Endpoints.newFixedEndpoint
+
+@Slf4j
+@Component
+class InteractiveNotificationCallbackHandler {
+  @Autowired
```
robzienert: Prefer constructor over property autowiring at all times.

echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/notification/InteractiveNotificationCallbackHandler.groovy
```diff
@@ -0,0 +1,97 @@
...
+   * @param event
+   */
+  Map processCallback(String source, Map content, HttpHeaders headers) {
+    log.debug("Received interactive notification callback request from ${source}")
+
+    InteractiveNotificationService notificationService = notificationServices.stream()
+      .find { it.supportsType(Notification.Type.valueOf(source.toUpperCase())) }
+
+    if (notificationService == null) {
+      throw new IllegalArgumentException("NotificationService for ${source} no registered")
```
robzienert: Please use `SpinnakerException` classes. If you need to, write a new exception class for this case.

echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/notification/InteractiveNotificationCallbackHandler.groovy
robzienert: These service objects are going to be something you'll want to cache. Creating new retrofit clients is not a cheap operation.

echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/slack/SlackAttachment.groovy
```diff
@@ -38,9 +45,60 @@ class SlackAttachment {
...
+  private List<SlackAction> extractSlackActions(InteractiveActions interactiveActions) {
+    List<SlackAction> slackActions = interactiveActions.actions.stream()
+      .filter() { it instanceof ButtonAction }
+      .map { SlackButtonAction.copy(it) }
+      .collect(Collectors.toList())
+    slackActions
+  }
+}
+
+class SlackAction {
```
robzienert: Is this class immutable? If so, maybe `final` on the properties?

echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/slack/SlackNotificationService.groovy
```diff
@@ -58,12 +82,95 @@ class SlackNotificationService implements NotificationService {
...
+      )
+    )
+  }
+
+  @Override
+  void respondToCallback(Map content) {
+    Map payload = parseSlackPayload(content)
+    log.debug("Responding to Slack callback via ${payload.response_url}")
+
+    // Example: https://hooks.slack.com/actions/TSJ9N5Z3J/906732166263/3zDfYbq62ei4eNELNJcAaK0C
```
robzienert: This URL is likely something you should sanitize before putting into version control.

echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/slack/SlackNotificationService.groovy
robzienert: Again, you should cache these services.

echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/controller/NotificationController.groovy
luispollo: No, just the comment. 😃

echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/notification/InteractiveNotificationCallbackHandler.groovy
luispollo: I had a hunch about that, but was waiting for the pros to confirm. 😉 

Would instantiating just the `Retrofit.Builder` in Spring config and then building the service object here be a good-enough compromise, or is the introspection of the service class in `create(SpinnakerService.clas)` that is costly?

echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/notification/InteractiveNotificationCallbackHandler.groovy
luispollo: OK, OK...👮 It was a single file and there were not any existing Java sources in this module, so I thought it was not a big deal. But will convert.

echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/slack/SlackAttachment.groovy
luispollo: I wasn't 100% sure how Jackson works with Groovy classes. Will give it a try.

luispollo:
@robzienert Thanks for the quick feedback. I was hoping for some general feedback on the overall design -- the routing of messages, the modeling of the interactive action objects, etc. Do you feel like this is generally going in the right direction?

echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/slack/SlackNotificationService.groovy
luispollo: How do I cache these if the URL is different every time? If we can assume the host and URI format are stable, then I can parameterize the call instead.

echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/api/Notification.groovy
```diff
@@ -50,4 +54,40 @@ class Notification {
...
+    String callbackServiceId
+    String callbackMessageId
+    String color = '#cccccc'
```
jonsie: Does `color` need to be a property on `InteractiveActions`?

echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/notification/InteractiveNotificationCallbackHandler.groovy
jonsie: Hmm, what would be in this map if the function were to return a map?  Maybe the return type should be void for now?  I think returning `null` due to the one Slack implementation should probably be avoided.

jonsie:
I think the general pattern looks good. 👍

echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/api/Notification.groovy
luispollo: I was loosely basing the design on the Slack API, and they associate the color with each "attachment" (a.k.a. "block") of a message -- it basically controls the color of the vertical bar to the left of each block/part of the message. Since we only support a single attachment (it was already like this before my change), I figured it would go with `InteractiveActions` which eventually gets translated into the attachment.

echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/notification/InteractiveNotificationCallbackHandler.groovy
luispollo: I have no idea what the other types of services we integrate with expect from these kinds of webhook interactions, so I was trying to make it as generic as possible -- take a JSON payload, return a JSON payload. If you got nothing to return, return null. Could change this to an `Optional` perhaps? Do RestController + jackson handle that automagically like they do nulls?

jonsie: I'm not sure if there's any automagic stuff regarding `RestController` and `Optional`, but let's use `Optional` here so we don't have to refactor this later when there is another interactive notification implementation that returns something.

echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/controller/NotificationController.groovy
gregjones: I think Slack will include a token/shared-secret in this request that can be used to verify it was the sender - can checking that be part of this flow?

luispollo: Yes, absolutely! I already had a `TODO` somewhere to do that. What I meant here was just that, unlike other calls coming through gate, this won’t be using a Spinnaker identity to authenticate.

luispollo:
@robzienert I think this is now in a decent state for a full review.

echo-notifications/echo-notifications.gradle
```diff
@@ -36,3 +37,14 @@ dependencies {
...
     testImplementation "org.apache.httpcomponents:httpclient"
 }
+
+sourceSets {
+  main {
+    java {
+      srcDirs = [] // don't compile Java code twice
+    }
+    groovy {
+      srcDirs = [ 'src/main/groovy', 'src/main/java' ]
```
luispollo: This was required due to introduction of the new Java sources in `echo-notification`, which depend on the Groovy classes.

echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/notification/NotificationTemplateEngine.groovy
```diff
@@ -129,4 +136,20 @@ class NotificationTemplateEngine {
         return renderer.render(document)
       }
     }
+
+  static class MarkdownPassThruFormatter implements Formatter {
```
luispollo: I found out that, probably due to recent updates to how the Slack clients handle rendering of text, the default formatter above was causing the client to display HTML code (e.g. `<em>bold</em>`) instead of rendering the original markdown (`*bold*`) properly.

echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/slack/SlackClient.groovy
```diff
@@ -33,9 +34,17 @@ interface SlackClient {
   /**
    * Documentation: https://api.slack.com/incoming-webhooks
    */
-  @POST('/{token}')
+  @POST('/services/{token}')
```
luispollo: Moved the `/services` prefix here from `SLACK_INCOMING_WEBHOOK` in `SlackConfig`, which is now also used for other API calls that don't use that same path.

echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/config/SlackConfig.groovy
```diff
@@ -40,38 +45,49 @@ import static retrofit.Endpoints.newFixedEndpoint
...
-  final static String SLACK_INCOMING_WEBHOOK = 'https://hooks.slack.com/services'
+  final static String SLACK_INCOMING_WEBHOOK = 'https://hooks.slack.com'
   final static String SLACK_CHAT_API = 'https://slack.com'
 
-  @Value('${slack.base-url:}')
-  String slackBaseUrl
+  @ConfigurationProperties(prefix = "slack")
+  @Component
+  class SlackProperties {
+    private String _baseUrl
```
emjburns: why the `_` here?

echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/slack/SlackService.groovy
```diff
@@ -17,27 +17,104 @@
...
+  Response sendCompactMessage(CompactSlackMessage message, String channel, boolean asUser) {
+    slackClient.sendMessage(configProperties.token, message.buildMessage(), channel, asUser)
+  }
 
-  Response sendCompactMessage(String token, CompactSlackMessage message, String channel, boolean asUser) {
-    slackClient.sendMessage(token, message.buildMessage(), channel, asUser)
+  Response sendMessage(SlackAttachment message, String channel, boolean asUser) {
+    configProperties.useIncomingWebhook ?
+      slackClient.sendUsingIncomingWebHook(configProperties.token, new SlackRequest([message], channel)) :
+      slackClient.sendMessage(configProperties.token, toJson(message), channel, asUser)
```
emjburns: Is the slack token constant, or does it change periodically? Wondering if this token is like a github token which is generated once, or a more message-specific token

echo-notifications/src/main/java/com/netflix/spinnaker/echo/notification/InteractiveNotificationCallbackHandler.java
```diff
@@ -0,0 +1,150 @@
...
+    }
+
+    final Notification.InteractiveActionCallback callback =
+        notificationService.parseInteractionCallback(request);
+
+    SpinnakerService spinnakerService = getSpinnakerService(callback.getServiceId());
+
+    log.debug("Routing notification callback to originating service " + callback.getServiceId());
+
+    // TODO: error handling (retries?)
```
emjburns: You should address this todo or remove the commented out code

emjburns:
@luispollo are there any breaking changes here? I don't know enough about the way that slack works, but I do see some changes in the urls and the tokens

echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/slack/SlackService.groovy
luispollo: Permanent (at least until you manually rotate it in the Slack admin UI). This was already configured for our deployment.

echo-notifications/src/main/java/com/netflix/spinnaker/echo/notification/InteractiveNotificationCallbackHandler.java
luispollo: I'm currently unable to respond to messages in the chat to create a thread, pending help from Slack support, which is why I got stuck here and added the `TODO`. Was hoping to tackle that in a later PR if that's OK. The next best thing would be to replace the buttons in the original message with some error information, which sounded less than ideal from a UX perspective. @erikmunson @gcomstock Any thoughts on this?

luispollo:
> @luispollo are there any breaking changes here? I don't know enough about the way that slack works, but I do see some changes in the urls and the tokens

I know... I was trying to avoid making too many changes to the existing Slack integration, but the code was pretty old and it was hard/inefficient to add new stuff, so I refactored things a bit. I tried to be as careful as possible to maintain the existing contracts. The only riskier change I'd say is that webhook URL, but it was only used in a single place, so I was hoping it would be fine. Now that I say that, though, I realize people may have extended these classes in their private repos, and that constant is public. I'll revert that change and introduce a new constant. Thanks for raising the question.

echo-notifications/src/main/java/com/netflix/spinnaker/echo/notification/InteractiveNotificationCallbackHandler.java
emjburns: Feel free to keep the `todo`, I'd just remove the commented out retry code.

echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/config/SlackConfig.groovy
luispollo: Because the actual property name is `baseUrl` which is now accessible via the getter/setter. This would much easier if it was Kotlin, but it's Groovy, so I had to use this trick... This is an example of refactoring how the Slack config was implemented but keeping the existing contract.

echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/api/Notification.groovy
robzienert: I think for something like this, having a `Map<String, String> attributes` might be better?

echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/config/SlackConfig.groovy
robzienert: I don't follow. Spring (should) see that there's a setter and use that, rather than using reflection to set the property directly.

echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/notification/NotificationService.groovy
```diff
@@ -17,9 +17,31 @@
...
 import com.netflix.spinnaker.echo.controller.EchoResponse
+import org.springframework.http.RequestEntity
+import org.springframework.http.ResponseEntity
 
 interface NotificationService {
   boolean supportsType(Notification.Type type)
   EchoResponse handle(Notification notification)
 }
+
+interface InteractiveNotificationService extends NotificationService {
```
robzienert: Would like to see docs on the interface.

robzienert:
👏 This is great!

echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/api/Notification.groovy
luispollo: I didn’t want to do that because I need to enforce this generic contract for any notification service implementation. That’s one of the requirements: they need to be able to provide this `serviceId` and `messageId` while handling the notification callback. Or did you mean about the `color` field only?

echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/config/SlackConfig.groovy
luispollo: The catch is that I need to be able to set the property but manipulate the return in the getter. I did this to maintain backwards compatibility with how config was being used.

echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/config/SlackConfig.groovy
```diff
@@ -37,41 +37,25 @@ import static retrofit.Endpoints.newFixedEndpoint
...
-    if (StringUtils.isNotBlank(slackBaseUrl)) {
-      endpoint = slackBaseUrl
-    } else {
-      endpoint = useIncomingWebHook ? SLACK_INCOMING_WEBHOOK : SLACK_CHAT_API;
-    }
-
-    log.info("Using Slack {}: {}.", useIncomingWebHook ? "incoming webhook" : "chat api", endpoint)
-
-    newFixedEndpoint(endpoint)
-  }
```
luispollo: These config properties are now encapsulated in `SlackLegacyProperties`.

luispollo:
@robzienert I ended up having to refactor the Slack integration further in order to support two different Slack client configs simultaneously in echo, one for the legacy bot integration and another for the new app integration. The legacy approach is [essentially deprecated by Slack](https://api.slack.com/custom-integrations) at this point, but we use it at Netflix and I suspect many other companies do too. The new Slack app approach is required to support interactive notifications. The reason I decided to support two client configs for each type of Slack integration is that we cannot simply convert our old bot into a new one, so adding a second config allows us to start using interactive notifications with a new bot while we transition away from the old one, which would still work in the meantime. Hope this will make sense.

If you're OK with this changes, please approve the PR so I can get this deployed on Monday for some testing before I take off for the next couple of weeks.

Thanks again for your reviews!

echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/api/Notification.groovy
robzienert: Just the `color` field.

luispollo:
> Wow, what a chunk of work this turned into. Great job - especially having to work through all the warts!

I know, right?! Little did I know when I set off on this mission! 😅 

> If you're interested, I'd be happy to pair up with you on paying down some of this tech debt in Echo in the future!

Absolutely! I'd be happy to help convert some of the modules to Kotlin. Curious if we could leave the tests alone, though? Anyway, that's a whole other mission! Thanks again!

echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/api/Notification.groovy
luispollo: OK. Let's refactor that some other time! 👍

echo-notifications/src/main/java/com/netflix/spinnaker/echo/notification/InteractiveNotificationCallbackHandler.java
gcomstock: Thanks for calling me out Luis and sorry for my delay. Sounds like you took the best course of action for now. Enjoy your trip!

PR #745, opened 2020-01-15
fix(email): Use correct Spinnaker link when sending email notifications

dbrougham:
Fix for https://github.com/spinnaker/spinnaker/issues/5314


ttomsu:
I'm not sure that this change doesn't introduce a different bug.

In spinnaker/spinnaker#5314, your example contains a `customBody`:

```
"notifications": [
      {
        "address": "dbrougham92@gmail.com",
        "customBody": "*Pipeline parameters:* ${parameters.toString()}\n\n [View the stage]({{link}}) here.",
        "customSubject": "${ templateVariables.appName } Production Deploy Failed",
        "level": "pipeline",
        "type": "email",
        "when": [
          "pipeline.starting"
        ]
      }
    ]
```

In the implementation of `sendMessage`, the body is [rendered differently if](https://github.com/spinnaker/echo/blob/f92a9ca366706826332cf5b0cf9692f3122b6148/echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/notification/EmailNotificationAgent.groovy#L102) `customBody` is present.

By changing the value of `link`, you're fixing your bug but breaking non-customBody templates (specifically [the pipeline template](https://github.com/spinnaker/echo/blob/f92a9ca366706826332cf5b0cf9692f3122b6148/echo-notifications/src/main/resources/templates/pipeline.ftl#L4)).

I've never worked with these templates, but maybe the `message` field + the stage.ftl or pipeline.ftl template would allow you to put your pipeline parameters in the body and still get a working link?

dbrougham:
Nice catch! 

That sounds like it would be valid work around.

I still want to tackle the underlying bug with emails using `customBody` and the `{{link}}` together. Will have another go at a code fix and do some more testing locally. I believe we could just use `{{link}}` directly for the link inside the `stage.ftl` and `pipeline.ftl` templates. This would result in both of these files being identical, so would then justify removing one of them and using a single template.

ttomsu:
@cfieber Would you be the person who knows the most about the templates in Echo? It looks like there is no difference between `pipeline.ftl` and `stage.ftl` anymore, but I've never touched this code before.

dbrougham:
@ttomsu - Now this has been approved, what are the next steps to get this merged into master?

PR #756, opened 2020-01-22
feat(plugins): Add kork-plugins dependency and import PluginsAutoConfiguration in echo-core

jonsie:


.gitignore
```diff
@@ -16,3 +16,4 @@ data
 .project
 .settings/
 bin/
+plugins/
```
robzienert: Pre-emptive, or is this something that pf4j is doing?

jonsie: When running locally with a remote update repository configured plugins are downloaded to the projects root/plugins directory (created if it doesn't exist).

PR #770, opened 2020-02-10
feat(telemetry): add ability to report deploy method if configured

dogonthehorizon:
This is a follow-up to spinnaker/kork#503 where we added deploy method and version to `kork-proto`. This change in echo allows events to be sent to the telemetry service with the deployment method and version _if defined_.

At the time of this PR there are currently no deployment methods (e.g. halyard) that support reporting this information (but it is on our immediate roadmap).


dogonthehorizon:
Build is failing because `echo` hasn't pulled in the version of `kork-proto` that has our changes :/
https://github.com/spinnaker/echo/pull/770/checks?check_run_id=436577669

What's the typical way we resolve these dependencies?

echo-telemetry/src/main/java/com/netflix/spinnaker/echo/config/TelemetryConfig.java
```diff
@@ -75,6 +77,7 @@ private OkClient telemetryOkClient(TelemetryConfigProps configProps) {
     String endpoint = DEFAULT_TELEMETRY_ENDPOINT;
     String instanceId = new ULID().nextULID();
     String spinnakerVersion = "unknown";
+    Map<String, String> deploymentMethod = new HashMap<>();
```
ttomsu: Please make this strongly typed

echo-telemetry/src/test/groovy/com/netflix/spinnaker/echo/telemetry/TelemetryEventListenerSpec.groovy
```diff
@@ -367,6 +367,77 @@ class TelemetryEventListenerSpec extends Specification {
...
+        assert e != null
+
+        SpinnakerInstance s = e.spinnakerInstance
+        assert s != null
+        assert s.deploymentMethod != null
+        assert s.deploymentMethod.type == deployEnumType
+        assert s.deploymentMethod.version == deployVersion
+      }
+
+      where:
```
ttomsu: I'd make this a test table if you're going to use that many pairs. 

http://spockframework.org/spock/docs/1.0/data_driven_testing.html#data-tables

echo-telemetry/src/test/groovy/com/netflix/spinnaker/echo/telemetry/TelemetryEventListenerSpec.groovy
```diff
@@ -367,6 +367,77 @@ class TelemetryEventListenerSpec extends Specification {
     }
   }
 
+  def "test event has deploy method information happy path"(String deploymentType, String deployVersion) {
```
ttomsu: Please add valid values to `validEvent` object.

echo-telemetry/src/test/groovy/com/netflix/spinnaker/echo/telemetry/TelemetryEventListenerSpec.groovy
```diff
@@ -367,6 +367,77 @@ class TelemetryEventListenerSpec extends Specification {
...
+        assert s.deploymentMethod.type == deployEnumType
+        assert s.deploymentMethod.version == deployVersion
+      }
+
+      where:
+      deploymentType << ["none", "other", "minnaker", "halyard", "armory_halyard", "kubernetes_operator"]
+      deployVersion << ["1.2.3", "3.2.1", "1.0.0", "3.3.3", "2.1.9", "99.99.99"]
+  }
+
+  def "test event, without deploy method info"() {
```
ttomsu: Aligned with other comments, I'd prefer to flip this. A full "validEvent" proto has _everything_, and there are tests that confirm that everything still works when _something_ is missing or not set.

ttomsu:
You'll need a Kork release to pull in the changes from `kork-proto`. I don't have time to shepherd that through today, but maybe someone else will need a release soon.

echo-telemetry/src/test/groovy/com/netflix/spinnaker/echo/telemetry/TelemetryEventListenerSpec.groovy
dogonthehorizon: Are there examples in the Spinnaker project of property/generative testing? Ideally we'd have tests for:
- expected enum values
- random strings that get passed in for type (and set to the appropriate default value)
- validating arbitrary version strings don't break the config object.

echo-telemetry/src/main/java/com/netflix/spinnaker/echo/config/TelemetryConfig.java
```diff
@@ -69,13 +70,31 @@ private OkClient telemetryOkClient(TelemetryConfigProps configProps) {
   @ConfigurationProperties(prefix = "telemetry")
   public static class TelemetryConfigProps {
 
+    public static final String DEFAULT_DEPLOY_METHOD = "none";
```
ttomsu: I'd probably prefer a `null` here rather than `none` - mostly because then the backend queries will have to distinguish `none` from `null` without much of a difference.

echo-telemetry/src/main/java/com/netflix/spinnaker/echo/telemetry/TelemetryEventListener.java
```diff
@@ -160,9 +155,22 @@ public void processEvent(Event event) {
...
+    DeploymentMethod deploymentMethod =
+        DeploymentMethod.newBuilder()
+            .setType(deployType)
+            .setVersion(telemetryConfigProps.getDeploymentMethod().getVersion().orElse("none"))
+            .build();
+
     SpinnakerInstance spinnakerInstance =
         SpinnakerInstance.newBuilder()
             .setId(hashedInstanceId)
+            .setDeploymentMethod(deploymentMethod)
```
ttomsu: I would probably say only add a deployment method iff the Optionals from getType and getVersion are not null. Otherwise, there's no added information.

echo-telemetry/src/test/groovy/com/netflix/spinnaker/echo/telemetry/TelemetryEventListenerSpec.groovy
ttomsu: I don't think so. It would be great if we could "fuzz test" our APIs, but AFAIK we don't have that.

echo-telemetry/src/test/groovy/com/netflix/spinnaker/echo/telemetry/TelemetryEventListenerSpec.groovy
```diff
@@ -367,6 +367,84 @@ class TelemetryEventListenerSpec extends Specification {
...
+      }
+
+      where:
+      deploymentType        | deployVersion
+      "none"                | "1.2.3"
+      "other"               | "3.2.1"
+      "minnaker"            | "3.3.3."
+      "halyard"             | "1.0.0"
+      "armory_halyard"      | "2.1.9"
+      "kubernetes_operator" | "88.88.88"
```
ttomsu: nitpick: these test cases aren't really different from one another. Do we need 6 of them?

castanedaci: Since an enum is being used for the type of deployment method we decided to test all probable values within the enum.
Please let me know if testing just a couple of these values is enough.

echo-telemetry/src/test/groovy/com/netflix/spinnaker/echo/telemetry/TelemetryEventListenerSpec.groovy
```diff
@@ -367,6 +381,84 @@ class TelemetryEventListenerSpec extends Specification {
     }
   }
 
+  def "test event has deploy method information happy path"(String deploymentType, String deployVersion) {
```
ttomsu: I don't think I've ever seen the test methods take parameters - are those necessary?

echo-telemetry/src/test/groovy/com/netflix/spinnaker/echo/telemetry/TelemetryEventListenerSpec.groovy
```diff
@@ -367,6 +381,84 @@ class TelemetryEventListenerSpec extends Specification {
...
+      }
+
+      where:
+      deploymentType        | deployVersion
+      "none"                | "1.2.3"
+      "other"               | "3.2.1"
+      "minnaker"            | "3.3.3."
+      "halyard"             | "1.0.0"
+      "armory_halyard"      | "2.1.9"
+      "kubernetes_operator" | "v0.3.0"
```
ttomsu: This brings up a good point of the "v" prefix - are you expecting it? I'd imagine that we'd want some consistency in the data - I'd probably opt to strip it but could be convinced otherwise.

echo-telemetry/src/main/java/com/netflix/spinnaker/echo/config/TelemetryConfig.java
```diff
@@ -75,7 +78,24 @@ private OkClient telemetryOkClient(TelemetryConfigProps configProps) {
     String endpoint = DEFAULT_TELEMETRY_ENDPOINT;
     String instanceId = new ULID().nextULID();
     String spinnakerVersion = "unknown";
+    DeploymentMethod deploymentMethod = new DeploymentMethod();
     int connectionTimeoutMillis = 3000;
     int readTimeoutMillis = 5000;
+
+    @Data
+    @NoArgsConstructor
+    @AllArgsConstructor
```
ttomsu: Do you need all of these? By my eyeball, it looks like you may only need `Data`, which has `NoArgsConstructor` when `AllArgsConstructor` is missing. IIRC Spring uses each fields setter method when deserializing the config, so AllArgsConstructor may not be necessary.

echo-telemetry/src/test/groovy/com/netflix/spinnaker/echo/telemetry/TelemetryEventListenerSpec.groovy
castanedaci: My bad, I took this version from release branches but ideally the versions will have consistency.

echo-telemetry/src/test/groovy/com/netflix/spinnaker/echo/telemetry/TelemetryEventListenerSpec.groovy
castanedaci: Since we are using [Spock data tables](http://spockframework.org/spock/docs/1.0/data_driven_testing.html) that's why we need parameters in this test method.
The reason for the use of data tables is that we want to be sure that all the possible scenarios for [deployment Type] (https://github.com/spinnaker/kork/blob/master/kork-proto/src/main/proto/stats/log.proto) are well-tested.

ttomsu: You don't need to specify the parameters - Spock will take care of that for you.

PR #777, opened 2020-02-14
feat(slack): Add ability for slack to expand user names

marchello2000:
For backward compatibility, this is driven by a config that's off by default


echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/slack/SlackClient.groovy
```diff
@@ -23,13 +23,20 @@ import retrofit.http.*
 
 interface SlackClient {
 
+  /**
+   * Post a message via Slack API
+   *
+   * see https://api.slack.com/methods/chat.postMessage
+   * One note: linkUserNames according to slack API is a boolean. But guess what? It doesn't work, it must be an int 0 or 1
```
emjburns: HA! ha. that's great.

PR #779, opened 2020-02-16
fix(notifications): Fix precedence of notification handlers in controller

luispollo:
I assumed the logical `&&` would short-circuit the rest of the expression, but obviously not. Reading the [groovy docs](https://docs.groovy-lang.org/latest/html/documentation/core-operators.html#_logical_operators), I would expect it to work, but clearly it doesn't.

I added a test where an interactive notification handler is registered before a non-interactive one, which was the situation we had in our deployment and the reason you see this line in the stack trace in the log, which was the hint that led me to finding the bug:

```
at com.netflix.spinnaker.echo.slack.SlackInteractiveNotificationService.handle(SlackInteractiveNotificationService.groovy)
```
The second hint was that the error we experienced was when trying to send a PagerDuty notification, which should obviously not be handled by the Slack implementation. 😝

### DIFF
--- a/echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/controller/NotificationController.groovy
+++ b/echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/controller/NotificationController.groovy
@@ -53,7 +53,7 @@ class NotificationController {
   EchoResponse create(@RequestBody Notification notification) {
     notificationServices?.find {
       it.supportsType(notification.notificationType) &&
-        !notification.isInteractive() || it instanceof InteractiveNotificationService
+        (!notification.isInteractive() || it instanceof InteractiveNotificationService)
     }?.handle(notification)
   }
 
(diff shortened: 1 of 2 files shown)


echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/controller/NotificationController.groovy
```diff
@@ -53,7 +53,7 @@ class NotificationController {
   EchoResponse create(@RequestBody Notification notification) {
     notificationServices?.find {
       it.supportsType(notification.notificationType) &&
-        !notification.isInteractive() || it instanceof InteractiveNotificationService
+        (!notification.isInteractive() || it instanceof InteractiveNotificationService)
```
luispollo: This was it... 😝

dreynaud: Thanks for fixing @luispollo but I got to say that I find this super hard to read and reason about. And if we fail to find a `notificationService` that satisfies the predicate, we just silently drop the notification? It seems like we should at the very least log an error message about it, maybe even emit a metric that we can get alerted about. What do you think?

luispollo: Hi @dreynaud, thanks for the feedback and sorry for the late reply. I'll add a logging statement there.

luispollo: P.S. In my defense, we were already dropping notifications that failed the previous check without logging... 😃 Not an excuse, though. I'll fix it.

PR #781, opened 2020-02-18
fix(pubsub): Default NodeIdentity to 'UnknownHost' to prevent NPE

ttomsu:
Fixes https://github.com/spinnaker/spinnaker/issues/5453

.. I _think_.

### DIFF
--- a/echo-pubsub-core/src/main/java/com/netflix/spinnaker/echo/pubsub/utils/NodeIdentity.java
+++ b/echo-pubsub-core/src/main/java/com/netflix/spinnaker/echo/pubsub/utils/NodeIdentity.java
@@ -25,21 +25,21 @@ import lombok.Getter;
 
 /** Provides an identity for each echo node. */
 public class NodeIdentity {
-  public static final String UNKNOWN_HOST = "UnknownHost";
+  private static final String UNKNOWN_HOST = "UnknownHost";
 
-  private String runtimeName;
-  private String hostName;
-  @Getter private String identity;
+  @Getter private final String identity;
 
   public NodeIdentity() {
     this("www.google.com", 80);
   }
 
   public NodeIdentity(String validationAddress, Integer validationPort) {
-    this.runtimeName = ManagementFactory.getRuntimeMXBean().getName();
-    this.hostName = resolveHostname(validationAddress, validationPort);
+    String runtimeName = ManagementFactory.getRuntimeMXBean().getName();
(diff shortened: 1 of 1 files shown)


echo-pubsub-core/src/main/java/com/netflix/spinnaker/echo/pubsub/utils/NodeIdentity.java
```diff
@@ -29,7 +29,7 @@
 
   private String runtimeName;
   private String hostName;
-  @Getter private String identity;
+  @Getter private String identity = UNKNOWN_HOST;
```
plumpy: I think this would be better in an `else` block in the constructor so that we're not setting this in two different places. Also you can make it `final` then. (And the other two fields can just be deleted, can't they? They're only used to calculate `identity`.)

ttomsu: Good suggestions, applied.

echo-pubsub-core/src/main/java/com/netflix/spinnaker/echo/pubsub/utils/NodeIdentity.java
```diff
@@ -25,21 +25,21 @@
...
 public class NodeIdentity {
-  public static final String UNKNOWN_HOST = "UnknownHost";
+  private static final String UNKNOWN_HOST = "UnknownHost";
 
-  private String runtimeName;
-  private String hostName;
-  @Getter private String identity;
+  @Getter private final String identity;
 
   public NodeIdentity() {
```
plumpy: nobody ever calls the other constructor, so can you just make `"www.google.com"` and `80` be constants?

And then, instead of doing a lot of expensive work in the constructor, which is generally discouraged, can you make it `private NodeIdentity(String identity)` and use a `public static NodeIdentity create()` method that does the heavy lifting?

ttomsu: Given that we will probably cherry pick this into a few previous branches, I would prefer to keep the change as small and understandable as possible. You're suggesting good changes, so lets fix the bug first and then improve the code health.

plumpy: Okay, fair enough.

ttomsu:

PR #783, opened 2020-02-18
chore(build): Update Dockerfiles for new cloudbuild configs

plumpy:
See spinnaker/buildtool#14

### DIFF
--- a/Dockerfile.compile
+++ b/Dockerfile.compile
@@ -1,6 +1,12 @@
-FROM gcr.io/spinnaker-marketplace/gradle_cache
-MAINTAINER delivery-engineering@netflix.com
-ENV GRADLE_USER_HOME /gradle_cache/.gradle
-COPY . compiled_sources
-WORKDIR compiled_sources
-RUN ./gradlew --no-daemon echo-web:installDist -x test
+FROM ubuntu:eoan
+MAINTAINER sig-platform@spinnaker.io
+
+RUN apt-get update \
+    && apt-get -y upgrade \
+    && apt-get install -y \
+        openjdk-8-jdk-headless \
+    && apt-get clean
+
+ENV GRADLE_USER_HOME /workspace/.gradle
+ENV GRADLE_OPTS -Xmx2048m
+CMD ./gradlew --no-daemon echo-web:installDist -x test
--- a/Dockerfile.compile
+++ b/Dockerfile.compile
@@ -1,12 +1,5 @@
-FROM ubuntu:eoan
+FROM openjdk:8
 MAINTAINER sig-platform@spinnaker.io
-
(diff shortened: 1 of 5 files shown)


Dockerfile.compile
```diff
@@ -1,6 +1,12 @@
...
-RUN ./gradlew --no-daemon echo-web:installDist -x test
\ No newline at end of file
+FROM ubuntu:eoan
+MAINTAINER sig-platform@spinnaker.io
+
+RUN apt-get update \
+    && apt-get -y upgrade \
+    && apt-get install -y \
+        openjdk-8-jdk-headless \
+    && apt-get clean
```
ttomsu: In the past, we've had flaky container builds doing an `apt upgrade`, not to mention that it takes a relatively long time. Since this is only an intermediate container, what do you think about simply basing it `FROM openjdk8`, just like `Dockerfile.java8`?

plumpy: Ooh, yeah, I forgot about those. That's way better, thanks. I'll work on updating all the PRs, make sure they still build, and then update this comment when I'm done.

plumpy: They all compiled fine with `openjdk:8`, so I've updated all 10 of the PRs whenever you get a chance to review, thanks.

Dockerfile.ubuntu
```diff
@@ -1,8 +1,7 @@
...
-MAINTAINER delivery-engineering@netflix.com
-COPY --from=compile /compiled_sources/echo-web/build/install/echo /opt/echo
+MAINTAINER sig-platform@spinnaker.io
+COPY echo-web/build/install/echo /opt/echo
 RUN apt-get update && apt-get -y install openjdk-11-jre-headless wget
 RUN adduser --disabled-login --system spinnaker
 USER spinnaker
-CMD ["/opt/echo/bin/echo"]
\ No newline at end of file
+CMD ["/opt/echo/bin/echo"]
```
ttomsu: Nit: You may want to use ENTRYPOINT here and elsewhere, based on my understanding from https://goinbigdata.com/docker-run-vs-cmd-vs-entrypoint/

This would allow someone to run the container passing new/different JVM flags (I think we've seen users in the past trying to pass -D args to overrride how Java handles SSL handshakes and whatnot).

Not sure if people would or how they would pass them through Halyard, but it seems more correct. Feel free to ignore until we get someone actually wanting this.

plumpy: Hm, that's definitely interesting... I skimmed and I'm not sure I totally get it, but it seems like you're right...

but that's really unrelated to these PRs (this line only got changed because `vim` added a newline) so I don't think this is the right time to change it anyway.

PR #790, opened 2020-02-27
refactor(api): Move EventListener and dependent classes to echo-api, remove previous RestEventParser

jonsie:
Todo:

- [x] Jackson serializer/deserializer for Event (to remove ` @JsonInclude(NON_NULL)` from echo-api).  Likely will be in the module `echo-api-jackson`.
- [x] Unify `ObjectMapper` configuration (so custom serializers/deserializers are always configured)

I also took the opportunity to rename `EchoEventListener` to just `EventListener`.


echo-api/src/main/java/com/netflix/spinnaker/echo/api/events/Event.java
```diff
@@ -14,22 +14,24 @@
...
 @Data
 public class Event {
   public Metadata details;
-  public Map content;
+  public Map<String, Object> content;
   public String rawContent;
 
-  @JsonInclude(JsonInclude.Include.NON_NULL)
-  public Map payload;
+  @JsonInclude(NON_NULL)
```
jonsie: Obviously this has to be removed.

echo-core/src/main/java/com/netflix/spinnaker/echo/jackson/mixin/EventMixin.java
```diff
@@ -0,0 +1,28 @@
...
+ */
+
+package com.netflix.spinnaker.echo.jackson.mixin;
+
+import static com.fasterxml.jackson.annotation.JsonInclude.Include.NON_NULL;
+
+import com.fasterxml.jackson.annotation.JsonInclude;
+import java.util.Map;
+
+public abstract class EventMixin {
```
jonsie: EventMixin' 🎧

echo-api/src/main/java/com/netflix/spinnaker/echo/api/events/Event.java
```diff
@@ -25,11 +24,9 @@
...
-  public Map content;
+  public Map<String, Object> content;
   public String rawContent;
-
-  @JsonInclude(JsonInclude.Include.NON_NULL)
-  public Map payload;
+  public Map<String, Object> payload;
 
   public String eventId = UUID.randomUUID().toString();
 }
```
robzienert: Is there value in `content` and `payload` being typed? I'd really like to avoid bringing `M<S, O>` into the `-api` modules if possible. Perhaps for Echo, we can't really do anything about it until the other services sending events in are well-defined?

echo-api/src/main/java/com/netflix/spinnaker/echo/api/events/EventListener.java
```diff
@@ -14,10 +14,10 @@
...
  */
 
-package com.netflix.spinnaker.echo.events;
+package com.netflix.spinnaker.echo.api.events;
 
-import com.netflix.spinnaker.echo.model.Event;
+import org.pf4j.ExtensionPoint;
 
-public interface EchoEventListener {
+public interface EventListener extends ExtensionPoint {
```
robzienert: Docs on interfaces.

echo-rest/src/main/java/com/netflix/spinnaker/echo/events/RestEventListener.java
```diff
@@ -74,21 +67,9 @@ public void processEvent(Event event) {
...
-                  eventMap.put("details", flatEvent.getDetails());
-                } else if (service.getConfig().getFlatten()) {
+                if (service.getConfig().getFlatten()) {
                   eventMap.put("content", mapper.writeValueAsString(eventMap.get("content")));
                   eventMap.put("details", mapper.writeValueAsString(eventMap.get("details")));
-                } else if (restEventParser.isPresent()) {
-                  eventMap =
-                      restEventParser
-                          .get()
-                          .parseEvent(mapper.convertValue(event, RestEventParser.Event.class));
```
robzienert: I see this code being removed here, but I don't recall seeing it put somewhere else. It's concerning to me that this didn't break any tests, either?

jonsie: This is the integration for the first extension point I added to echo (`RestEventParser`), but I think there's more value in pulling it out and redesigning how we build the echo-api.  It's not being used at all right now.

echo-api/src/main/java/com/netflix/spinnaker/echo/api/events/Event.java
jonsie: I think I'll merge as is and then explore how `content` and `payload` could be typed.  There's a lot to unwind WRT how `Event` and `TriggerEvent` (and the various `TriggerEvent` types) deserialize the `content` field and I don't feel comfortable making any changes there yet.

echo-api/src/main/java/com/netflix/spinnaker/echo/api/events/Metadata.java
```diff
@@ -33,6 +36,6 @@
   private String project;
   private String application;
   private String _content_id;
-  private Map attributes;
-  private HttpHeaders requestHeaders = new HttpHeaders();
+  private Map<String, String> attributes;
+  private Map<String, List<String>> requestHeaders = new HashMap<>();
```
luispollo, line 40: @jonsie It looks like this broke code that relied on the assumption of headers being converted to lowercase by `HttpHeaders`, as in #840. I think it might be best to revert this line instead of fixing each specific downstream issue that comes up as a result, unless there was a good reason to convert this particular property to a `Map`.

jonsie: Ah sorry, I didn't catch this.  Unfortunately, we can't pull in the Spring dependency into `echo-api` as this module needs to be dependency free (basically, just Java) , which is why I removed `HttpHeaders`.

jonsie: This might do the trick:

```
private Map<String, List<String>> requestHeaders = new TreeMap<>(String.CASE_INSENSITIVE_ORDER);
```

jonsie: https://github.com/spinnaker/echo/pull/863

PR #792, opened 2020-03-02
refactor(core): push halyard defaults into base configs defaults

link108:
This is similar to https://github.com/spinnaker/clouddriver/pull/4368. 

In order for Halyard to deploy Echo, it downloads `halconfig/echo.yml` from a gcs bucket tied to the version of echo. This causes issues when Halyard deploys plugins to echo, as `halconfig/echo.yml` contains `spinnaker` as a top level key. This PR moves most of the configuration out of `halconfig/echo.yml` and into `orca-web/config/orca.yml` to be defaults. 

There was a good amount of discussion in Slack on this topic, please see: https://spinnakerteam.slack.com/archives/CK9FK4XDF/p1582934283018700


echo-web/config/echo.yml
```diff
@@ -2,16 +2,24 @@ server:
   port: 8089
 
 spinnaker:
+  inMemory:
```
ezimanyi: Do you know where this is used?  I suppose it doesn't hurt anything to have this here if it's unused, but I couldn't find this ever being read.

echo-web/config/echo.yml
```diff
@@ -2,16 +2,24 @@ server:
...
+  enabled: ${services.front50.enabled:false}
+  baseUrl: ${services.front50.baseUrl:http://localhost:8080}
+
+igor:
+  enabled: ${services.igor.enabled:false}
+  baseUrl: ${services.igor.baseUrl:http://localhost:8088}
 
 orca:
-  baseUrl: 'http://localhost:8083'
+  enabled: ${services.orca.enabled:false}
```
ezimanyi: This will change the default, as currently the code defaults `orca.enabled` to `true`.  I think this should default to `true`.

echo-web/config/echo.yml
```diff
@@ -2,16 +2,24 @@ server:
...
+    enabled: true
   cassandra:
     cluster: 'spinnaker'
     keyspace: 'echo'
-  baseUrl: 'http://localhost:9000'
+  baseUrl: ${services.deck.baseUrl:http://localhost:9000}
 
 front50:
-  baseUrl: 'http://localhost:8080'
+  enabled: ${services.front50.enabled:false}
```
ezimanyi: It looks like we never actually read this in, so we should probably omit it to avoid confusion...from what I can tell `front50` is *always* enabled.

echo-web/config/echo.yml
```diff
@@ -25,6 +33,20 @@ webhooks:
     enabled: false
     sources: []
 
+scheduler:
+  enabled: ${services.echo.cron.enabled:true}
+  threadPoolSize: 20
```
ezimanyi: This defaults to 16 in code; I'd rather just remove it from the config file and fall back on that default.

echo-web/config/echo.yml
```diff
@@ -25,6 +33,20 @@ webhooks:
...
+  triggeringEnabled: true
+  pipelineConfigsPoller:
```
ezimanyi: This block is entirely already defaulted to these values in code; let's just let the code do the defaulting.

echo-web/config/echo.yml
```diff
@@ -25,6 +33,20 @@ webhooks:
     enabled: false
     sources: []
 
+scheduler:
+  enabled: ${services.echo.cron.enabled:true}
+  threadPoolSize: 20
+  triggeringEnabled: true
```
ezimanyi: I don't see this parameter being read anywhere.

echo-web/config/echo.yml
link108: removing this, I also don't see it being used

echo-web/config/echo.yml
link108: updating, odd that the configs had default to false, but it was defaulted to true in the code O_o

PR #793, opened 2020-03-03
chore(dependencies): update korkVersion

claymccoy:
This is for consistency with the other spin services. Mainly so that plugins can be written the same way across services. It would be nice to keep kork versions consistent for a release.


dreynaud:
thanks @claymccoy, you beat me to it 🙏

PR #801, opened 2020-03-07
Parameters in cron triggers

dreynaud:
Triggers support a parameters map, but for cron triggers it gets lost along the way and never makes it to orca, meaning that only default parameter values can be used for cron triggers.

To work around this, users have been duplicating pipelines with different default values and different triggers. This does not scale for obvious reasons and is just generally cumbersome.


dreynaud:
@luispollo good idea, I'm not sure what docs this should go into, pipeline json in general is quite underspecified. I'll announce it and see if people have suggestions 👍

PR #804, opened 2020-03-11
chore(stats): Rename Spring config props from telemetry to stats

ttomsu:
This is the core of this request in https://github.com/spinnaker/halyard/pull/1565#discussion_r390586727

The rest of the change involves renaming the whole module and a whole lot of `s/telemetry/stats/g` changes - I'd like to confirm that this minimal, understandable change actually works as expected and then we can rename the module in a separate cleanup PR.


ttomsu:

PR #816, opened 2020-03-19
fix(front50): Add SpinnakerRequestInterceptor to Front50 OkHttpClient

deepakr199:
This fixes [spinnaker/issues/5035](https://github.com/spinnaker/spinnaker/issues/5035) and [spinnaker/issues/4941](https://github.com/spinnaker/spinnaker/issues/4941)

This fix adds SpinnakerRequestInterceptor to Front50 OkHttpClient in Echo. 

Currently, Front50 OkHttpClient in Echo does not propagate the AuthenticationRequest [Headers](https://github.com/spinnaker/kork/blob/v7.5.1/kork-security/src/main/java/com/netflix/spinnaker/security/AuthenticatedRequest.java#L39) to Front50 on Http calls. Reason for this behaviour is due to Front50 Client created without SpinnakerRequestInterceptor. This interceptor ensures Headers are propagated as part of request headers on Http calls.

This has caused Echo not able to get the latest pipeline configuration from Front50 and rely on Pipeline configuration cache.
- Before spinnaker version 1.18, Echo used to call Front50 '[GetHistory](https://github.com/spinnaker/front50/blob/version-0.21.0/front50-web/src/main/groovy/com/netflix/spinnaker/front50/controllers/PipelineController.groovy#L98)' endpoint to get pipeline configuration and was returned empty object due to @PostFilter annotation. This forces Echo to pick up configuration from Pipeline cache. 
- With version 1.18, Echo calls Front50 '[GetPipeline](https://github.com/spinnaker/front50/blob/version-0.21.0/front50-web/src/main/groovy/com/netflix/spinnaker/front50/controllers/PipelineController.groovy#L105)' endpoint will return 403 due to @PostAuthorize which results in again picking up configuration from Pipeline cache.

Note: SpinnakerRequestInterceptor is already set in Igor [OkHttpClient](https://github.com/spinnaker/echo/blob/version-2.10.1/echo-pipelinetriggers/src/main/java/com/netflix/spinnaker/echo/config/IgorConfig.java#L54) configuration in Echo, so Auth headers are propagated on http calls to Igor.

Edit: As per @plumpy feedback, changing the Igor OkHttpClient configuration to use `SpinnakerRequestInterceptor`

### DIFF
--- a/echo-web/src/main/java/com/netflix/spinnaker/echo/config/Front50Config.java
+++ b/echo-web/src/main/java/com/netflix/spinnaker/echo/config/Front50Config.java
@@ -18,6 +18,7 @@ package com.netflix.spinnaker.echo.config;
 
 import com.netflix.spinnaker.config.OkHttpClientConfiguration;
 import com.netflix.spinnaker.echo.services.Front50Service;
+import com.netflix.spinnaker.okhttp.SpinnakerRequestInterceptor;
 import com.netflix.spinnaker.retrofit.Slf4jRetrofitLogger;
 import com.squareup.okhttp.ConnectionPool;
 import com.squareup.okhttp.OkHttpClient;
@@ -64,13 +65,17 @@ public class Front50Config {
 
   @Bean
   public Front50Service front50Service(
-      Endpoint front50Endpoint, OkHttpClient okHttpClient, LogLevel retrofitLogLevel) {
+      Endpoint front50Endpoint,
+      OkHttpClient okHttpClient,
+      LogLevel retrofitLogLevel,
+      SpinnakerRequestInterceptor spinnakerRequestInterceptor) {
     log.info("front50 service loaded");
 
     return new Builder()
         .setEndpoint(front50Endpoint)
         .setConverter(new JacksonConverter())
         .setClient(new OkClient(okHttpClient))
+        .setRequestInterceptor(spinnakerRequestInterceptor)
         .setLogLevel(retrofitLogLevel)
         .setLog(new Slf4jRetrofitLogger(Front50Service.class))
         .build()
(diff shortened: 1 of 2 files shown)


echo-web/src/main/java/com/netflix/spinnaker/echo/config/Front50Config.java
```diff
@@ -64,13 +65,17 @@ public Endpoint front50Endpoint(@Value("${front50.base-url}") String front50Base
 
   @Bean
   public Front50Service front50Service(
-      Endpoint front50Endpoint, OkHttpClient okHttpClient, LogLevel retrofitLogLevel) {
+      Endpoint front50Endpoint,
+      OkHttpClient okHttpClient,
+      LogLevel retrofitLogLevel,
+      RequestInterceptor spinnakerRequestInterceptor) {
```
plumpy: This should be `SpinnakerRequestInterceptor`... what we want is not just _any_ `RequestInterceptor`, but rather a very specific one.

plumpy: Looks like you probably copied this from `IgorConfig.java`... could you fix it there too?

deepakr199: Currently, there is only one implementation of `RequestInterceptor` which is `SpinnakerRequestInterceptor`. So this configuration will always pick up `SpinnakerRequestInterceptor`. 

But I do agree with your point of making it specific, so in future, there will be no conflict and indeed all this configuration need is `SpinnakerRequestInterceptor`. So will address your comment. Thanks for pointing it out.

plumpy:
Thanks for finding and fixing this, I really appreciate it!

plumpy:

plumpy:
Can you let me know what issues this was causing for you?

deepakr199:
> Can you let me know what issues this was causing for you?

Hi @plumpy, even we faced the same as stated in the issues [spinnaker/issues/5035](https://github.com/spinnaker/spinnaker/issues/5035) and [spinnaker/issues/4941](https://github.com/spinnaker/spinnaker/issues/4941)
Apart from that, with 1.18, there was frequent flooding of error logs on starting the pipeline since Echo to Front50 call throws 403

plumpy:
Okay, thanks for the info. We'll merge this into 1.19.3 and 1.18.7.

Thanks again for the patch!

PR #821, opened 2020-03-24
fix(notifications): remove notification type enum

claymccoy:
This takes the Notification.Type enum and makes it a simple String. This makes it possible to create a NotificationListener extension point and plugins can add their own notification types once this is no longer a closed enum. I'm not a fan of Strings, and first tried to keep this typed using extendable enums. Basically turning Notification.Type into an interface that the enums and other enums/classes could use. This got weird with classpath issues (using equals on type with the plugins), and JSON deserialization on type. I experimented with a custom deserializer for notification type and that could work, but ultimately Notification will probably end up in echo-plugins-api and the deserializer will have to live there too and that adds Jackson deps to echo-plugins-api.
Then when it got overcomplicated I realized that we aren't really making use of the typed nature of notification type anyway and there is already a precedent for simply using Strings. Event types are Strings. Even notification types are already Strings. There are basically two separate notification systems. One based on AbstractEventNotificationAgent and the other based on NotificationService. And there are redundant ways to define notification type. AbstractEventNotificationAgent.getNotificationType() are Strings. But they are always lowercase and the enums are always uppercase.
I like what was done in GooglePubsubNotificationEventListener where it inherits from both. At least everything is smashed together in one place and the notification type can be defined once. I think this is a good place to start with an EventListener extension point, but we can probably get to a better abstraction where we don't need both supportsType() and getNotificationType(). We can also probably find some commonality between sendNotifications() and handle() or at least give them more descriptive names.
NotificationService.supportsType(String type) could allow a notification service to support multiple notification types. But that isn't actually made use of yet and conflicts the usage of getNotificationType() which focuses on a single type.
I tested this with Slack and email notifications from stages, and also Slack notifications posting to the controller which hits both AbstractEventNotificationAgent and NotificationService.

### DIFF
--- a/echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/bearychat/BearychatNotificationService.groovy
+++ b/echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/bearychat/BearychatNotificationService.groovy
@@ -28,7 +28,6 @@ import org.springframework.stereotype.Component
 @Component
 @ConditionalOnProperty('bearychat.enabled')
 class BearychatNotificationService implements NotificationService {
-  private static Notification.Type TYPE = Notification.Type.BEARYCHAT
 
   @Autowired
   BearychatService bearychatService
@@ -40,8 +39,8 @@ class BearychatNotificationService implements NotificationService {
   String token
 
   @Override
-  boolean supportsType(Notification.Type type) {
-    return type == TYPE
+  boolean supportsType(String type) {
+    return "BEARYCHAT".equals(type.toUpperCase())
   }
 
   @Override
--- a/echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/email/EmailNotificationService.groovy
+++ b/echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/email/EmailNotificationService.groovy
@@ -37,7 +37,6 @@ import javax.mail.internet.MimeMessage
 @Component
 @ConditionalOnProperty('mail.enabled')
 class EmailNotificationService implements NotificationService {
-  private static Notification.Type TYPE = Notification.Type.EMAIL
 
   @Autowired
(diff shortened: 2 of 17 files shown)


echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/bearychat/BearychatNotificationService.groovy
```diff
@@ -40,8 +39,8 @@ class BearychatNotificationService implements NotificationService {
   String token
 
   @Override
-  boolean supportsType(Notification.Type type) {
-    return type == TYPE
+  boolean supportsType(String type) {
+    return "BEARYCHAT".equals(type.toUpperCase())
```
jonsie: Nit: I wonder if we should have the type be a constant member field on the class, like `private static final TYPE = "BEARYCHAT"`.  It may make this code easier to read.

claymccoy: The only reason that I didn't was because it is only used once. If there were multiple usages I'd definitely do something like that. In the future, I'd like to do something that combines supportsType and getNotificationType.

echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/email/EmailNotificationService.groovy
```diff
@@ -49,8 +48,8 @@ class EmailNotificationService implements NotificationService {
   String from
 
   @Override
-  boolean supportsType(Notification.Type type) {
-    return type == TYPE
+  boolean supportsType(String type) {
+    return "EMAIL".equals(type.toUpperCase())
```
marchello2000: i think it would at least be nice to have the string constants defined on the `Notification` class (or somewhere else)

PR #837, opened 2020-04-04
feat(triggers): add execution status as part of orca service pipeline response

fkyc:
Add execution status as part of orca service pipeline response. 

Here is the response in Orca: https://github.com/spinnaker/orca/blob/94423342e309ed3c545c9b312fd0d7597f8bbf76/orca-api/src/main/java/com/netflix/spinnaker/orca/api/pipeline/models/PipelineExecution.java#L89

Here is the ExecutionStatus in Orca:
https://github.com/spinnaker/orca/blob/94423342e309ed3c545c9b312fd0d7597f8bbf76/orca-api/src/main/java/com/netflix/spinnaker/orca/api/pipeline/models/ExecutionStatus.java

The execution status of a pipeline can be used by event handlers to determine what to do.

@robzienert

### DIFF
--- a/echo-pipelinetriggers/src/main/java/com/netflix/spinnaker/echo/pipelinetriggers/orca/OrcaExecutionStatus.java
+++ b/echo-pipelinetriggers/src/main/java/com/netflix/spinnaker/echo/pipelinetriggers/orca/OrcaExecutionStatus.java
@@ -20,7 +20,7 @@ import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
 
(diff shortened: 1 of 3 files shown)


robzienert:
This is confusing. Why not just use `ExecutionStatus`?

robzienert:
Oh derp. I'm in Echo, not Orca.

My apologies. The class should just be called `PipelineExecutionStatus`. A `Task` is different from a Pipeline.

fkyc:
@robzienert hmm should we just call it `OrcaExecutionStatus`? Seems like it's being used by a lot of task results in Orca.

PR #840, opened 2020-04-07
fix(hooks): Fix github webhooks with secrets

ferwguerra:
**Issue:**
The pipelines are no longer being triggered by Github webhooks if there is a secret required. Removing the secret allows them to function normally.

This issue was reported [here](https://github.com/spinnaker/spinnaker/issues/5645).

**Problem:**
The headings are all lowercase, so they are now not found.
This previously worked because the **org.springframework.http.HttpHeaders** class was used but was replaced by **java.util.Map**, which is case sensitive. That change was made [here](https://github.com/spinnaker/echo/commit/784909eaffd14dcf9bdd358e3792cc3160328d4d#diff-1558ece8b8cf054fbba6806681248cc1R39).

image

### DIFF
--- a/echo-pipelinetriggers/src/main/java/com/netflix/spinnaker/echo/pipelinetriggers/eventhandlers/GitEventHandler.java
+++ b/echo-pipelinetriggers/src/main/java/com/netflix/spinnaker/echo/pipelinetriggers/eventhandlers/GitEventHandler.java
@@ -43,7 +43,7 @@ import org.springframework.stereotype.Component;
 @Slf4j
 public class GitEventHandler extends BaseTriggerEventHandler<GitEvent> {
   private static final String GIT_TRIGGER_TYPE = "git";
-  private static final String GITHUB_SECURE_SIGNATURE_HEADER = "X-Hub-Signature";
+  private static final String GITHUB_SECURE_SIGNATURE_HEADER = "x-hub-signature";
   private static final List<String> supportedTriggerTypes =
       Collections.singletonList(GIT_TRIGGER_TYPE);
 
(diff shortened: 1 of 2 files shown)


gregjones:
Is it possible they're lowercase when they arrive in our setup because of the load-balancer doing normalization? Maybe it makes sense to go back to using a case-insensitive lookup, or else normalize the case when the headers are copied?

ferwguerra:
> Is it possible they're lowercase when they arrive in our setup because of the load-balancer doing normalization? Maybe it makes sense to go back to using a case-insensitive lookup, or else normalize the case when the headers are copied?

The headers are mapped using **org.springframework.http.HttpHeaders** and that's the reason they are lowercase.
You can check running echo and sending this request:

`curl --location --request POST 'http://localhost:8089/webhooks/git/github' \
--header 'User-Agent: GitHub-Hookshot/XXXXX' \
--header 'X-GitHub-Enterprise-Host: ${HOST}' \
--header 'X-GitHub-Enterprise-Version: 2.20.1' \
--header 'X-GitHub-Event: push' \
--header 'X-Hub-Signature: sha1=XXXXXXXXX' \
--header 'Content-Type: application/json' \
--data-raw '{
     ...
    }
}'`

xavileon:
@spinnaker/oss-approvers

xavileon:
@spinnaker/oss-approvers any comment on this? thanks!

ezimanyi:
@jonsie : Mind taking a look at this as it might be tangentially related to your PR #790.  Thanks!

wli66666:
any estimated time for merging this ? since this is a live bug.....

sofam:
We are also hit by this.

iiro:
We as well.

jonsie:
@ferwguerra Thanks for the contribution, and sorry for introducing this bug.  I'm proposing an alternative fix here that addresses the underlying issue:

https://github.com/spinnaker/echo/pull/863

jonsie:
Addressed in https://github.com/spinnaker/echo/pull/863

PR #857, opened 2020-04-18
feat(triggers): add trigger as part of orca service pipeline response

fkyc:
Add trigger as part of orca service pipeline response. The trigger of a pipeline execution can be used by event handlers to determine what to do.

Here is the response in Orca: https://github.com/spinnaker/orca/blob/ae2b8043e538b3ed0db39b61c8da0d2e36ee8a5a/orca-api/src/main/java/com/netflix/spinnaker/orca/api/pipeline/models/PipelineExecution.java#L105

The trigger class in Orca:
https://github.com/spinnaker/orca/blob/ae2b8043e538b3ed0db39b61c8da0d2e36ee8a5a/orca-api/src/main/java/com/netflix/spinnaker/orca/api/pipeline/models/Trigger.java

The trigger class in Echo (should be able to map what's coming in from Orca):
https://github.com/spinnaker/echo/blob/e9e72b09fc9e01ad842c1134df820516bdace721/echo-model/src/main/java/com/netflix/spinnaker/echo/model/Trigger.java

@robzienert


robzienert:
Should work!

PR #863, opened 2020-04-20
fix(headers): Metadata requestHeaders lookup should be case insensitive

jonsie:
In https://github.com/spinnaker/echo/pull/790 I removed `HttpHeaders` so that `echo-api` could be dependency free (basically, just Java) -- this is important for the ongoing extensibility / plugin work we are doing.

However, I overlooked that `HttpHeaders` builds a case insensitive map.  This resolves that issue.

Resolves: https://github.com/spinnaker/spinnaker/issues/5645
Relates to: https://github.com/spinnaker/echo/pull/840

PR #865, opened 2020-04-20
fix(headers): Metadata requestHeaders lookup should be case insensitive

jonsie:
Same as https://github.com/spinnaker/echo/pull/863 but without the tests so as not to pull in the Kotlin dependency (which makes cherry picking the change complicated).

### DIFF
--- a/echo-api/src/main/java/com/netflix/spinnaker/echo/api/events/Metadata.java
+++ b/echo-api/src/main/java/com/netflix/spinnaker/echo/api/events/Metadata.java
@@ -17,9 +17,9 @@
 package com.netflix.spinnaker.echo.api.events;
 
 import java.util.Date;
-import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.TreeMap;
 import lombok.AllArgsConstructor;
 import lombok.Data;
 import lombok.NoArgsConstructor;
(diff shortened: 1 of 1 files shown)


psalaberria002:
Was this released in 1.19.6? We are still getting the same error.

ezimanyi:
It was released in 1.19.6; I'd have to defer to @jonsie about why the error would still be happening with this fix.

psalaberria002:
I built Echo locally on top of this patch and printed the headers from the TreeMap right before the check. `x-hub-signature` is in place, but the error still fires.
```
2020-04-28 20:25:27.036  WARN 1 --- [RxIoScheduler-2] c.n.s.e.p.eventhandlers.GitEventHandler  : [accept, accept-encoding, connection, content-length, content-type, host, user-agent, x-hub-signature, x-spinnaker-anonymous, x-spinnaker-request-id, x-spinnaker-user-origin]
2020-04-28 20:25:27.036  WARN 1 --- [RxIoScheduler-2] c.n.s.e.p.eventhandlers.GitEventHandler  : Received GitEvent from Github without secure signature for trigger configured with a secret
```

jonsie:
I thought this error was accessing the property in the map.  Is GitHub expecting the header to be lower case? Is this a network error?

psalaberria002:
Github is sending it as X-Hub-Signature. There is some logic in Echo converting those to lower case, and for some reason which I cannot figure out containsKey seems to be doing a case sensitive comparison. Do you know where the headers are set?

psalaberria002:
I get a NullPointerException when calling `gitEvent.getDetails().getRequestHeaders().comparator().toString()` which suggests there is some other place in the code where the requestHeaders variable is initialized without the comparator.

jonsie:
If you're spinning your wheels on it, I think a PR to just set `private static final String GITHUB_SECURE_SIGNATURE_HEADER = "x-hub-signature";` is fine.

PR #868, opened 2020-04-21
fix: use a pipeline parameter in the email address field (#5697)

atyutyunnik:
This is for issue https://github.com/spinnaker/spinnaker/issues/5697


atyutyunnik:
discovered more issues - I would like to recall this one

PR #869, opened 2020-04-21
fix: enable use of pipeline parameter in the email address field (#5697)

atyutyunnik:
The change enables use of pipeline parameters in address/cc fields of email notifications - can be either:
 "notifications": [
    {
      "address": "${parameters.Email}",
      "level": "stage",
      "message": {
...

or a longer form:
 "notifications": [
    {
      "address": "${execution.trigger.parameters.Email}",
      "level": "stage",
      "message": {
...


atyutyunnik:
forgot to cross-reference a bug: https://github.com/spinnaker/spinnaker/issues/5697

PR #877, opened 2020-04-28
fix(headers): Lowercase X-Hub-Signature for a successful lookup

psalaberria002:
As discussed in https://github.com/spinnaker/echo/pull/865 

Lower casing the github webhook secret header key for the lookup solves this specific issue.

For the record, metadata requestHeaders lookup is still case sensitive, and I am not solving that in this PR.


jonsie:
@Mergifyio backport release-1.19.x

PR #884, opened 2020-05-05
feat(plugins): Add new plugin trigger type

robzienert:


echo-model/src/main/java/com/netflix/spinnaker/echo/model/trigger/PluginEvent.java
```diff
@@ -0,0 +1,41 @@
...
+  private Content content;
+
+  @Data
+  @JsonIgnoreProperties(ignoreUnknown = true)
+  public static class Content {
+    private String pluginId;
+    private String version;
+    private String releaseDate;
+    private List<Map<String, String>> requires;
+    private String binaryUrl;
```
jonsie, line 39: Should we add the `state` field here?

echo-model/src/main/java/com/netflix/spinnaker/echo/model/Trigger.java
```diff
@@ -157,6 +158,13 @@ public String toString() {
   // Artifact constraints
   List<String> expectedArtifactIds;
 
+  // Configuration for plugin triggers
+  String pluginId;
+  String version;
+  String releaseDate;
+  List<Map<String, String>> requires;
+  String binaryUrl;
```
jonsie, line 166: Should we add the `state` field here?

PR #885, opened 2020-05-06
fix(quietperiod): Fixup config props

marchello2000:
My change in https://github.com/spinnaker/echo/pull/784 caused issues for anyone who had the quiet period properties set as I removed the setters
This change fixes that


marchello2000:
@Mergifyio backport release-1.20.x

PR #922, opened 2020-06-02
Revert "fix: enable use of pipeline parameter in the email address field (#5697)"

marchello2000:
Reverts spinnaker/echo#869

This change did not properly address the underlying issue which was in `orca` (https://github.com/spinnaker/orca/pull/3712). Furthermore, this change in `echo` introduced a regression in the `RestListener` which prevented it from actually sending any events if there was an email notification also present on the stage/pipeline

### DIFF
--- a/echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/notification/EmailNotificationAgent.groovy
+++ b/echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/notification/EmailNotificationAgent.groovy
@@ -81,20 +81,9 @@ class EmailNotificationAgent extends AbstractEventNotificationAgent {
 
     String link = "${spinnakerUrl}/#/applications/${application}/${config.type == 'stage' ? 'executions/details' : config.link }/${event.content?.execution?.id}"
 
-    def engine = new groovy.text.SimpleTemplateEngine()
-    event.content.parameters = event.content?.execution?.trigger?.parameters
-
-    def expandEmail = { address ->
-       if (address != null && address != "") {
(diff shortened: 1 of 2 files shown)


marchello2000:
@atyutyunnik FYI.

atyutyunnik:
> @atyutyunnik FYI.
sorry about the regression. What was the issue in RestListener? I wish to analyze and learn from it. The fix did feel like a hack but a localized to email, so I felt like risk was minimal.

marchello2000:
> > @atyutyunnik FYI.
> > sorry about the regression. What was the issue in RestListener? I wish to analyze and learn from it. The fix did feel like a hack but a localized to email, so I felt like risk was minimal.

No problem! it's a convoluted system and it's hard to find the root cause. In this case I knew it because I was the one who originally introduced the problem :( 

The issue was that the Groovy `TemplateEngine` introduced an `out` key/value into the map that was of type `PrintWriter` which Jackson was unable to serialize, so we got:
Why `TemplateEngine` added it, i am not sure..

```
IllegalArgumentException: No serializer found for class java.io.PrintWriter and no properties discovered to create BeanSerializer (to avoid exception, disable SerializationFeature.FAIL_ON_EMPTY_BEANS) (through reference chain: com.netflix.spinnaker.echo.api.events.Event[\"content\"]->java.util.LinkedHashMap[\"out\"])
```

barra51:
Hi @marchello2000 ! First of all, thank for your efforts on having this fixed... I came on a scenario which I need to pass a parameter to send the email. Is there any future plan to support this?

PR #932, opened 2020-06-15
fix(notification): Support channel ID as Slack notification address

gomid:
This PR is to fix spinnaker/spinnaker#5841.

Slack API for sending notification supports channel name, channel name with `#` prefix and channel ID as its address. These 3 options can be used interchangeably.
Echo's Slack notification service doesn't support putting channel ID in the address field.

Before this PR, SlackNotificationService prefix the channel address with `#` if it's missing. If the channel ID is input by the user, it won't work properly with the `#` prefix.

After this PR, Slack notification service keep channel address as is. Users are allowed to specify channel ID.

### DIFF
--- a/echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/notification/SlackNotificationAgent.groovy
+++ b/echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/notification/SlackNotificationAgent.groovy
@@ -99,7 +99,7 @@ class SlackNotificationAgent extends AbstractEventNotificationAgent {
         .replace("{{link}}", link ?: "")
     }
 
-    String address = preference.address.startsWith('#') ? preference.address : "#${preference.address}"
+    String address = preference.address
 
     Response response
     if (sendCompactMessages) {
--- a/echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/slack/SlackNotificationService.groovy
+++ b/echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/slack/SlackNotificationService.groovy
@@ -53,7 +53,7 @@ class SlackNotificationService implements NotificationService {
     def text = notificationTemplateEngine.build(notification, NotificationTemplateEngine.Type.BODY)
     notification.to.each {
       def response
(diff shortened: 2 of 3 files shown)


evansb:
There is a typo on your commit message `notificatoin` -> `notification`

maggieneterval:
@gomid I'm likely not the best person to review this PR since I've never configured Slack with Spinnaker, but are we sure this is a backwards-compatible change? Slack's [docs](https://api.slack.com/methods/chat.postMessage#channels) seem to indicate that only when the authenticated user is a member of the group can the `#` be omitted from the name. Is this an assumption we can safely make?

gomid:
> @gomid I'm likely not the best person to review this PR since I've never configured Slack with Spinnaker, but are we sure this is a backwards-compatible change? Slack's [docs](https://api.slack.com/methods/chat.postMessage#channels) seem to indicate that only when the authenticated user is a member of the group can the `#` be omitted from the name. Is this an assumption we can safely make?

Hi @maggieneterval, thanks for your comment!
As I mentioned in the issue spinnaker/spinnaker#5841, I believe the documentation is not accurate and the `#` prefix is optional in all cases.

For a public channel, the Slack bot can send message without `#` prefix, regardless of whether it's a member of the channel or not.
Here I created a new public channel called `slack-api-test` and the bot was not a member of it. The API call was successful.
Screenshot 2020-06-16 at 9 55 23 PM

For a private channel, the Slack bot **must be a member of the group** before it can send any message. And the channel name doesn't need `#` prefix neither.
Here I created a private channel called `slack-bot-test-private`. The first API call failed because the bot is not invited to the channel, and the second one succeeded after the bot was added to the channel.
Screenshot 2020-06-16 at 9 52 54 PM

The assumption of this PR has been verified against the API and therefore the change is backward compatible.

maggieneterval:
@gomid thanks for the explanation! LGTM!

gomid:
@Mergifyio backport release-1.19.x

gomid:

gomid:
@spinnakerbot cherry-pick 1.20

PR #943, opened 2020-06-18
refactor(stats): add a simpler way to collect more data for stats

plumpy:
I made a `TelemetryEventDataProvider` interface. All registered implementations are called by the `TelemetryEventListener` to fill out the stats payload we send over to the community metrics.

I did this because @robzienert  and I are talking about adding more complicated things. I can imagine them involving RPCs and caching and it's going to be much simpler to write and test if these implementations are factored out to separate classes instead of all jammed into `TelemetryEventListener`.


echo-telemetry/src/main/java/com/netflix/spinnaker/echo/config/TelemetryConfig.kt
```diff
@@ -0,0 +1,87 @@
...
+    httpClient.setConnectTimeout(configProps.connectionTimeoutMillis.toLong(), TimeUnit.MILLISECONDS)
+    httpClient.setReadTimeout(configProps.readTimeoutMillis.toLong(), TimeUnit.MILLISECONDS)
+    return OkClient(httpClient)
+  }
+
+  @ConfigurationProperties(prefix = "stats")
+  class TelemetryConfigProps {
+
+    companion object {
+      const val DEFAULT_TELEMETRY_ENDPOINT = "https://stats.spinnaker.io/log"
```
ezimanyi, line 71: Should this be `https://stats.spinnaker.io/log` or `https://stats.spinnaker.io`?  Halyard sets it to `https://stats.spinnaker.io` without the `/log`...presumably we want to include the `/log`?  I guess that's not really related to these changes, but just checking.

echo-telemetry/src/main/java/com/netflix/spinnaker/echo/telemetry/ExecutionDataProvider.kt
```diff
@@ -0,0 +1,142 @@
...
+
+  private val objectMapper = EchoObjectMapper.getInstance()
+
+  override fun populateData(echoEvent: EchoEvent, statsEvent: StatsEvent): StatsEvent {
+    val content = objectMapper.convertValue(echoEvent.getContent(), EventContent::class.java)
+    val execution = content.execution
+
+    var executionType = getExecutionType(execution.type)
+
+    if (execution.source.type.equals("templatedPipeline", ignoreCase = true)) {
```
ezimanyi, line 43: Wow, just learned that this works in `kotlin` even if `execution.source.type` is `null`.

echo-telemetry/src/main/java/com/netflix/spinnaker/echo/telemetry/TelemetryEventListener.java
```diff
@@ -70,26 +48,27 @@
...
-    this.telemetryConfigProps = telemetryConfigProps;
     this.registry = registry;
+    this.dataProviders = dataProviders;
   }
 
   @Override
   public void processEvent(Event event) {
+
+    // These preconditions are also guaranteed to TelemetryEventDataProvider, so don't change them
+    // without checking the implementations.
```
ezimanyi, line 71: Not changed here, but while here maybe change `log.debug("Telemetry not sent: type '{}' not whitelisted ", eventType);` a few lines down to "type not configured" or something similar?

plumpy: Great catch, thanks!

echo-telemetry/src/main/java/com/netflix/spinnaker/echo/telemetry/ExecutionDataProvider.kt
plumpy: I know, it's so good!

echo-telemetry/src/main/java/com/netflix/spinnaker/echo/config/TelemetryConfig.kt
plumpy: Good catch! It looks like `/log` is appended to this URL by retrofit because we've set that as the path in [TelemetryService](https://github.com/spinnaker/echo/blob/3081a7f435b48dd75d970631bd99b112dde20ad4/echo-telemetry/src/main/java/com/netflix/spinnaker/echo/telemetry/TelemetryService.java#L25). 

Confusingly, you get a `200` response no matter what `stats.spinnaker.io` URL you send data to. So I don't know if data sent to other paths gets recorded or silently discarded... but clearly `/log` is the intended URL, so I fixed it here, thanks!

PR #947, opened 2020-06-19
feat(stats): Persist the unique id to redis

ezimanyi:
Currently the only place that the unique id for a Spinnaker instance is persisted is in the hal config. This works for users using halyard as it gets generated the first time it is needed then gets re-used on subsequent deployments. (It does have the failure mode of copying the hal config as then the new instance would have the same id as the old one.)

This doesn't work for users either deploying without Halyard or deploying via kleat. In this case there is nowhere to store this uniqueid (as kleat explicitly treats the config only as input).

To solve this problem, we'll store the unique id in redis. On startup, we'll check to see if there's a value in redis; if there is we'll use that, if not we'll use the configured value (and persist it).

### DIFF
--- a/echo-telemetry/src/main/java/com/netflix/spinnaker/echo/telemetry/InstanceIdSupplier.kt
+++ b/echo-telemetry/src/main/java/com/netflix/spinnaker/echo/telemetry/InstanceIdSupplier.kt
@@ -1,0 +1,56 @@
+/*
+ * Copyright 2020 Google, LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License")
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.spinnaker.echo.telemetry
+
+import com.netflix.spinnaker.echo.config.TelemetryConfig
+import com.netflix.spinnaker.kork.jedis.RedisClientSelector
+import com.netflix.spinnaker.kork.jedis.exception.RedisClientNotFound
+import java.lang.RuntimeException
+import org.slf4j.Logger
+import org.slf4j.LoggerFactory
+import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty
+import org.springframework.stereotype.Component
+import redis.clients.jedis.commands.JedisCommands
+
+val UNIQUE_ID_KEY = "spinnaker:stats:instanceId"
+
+@Component
+@ConditionalOnProperty("stats.enabled")
+class InstanceIdSupplier(
+  private val config: TelemetryConfig.TelemetryConfigProps,
+  redisSelector: RedisClientSelector?
+) {
+  private val log: Logger = LoggerFactory.getLogger(this.javaClass)
+  private val redis = try { redisSelector?.primary("default") } catch (e: RedisClientNotFound) { null }
+  val uniqueId = getOrSetId(config.instanceId)
+
+  /**
+   * Returns the current instance id for this Spinnaker instance. If no id is currently set,
+   * sets it to [newId] and returns [newId].
+   */
+  private fun getOrSetId(newId: String): String {
+    return try {
+      redis?.withCommandsClient<String> { c: JedisCommands ->
+        c.setnx(UNIQUE_ID_KEY, newId)
+        c.get(UNIQUE_ID_KEY)
+      }
+    } catch (e: RuntimeException) {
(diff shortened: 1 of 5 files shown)


ezimanyi:
As far as I can tell this is Kotlin code that I just wrote... 😁

echo-telemetry/src/main/java/com/netflix/spinnaker/echo/telemetry/InstanceIdSupplier.kt
```diff
@@ -0,0 +1,56 @@
...
+   * Returns the current instance id for this Spinnaker instance. If no id is currently set,
+   * sets it to [newId] and returns [newId].
+   */
+  private fun getOrSetId(newId: String): String {
+    return try {
+      redis?.withCommandsClient<String> { c: JedisCommands ->
+        c.setnx(UNIQUE_ID_KEY, newId)
+        c.get(UNIQUE_ID_KEY)
+      }
+    } catch (e: RuntimeException) {
```
plumpy, line 51: since Kotlin ignores checked exceptions (:cry:), you need to catch `Exception` here... what this does, confusingly, is catch all unchecked exceptions, but throw checked ones.

echo-telemetry/src/main/java/com/netflix/spinnaker/echo/telemetry/InstanceIdSupplier.kt
```diff
@@ -0,0 +1,56 @@
...
+
+@Component
+@ConditionalOnProperty("stats.enabled")
+class InstanceIdSupplier(
+  private val config: TelemetryConfig.TelemetryConfigProps,
+  redisSelector: RedisClientSelector?
+) {
+  private val log: Logger = LoggerFactory.getLogger(this.javaClass)
+  private val redis = try { redisSelector?.primary("default") } catch (e: RedisClientNotFound) { null }
+  val uniqueId = getOrSetId(config.instanceId)
```
plumpy, line 39: This does blocking IO during object construction, which is frowned upon. You can do this, which will run the method the first time `uniqueId` is referenced in a client:
`val uniqueId by lazy { getOrSetId(config.instanceId) }`
It's a little weird since callers have no way of knowing that accessing `uniqueId` might do I/O, but since it's only really used one place, I think it's probably fine.

Alternately, you could use a Spring `@PostConstruct` annotation, but then `uniqueId` has to be a `var String?` instead of `val String`, which is annoying...

echo-telemetry/src/test/kotlin/com/netflix/spinnaker/echo/telemetry/InstanceIdSupplierTest.kt
```diff
@@ -0,0 +1,73 @@
...
+
+import com.netflix.spinnaker.echo.config.TelemetryConfig.TelemetryConfigProps
+import com.netflix.spinnaker.kork.jedis.EmbeddedRedis
+import com.netflix.spinnaker.kork.jedis.JedisClientDelegate
+import com.netflix.spinnaker.kork.jedis.RedisClientSelector
+import org.junit.jupiter.api.Test
+import strikt.api.expectThat
+import strikt.assertions.isEqualTo
+
+class InstanceIdSupplierTest {
```
plumpy, line 27: Is there a way to mock some part of Redis so that you can have `c.setnx` throw an exception?

echo-telemetry/src/main/java/com/netflix/spinnaker/echo/telemetry/InstanceIdSupplier.kt
ezimanyi: Oh no...thanks for pointing this out!  Will fix it!

echo-telemetry/src/main/java/com/netflix/spinnaker/echo/telemetry/InstanceIdSupplier.kt
ezimanyi: I had this as `lazy` initially, then thought maybe that solution was overly-complex so reverted back to regular initialization.  I'll replace it with `lazy`!

echo-telemetry/src/test/kotlin/com/netflix/spinnaker/echo/telemetry/InstanceIdSupplierTest.kt
ezimanyi: I started looking into that, and then felt like I was basically testing the implementation of our redis autowiring (and that the test would mostly catch any changes to how we do that). But given my catch statement was incorrect and I didn't catch it, let me see if I can come up with a reasonable test.

plumpy: I think there are other reasons it could throw an exception like a network issue or something. But yeah, I dunno, if it's a ton of work, it's not really that big of a deal.

ezimanyi: I was able to reasonably mock a `JedisException` so now have a test that ensures we catch these appropriately.

PR #948, opened 2020-06-19
feat(stats): Enable stats by default in echo

ezimanyi:
The rollout plan for stats involved having the stats become opt-out (ie enabled by default) as of Spinnaker 1.19. While this was implemented, it was only done at the Halyard level; Halyard now explicitly sends 'stats.enabled = true' by default (and sends 'stats.enabled = false' for users who have disabled stats collection).

This is confusing, as stats collection is now opt-out for Halyard users but still opt-in for anyone writing their own config. As part of the effort to have services own and default their own config, update echo to be consistent with Halyard (and with the intent of the rollout plan).

With this change, echo will send usage statistics for anyone who does not have a 'stats.enabled' field in their config. Halyard users should all have this field already set to either true/false, so this would primarily affect users either writing their own config or upcoming users of kleat.

### DIFF
--- a/echo-telemetry/src/main/java/com/netflix/spinnaker/echo/config/TelemetryConfig.kt
+++ b/echo-telemetry/src/main/java/com/netflix/spinnaker/echo/config/TelemetryConfig.kt
@@ -33,7 +33,7 @@ import retrofit.client.OkClient
 import retrofit.converter.JacksonConverter
 
 @Configuration
-@ConditionalOnProperty("stats.enabled")
+@ConditionalOnProperty(value = ["stats.enabled"], matchIfMissing = true)
 @EnableConfigurationProperties(TelemetryConfigProps::class)
 open class TelemetryConfig {
 
--- a/echo-telemetry/src/main/java/com/netflix/spinnaker/echo/telemetry/ExecutionDataProvider.kt
+++ b/echo-telemetry/src/main/java/com/netflix/spinnaker/echo/telemetry/ExecutionDataProvider.kt
@@ -29,7 +29,7 @@ import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty
 import org.springframework.stereotype.Component
 
 @Component
-@ConditionalOnProperty("stats.enabled")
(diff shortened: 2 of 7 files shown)


ezimanyi:
For reference:
* The RFC on stats collection: spinnaker/spinnaker#4738
* The issue tracking making this opt-out for 1.19: spinnaker/spinnaker#5432
* The Halyard PR implementing it: spinnaker/halyard#1565

As noted above, this simply moves the implementation from Halyard to Echo, matching what I believe was the intent.  I believe the intent was to enable stats by default (with an opt-out) for all users rather than to scope that to only Halyard users which is what was actually implemented and which this fixes.

plumpy:
Can't you just change `TelemetryConfigProps.enabled` to default to `true`?

ezimanyi:
@plumpy : I don't think that `TelemetryConfigProps.enabled` would even get read in that case, as we're gating the loading of all of the configuration classes with `@ConditionalOnProperty("stats.enabled")` which I believe is just looking at the raw values in the config (not any fields we happen to be mapping them to in Java).

The other alternative (maybe better) is to add this to the default `echo.yml` so there's always a default and we don't need to add this defaulting at the top of every conditional class:
```
stats:
  enabled: true
```

I think a pattern we use in other microservices that may be better is that we don't by default scan every class under `com.netflix.spinnaker.echo` for components; instead we only scan `com.netflix.spinnaker.config`.  Then we would only have needed to gate `TelemetryConfig` with this, and have it add `com.netflix.spinnaker.echo.telemetry` to the classpath scanning.  But echo for some reason defaults to scanning all of `com.netflix.spinnaker.echo` so we also need to gate these classes.

ezimanyi:
Yeah, I confirmed that flipping that value has no effect because we only actually read it if the `@ConditionalOnProperty` at the top of the class passes.

cfieber:
I'm definitely in favor of `matchIfMissing` vs updating echo.yml to define what the expected default should be.

We don't pick up any of the yml service config files when we pull in these modules as libraries.

PR #972, opened 2020-07-09
feat(plugins): Add pluginEventType to PluginEvent

christopherthielen:


echo-model/src/main/java/com/netflix/spinnaker/echo/model/Trigger.java
```diff
@@ -160,6 +160,7 @@ public String toString() {
   List<String> expectedArtifactIds;
 
   // Configuration for plugin triggers
+  String pluginEventType;
```
christopherthielen, line 163: it's kind of awkward how `pluginEventType` is at the same level (the top) as all the plugin release fields in the Trigger model

echo-pipelinetriggers/src/main/java/com/netflix/spinnaker/echo/pipelinetriggers/eventhandlers/PluginEventHandler.java
```diff
@@ -49,19 +50,26 @@ public PluginEventHandler(
...
-            .version(event.getContent().getVersion())
-            .releaseDate(event.getContent().getReleaseDate())
-            .requires(event.getContent().getRequires())
-            .parsedRequires(event.getContent().getParsedRequires())
-            .binaryUrl(event.getContent().getBinaryUrl())
-            .sha512sum(event.getContent().getSha512sum())
-            .build();
+    return trigger -> {
+      String pluginEventType =
+          Optional.ofNullable(event.getDetails().getAttributes())
```
christopherthielen, line 55: This could also be moved to `event.content.pluginEventType`

echo-model/src/main/java/com/netflix/spinnaker/echo/model/Trigger.java
robzienert: The trigger model is just awful. I think it makes sense to follow convention for now and we can tease `Trigger` into something else another time.

PR #974, opened 2020-07-09
feat(plugins): Support filtering plugin triggers based on the pluginEventType field

christopherthielen:
Defaults both the Trigger Config's `pluginEventType` and event's `details.attributes.pluginEventType` to `PUBLISHED` if they don't exist


echo-pipelinetriggers/src/main/java/com/netflix/spinnaker/echo/pipelinetriggers/eventhandlers/PluginEventHandler.java
```diff
@@ -32,6 +32,7 @@
 @Component
 public class PluginEventHandler extends BaseTriggerEventHandler<PluginEvent> {
 
+  private static final String DEFAULT_TRIGGER_EVENT_TYPE = "PUBLISHED";
```
christopherthielen, line 35: I guess I should rename this to be consistent.

PR #981, opened 2020-07-16
feat(notifications): adds dynamic extension notification parameters

danielpeach:
Adds the ability to define notification parameters via the `NotificationAgent` extension point. The idea is that Deck will generate form fields from these parameters. 

I have corresponding Gate and Deck PRs that I'll open if this one looks OK.


echo-api/src/main/java/com/netflix/spinnaker/echo/api/events/NotificationParameter.java
```diff
@@ -0,0 +1,46 @@
...
+package com.netflix.spinnaker.echo.api.events;
+
+import lombok.Data;
+import lombok.NoArgsConstructor;
+import lombok.NonNull;
+
+/** Definition of a notification parameter that a user can configure. */
+@Data
+@NoArgsConstructor
+public class NotificationParameter {
```
danielpeach, line 26: I stole this from Orca's preconfigured job stage parameters. Maybe someday if the abstraction is right this model could live in Kork and be shared across services that want Deck to generate form fields.

echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/controller/NotificationController.java
```diff
@@ -0,0 +1,111 @@
...
+      method = RequestMethod.POST,
+      consumes = MediaType.APPLICATION_FORM_URLENCODED_VALUE,
+      produces = MediaType.APPLICATION_JSON_VALUE)
+  public ResponseEntity<String> processCallback(
+      @PathVariable String source, RequestEntity<String> request) {
+    return interactiveNotificationCallbackHandler.processCallback(source, request);
+  }
+
+  @GetMapping("/metadata")
+  public List<NotificationAgent> getNotificationTypeMetadata() {
```
danielpeach, line 108: This handler should be the only difference between the old Groovy controller and this file.

jonsie: Thanks for converting this.

PR #987, opened 2020-08-03
fix(trigger): handle quartz errors better

marchello2000:
We want to know when `echo` fails to register a trigger with quartz but only if it's not a user error (e.g. bad cron express, cron expression in the past, etc).
Quartz, unfortunately, often throws a non-descript `SchedulerException` which doesn't provide machine identifiable way to decipher if this is a user error or a system error.

This change replicates some logic that's in [quartz](https://github.com/quartz-scheduler/quartz/blob/939a9a25fbf3b374608de59e413405aebe10c089/quartz-core/src/main/java/org/quartz/core/QuartzScheduler.java#L882-L896)
to decide if the trigger is valid, if not we will log a warning, but won't increment any failure metrics.
If the are exceptions/errors outside of the trigger validation path, those will still log errors and increment failure metrics

### DIFF
--- a/echo-scheduler/src/main/groovy/com/netflix/spinnaker/echo/scheduler/actions/pipeline/PipelineConfigsPollingJob.groovy
+++ b/echo-scheduler/src/main/groovy/com/netflix/spinnaker/echo/scheduler/actions/pipeline/PipelineConfigsPollingJob.groovy
@@ -19,7 +19,6 @@ package com.netflix.spinnaker.echo.scheduler.actions.pipeline
 import com.netflix.spectator.api.Registry
 import com.netflix.spinnaker.echo.model.Trigger
 import com.netflix.spinnaker.echo.pipelinetriggers.PipelineCache
-import com.netflix.spinnaker.kork.exceptions.ConfigurationException
 import groovy.util.logging.Slf4j
 import org.quartz.CronTrigger
 import org.quartz.DisallowConcurrentExecution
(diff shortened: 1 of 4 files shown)


robzienert:
Tests would be good.

marchello2000:
> Tests would be good.

yes, agree! I will try to add some - i got a bit overwhelmed looking at it this morning (re: tests) but I agree and will make some time for it later today

echo-scheduler/src/main/groovy/com/netflix/spinnaker/echo/scheduler/actions/pipeline/PipelineConfigsPollingJob.groovy
```diff
@@ -169,24 +176,69 @@ class PipelineConfigsPollingJob implements Job {
...
+      return false
+    }
+
+    // It's possible the user has created a trigger that will never fire (e.g. a trigger in the past)
+    // That's ok, just don't even bother creating it
+    try {
+      boolean willTriggerFire = willTriggerFire(trigger)
+
+      if (!willTriggerFire) {
+        throw new SchedulerException("Trigger is in the past")
```
ajordens, line 205: Going to get a noisy stacktrace when this fires ... perhaps that's ok ... :shrug:

marchello2000: yes, at least it's not going generate a failure metric.. this code (even though i wrote it... is giving me a headache...)
for a while, i wanted to add some logic that would not spit out an error for the same trigger over and over again (since we sync every ~60s) maybe one day..

PR #1023, opened 2020-09-26
fix(provider/aws): AWS credentials at pubsub.amazon, issue #6061

atyutyunnik:
Credentials for AWS pubsub can now be specified like this:
```pubsub:
  enabled: true
  amazon:
    enabled: true
    accessKeyId: AXXXXXXXXXXXXXXXXX
    secretAccessKey: 9xxxxxxxxxxxxxxxxxxxxxxxxxxxx
```

Verified with my Spinnaker setup outside of AWS. I don't see how I can provide automated unit- and end-to-end testing without access credentials.


atyutyunnik:
This is for https://github.com/spinnaker/spinnaker/issues/6061

robzienert (commented):
Strongly recommend against using static AWS credentials - what purpose is this for?

atyutyunnik:
> Strongly recommend against using static AWS credentials - what purpose is this for?

Our Spinnaker runs in a local cluster on prem, outside of AWS environment. What's the alternative? Use of static AWS credentials (accessKeyId/secretAccessKey) in Spinnaker is quite a norm, I could tell:
```deploymentConfigurations.providers.aws.accessKeyId/secretAccessKey
deploymentConfigurations.persistentStorage.s3.accessKeyId/secretAccessKey
deploymentConfigurations.ci.codebuild.accessKeyId/secretAccessKey
deploymentConfigurations.artifacts.s3.accounts.*.accessKeyId/secretAccessKey

```

cfieber:
see https://github.com/spinnaker/spinnaker/issues/6061#issuecomment-713740678

If you are using something that is part of the Spinnaker project to deploy Spinnaker we would welcome a PR there that allows specifying AWS creds and helps in dropping them into the right spot in the deploy manifest

PR #1025, opened 2020-09-29
feat(slack): flag to use interactive bot

emjburns:
A flag (defaulting to false) to use the new slack bot even if you have no interactive actions.

This defaults to the old behavior if you don't specify the flag in your notification payload.

This PR also adds support for the subject line of a slack message. Previously we were using "Spinnaker Notification" which is both redundant (because of the name of the bot) and not useful (because we are sending a subject in the slack notification). 
This shows previous (default subject) and new (a subject that I sent)
Screen Shot 2020-09-29 at 1 05 16 PM

### DIFF
--- a/echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/api/Notification.groovy
+++ b/echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/api/Notification.groovy
@@ -29,9 +29,10 @@ class Notification {
   Source source
   Map<String, Object> additionalContext = [:]
   InteractiveActions interactiveActions
+  Boolean useInteractiveBot = false
 
   boolean isInteractive() {
-    interactiveActions != null && !interactiveActions.actions.empty
+    useInteractiveBot || (interactiveActions != null && !interactiveActions.actions.empty)
   }
 
   static class Source {
--- a/echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/slack/SlackNotificationService.groovy
+++ b/echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/slack/SlackNotificationService.groovy
@@ -50,6 +50,7 @@ class SlackNotificationService implements NotificationService {
   @Override
   EchoResponse.Void handle(Notification notification) {
     log.debug("Handling Slack notification to ${notification.to}")
+    def subject = notificationTemplateEngine.build(notification, NotificationTemplateEngine.Type.SUBJECT)
     def text = notificationTemplateEngine.build(notification, NotificationTemplateEngine.Type.BODY)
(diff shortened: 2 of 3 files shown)


echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/slack/SlackNotificationService.groovy
```diff
@@ -58,7 +59,7 @@ class SlackNotificationService implements NotificationService {
         response = slack.sendCompactMessage(new CompactSlackMessage(text), address, true)
       } else {
         response = slack.sendMessage(
-          new SlackAttachment("Spinnaker Notification", text, (InteractiveActions)notification.interactiveActions),
+          new SlackAttachment(subject, text, (InteractiveActions)notification.interactiveActions),
           address, true)
```
emjburns, line 63: Now the subject here will show up!
https://github.com/spinnaker/keel/blob/master/keel-echo/src/main/kotlin/com/netflix/spinnaker/keel/echo/ManualJudgementNotifier.kt#L115

luispollo: Yes, thank you, thank you! I totally forgot about this bug!

echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/api/Notification.groovy
```diff
@@ -29,9 +29,10 @@ class Notification {
   Source source
   Map<String, Object> additionalContext = [:]
   InteractiveActions interactiveActions
+  Boolean useInteractiveBot = false
```
luispollo, line 32: Oh, good idea, @emjburns!

echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/api/Notification.groovy
```diff
@@ -29,9 +29,10 @@ class Notification {
...
 
   boolean isInteractive() {
-    interactiveActions != null && !interactiveActions.actions.empty
+    useInteractiveBot || (interactiveActions != null && !interactiveActions.actions.empty)
```
luispollo, line 35:
```suggestion
    forceInteractiveBot || (interactiveActions != null && !interactiveActions.actions.empty)
```

luispollo (approved):
Minor naming suggestion, otherwise LGTM! Thanks for doing this!

emjburns:
@luispollo I feel like if I was reading it, both are equally clear. Either you specify to use the bot, or it auto detects based on the payload. I'm going to keep it as is

PR #1049, opened 2020-10-29
fix(retrofit): cleanup extra retrofit client beans

marchello2000:
There were a bunch of beans for retrofitClient because we set `spring.main.allow-bean-definition-overriding=true` they weren't flagged but it makes things confusing so removing all but one of them.

All clients that are used for internal Spinnaker service communications are created with the clientProvider kork utility


echo-core/src/main/java/com/netflix/spinnaker/echo/config/EchoRetrofitConfig.java
```diff
@@ -26,7 +25,7 @@
 @Slf4j
 public class EchoRetrofitConfig {
   @Bean
-  public Ok3Client ok3Client(OkHttp3ClientConfiguration okHttpClientConfig) {
-    return new Ok3Client(okHttpClientConfig.create().build());
+  public Ok3Client ok3Client() {
+    return new Ok3Client();
   }
```
robzienert, line 30: Why remove `OkHttp3ClientConfiguration`?

marchello2000: because it was actually coming through with our custom SSL config which is not what we want anymore - that is all handled by the clientServiceProvider. @srekapalli and I tried to use his new `serviceClientProvider.getService` with `isSecure=false` but that was not working because as soon as `sslSocketFactory` even if it's the default factory the SSL communication was breaking (at least to slack). I will defer to @srekapalli to looking into that

srekapalli:
> There were a bunch of beans for retrofitClient because we set `spring.main.allow-bean-definition-overriding=true` they weren't flagged but it makes things confusing so removing all but one of them.
> 

Bean overriding will give some randomness on which bean is active during runtime but so far all client beans are built the same way , so we were Ok I guess.

echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/config/RetrofitConfig.groovy
```diff
@@ -1,43 +0,0 @@
...
-import retrofit.RestAdapter
-import retrofit.client.Client
-import retrofit.client.OkClient
-
-@Configuration
-@Slf4j
-@CompileStatic
-class RetrofitConfig {
-
-    @Bean
```
srekapalli, line 34: Since we removed a type of this bean and we haven't changed any places where we refer this, not sure how the wiring works now ..?

cfieber (approved):

robzienert (approved):

PR #1050, opened 2020-10-30
fix(discovery): Respect discovery status

marchello2000:
Add discovery status check for triggering, polling pipelines configs, and pausing/resuming quartz based on discovery status

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -17,6 +17,7 @@
 plugins {
   id 'io.spinnaker.project' version "$spinnakerGradleVersion" apply false
   id "org.jetbrains.kotlin.jvm" version "$kotlinVersion" apply false
+  id 'com.adarshr.test-logger' version '2.1.0'
 }
 
 subprojects {
@@ -34,6 +35,7 @@ subprojects {
     apply plugin: "java-library"
     apply plugin: "groovy"
     apply plugin: 'jacoco'
+    apply plugin: 'com.adarshr.test-logger'
 
     test {
       testLogging {
--- a/echo-core/src/main/java/com/netflix/spinnaker/echo/notification/AbstractEventNotificationAgent.java
+++ b/echo-core/src/main/java/com/netflix/spinnaker/echo/notification/AbstractEventNotificationAgent.java
@@ -164,7 +164,7 @@ public abstract class AbstractEventNotificationAgent implements EventListener {
     }
 
     List<Map<String, Object>> notifications =
-        (List<Map<String, Object>>) context.get("notifications");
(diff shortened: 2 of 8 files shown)


echo-core/src/main/java/com/netflix/spinnaker/echo/notification/AbstractEventNotificationAgent.java
```diff
@@ -164,7 +164,7 @@ public void processEvent(Event event) {
     }
 
     List<Map<String, Object>> notifications =
-        (List<Map<String, Object>>) context.get("notifications");
+        (List<Map<String, Object>>) context.getOrDefault("notifications", Collections.emptyList());
```
marchello2000, line 167: oops, this was from an exception I noticed when somehow `notifications` is inexplicably `null`...  Since it's a pretty benign change, will just leave it as part of this PR unless anyone objects

build.gradle
```diff
@@ -34,6 +35,7 @@ subprojects {
     apply plugin: "java-library"
     apply plugin: "groovy"
     apply plugin: 'jacoco'
+    apply plugin: 'com.adarshr.test-logger'
```
marchello2000, line 38: added this for nicer test gradle output

echo-core/src/main/java/com/netflix/spinnaker/echo/DiscoveryActivator.java
```diff
@@ -0,0 +1,44 @@
...
+import com.netflix.spinnaker.kork.discovery.InstanceStatus;
+import com.netflix.spinnaker.kork.discovery.RemoteStatusChangedEvent;
+import java.util.concurrent.atomic.AtomicBoolean;
+import lombok.extern.slf4j.Slf4j;
+import org.springframework.context.ApplicationListener;
+import org.springframework.stereotype.Component;
+
+@Slf4j
+@Component
+public class DiscoveryActivator implements ApplicationListener<RemoteStatusChangedEvent> {
```
cfieber, line 27: you should be able to delete this and use the one from kork-core:

https://github.com/spinnaker/kork/blob/master/kork-core/src/main/java/com/netflix/spinnaker/kork/discovery/DiscoveryStatusListener.java#L24

cfieber (approved):
LGTM aside from the one class that is duplicated

echo-core/src/main/java/com/netflix/spinnaker/echo/DiscoveryActivator.java
marchello2000: Nice, thank you! removing

PR #1064, opened 2020-12-14
fix(bitbucket): prevent NPE when handling bitbucket payload

danielpeach:
prevent NPE when handling bitbucket payload

### DIFF
--- a/echo-webhooks/src/main/groovy/com/netflix/spinnaker/echo/scm/BitbucketWebhookEventHandler.java
+++ b/echo-webhooks/src/main/groovy/com/netflix/spinnaker/echo/scm/BitbucketWebhookEventHandler.java
@@ -30,11 +30,11 @@ import org.springframework.stereotype.Component;
 
 @Component
 @Slf4j
-public class BitbucketWehbookEventHandler implements GitWebhookHandler {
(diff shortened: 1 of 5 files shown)


echo-webhooks/src/main/groovy/com/netflix/spinnaker/echo/scm/BitbucketWebhookEventHandler.java
```diff
@@ -244,7 +244,7 @@ private void handleBitbucketServerEvent(Event event, Map postedEvent) {
   }
 
   private String emptyOrDefault(String test, String def) {
-    return !test.isEmpty() ? test : def;
+    return (test != null && !test.isEmpty()) ? test : def;
```
danielpeach, line 247: Test fails without this change.

danielpeach: (On line 162 `bitbucketCloudEvent.repository.owner.username` can be null.)

link108 (approved):

danielpeach:
@Mergifyio backport release-1.24.x

PR #1067, opened 2020-12-23
feat(triggers): Support shared secret for git triggers

gavinbunney:
This adds a `trigger.git.shared-secret` config option to allow security of git triggers using a shared secret, rather than per-pipeline. This is needed to allow for a centralized github webhook trigger setup (secret rendered server-side in echo config), which in our org GitHub is restricted to owners only, so only a select group of people have access to what the shared secret should be set to.


gavinbunney:
@spinnaker/reviewers

gavinbunney:
Hi @robzienert, could you take a look at this PR? Thanks!

blaquezzz (approved):

robzienert (approved):

PR #1071, opened 2021-01-16
chore(pubsub): don't allow overriding of bean

emjburns:
A while ago I added the ability to override the pubsub event handler because I was doing custom stuff. I'm removing this capability because the way we ended up implementing things means its fine for the handlers to run side by side.

Also making a small change to the pubsub handler to read the list of supported trigger types from the already defined function.


cfieber (approved):
LGTM

PR #1074, opened 2021-01-20
fix(notifications): Update github status notification endpoint default to not be quoted

sgarlick987:
Fixes https://github.com/spinnaker/spinnaker/issues/6304 by removing the single quotes from the default value and updates tests.

I'm not sure if there is another way other than the 2 test classes to test for the config property being set and not, but if there is better way can update.


sgarlick987:
I'll update those commit messages and repush.

jasonmcintosh:
@sgarlick987 Please fix the PR title too - should be something like `fix(notifications): Update github endpoint to ..`

dogonthehorizon (approved):

anthonywoo:
Any updates on this or documentation on how to set the `github-status.endpoint` to get around this bug? I would love to use the github status notification in my pipelines. Thanks

PR #1078, opened 2021-01-29
feat(constraint): support jsonpath expression in payload constraint key

edgarulg:
This changes is about support [JsonPath](https://github.com/json-path/JsonPath) expressions in the payload constraints for Webhook triggers.

The initial issue opened for request this is here: https://github.com/spinnaker/spinnaker/issues/3643

This feature enable users to include a JsonPath expression in the key of a payload constraint, so it first check if the key is present in first level json key (as it currently works). If the key is not present in the first level json, it will try to search the value in the payload using the key as a jsonpath expression then if exists it try to match with the regex provided.

Payload Constraint examples:
  **Key** | **Value**  
foo   :   bar
foo   :   *ar
$.foo.bar  :  *ar

The json expression must be a valid jsonpath expression and it must point to a value that can be cast to String. (String, Number, Boolean, etc). A List<String> is a valid value, and the regex will be applied to every string on the list.

The constraint will return false when the expression points to an Object, List<Objects> or the field does not exist.

### DIFF
--- a/echo-pipelinetriggers/src/main/java/com/netflix/spinnaker/echo/pipelinetriggers/artifacts/ArtifactMatcher.java
+++ b/echo-pipelinetriggers/src/main/java/com/netflix/spinnaker/echo/pipelinetriggers/artifacts/ArtifactMatcher.java
@@ -112,7 +112,7 @@ public class ArtifactMatcher {
 
     for (Object key : constraints.keySet()) {
       if (!payload.containsKey(key) || payload.get(key) == null) {
-        log.info("key not present in payload, needs to check with jsonpath");
+        log.debug("key not present in payload, needs to check with jsonpath");
         List<String> values = getValueUsingJsonPath(documentContext, key.toString());
         if (values != null && anyMatch(constraints.get(key).toString(), values)) {
           continue;
--- a/echo-pipelinetriggers/src/main/java/com/netflix/spinnaker/echo/pipelinetriggers/artifacts/ArtifactMatcher.java
+++ b/echo-pipelinetriggers/src/main/java/com/netflix/spinnaker/echo/pipelinetriggers/artifacts/ArtifactMatcher.java
@@ -16,6 +16,8 @@
 
 package com.netflix.spinnaker.echo.pipelinetriggers.artifacts;
 
+import com.google.gson.Gson;
+import com.jayway.jsonpath.*;
 import com.netflix.spinnaker.echo.model.Trigger;
 import com.netflix.spinnaker.kork.artifacts.model.Artifact;
 import com.netflix.spinnaker.kork.artifacts.model.ExpectedArtifact;
@@ -27,6 +29,11 @@ import lombok.extern.slf4j.Slf4j;
 
 @Slf4j
 public class ArtifactMatcher {
+
+  private static final Gson gson = new Gson();
+  private static final Configuration conf =
+      Configuration.defaultConfiguration().setOptions(Option.SUPPRESS_EXCEPTIONS);
+
   public static boolean anyArtifactsMatchExpected(
       List<Artifact> messageArtifacts,
       Trigger trigger,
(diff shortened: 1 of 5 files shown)


echo-pipelinetriggers/src/main/java/com/netflix/spinnaker/echo/pipelinetriggers/artifacts/ArtifactMatcher.java
```diff
@@ -86,6 +93,55 @@ public static boolean isConstraintInPayload(final Map constraints, final Map pay
...
+    for (Object key : constraints.keySet()) {
+      if (!payload.containsKey(key) || payload.get(key) == null) {
+        String value;
+        try {
+          value = documentContext.read(key.toString(), String.class);
+        } catch (InvalidPathException e) {
+          return false;
+        }
+        if (value != null) {
+          return matches(constraints.get(key).toString(), value);
```
danielpeach, line 122: What happens if you have two or more jsonpath constraints? It looks like you'll return `true` early here without checking the second constraint.

echo-pipelinetriggers/src/test/groovy/com/netflix/spinnaker/echo/pipelinetriggers/artifacts/ArtifactMatcherSpec.groovy
```diff
@@ -40,78 +40,255 @@ class ArtifactMatcherSpec extends Specification {
...
+        "ref": null,
+        "isActive": false
+      ]
+    ]
+  ]
+
   def "matches when constraint is partial word"() {
     when:
-    boolean result = ArtifactMatcher.isConstraintInPayload(shortConstraint, matchPayload)
+    boolean resultOld = ArtifactMatcher.isConstraintInPayload(shortConstraint, matchPayload)
```
danielpeach, line 136: Is `isConstraintInPayload` being used anywhere else? I don't fully understand why these tests are being combined.

echo-pipelinetriggers/src/main/java/com/netflix/spinnaker/echo/pipelinetriggers/artifacts/ArtifactMatcher.java
```diff
@@ -86,6 +93,55 @@ public static boolean isConstraintInPayload(final Map constraints, final Map pay
...
+  public static boolean isJsonPathConstraintInPayload(final Map constraints, final Map payload) {
+    String json = gson.toJson(payload);
+    DocumentContext documentContext = JsonPath.using(conf).parse(json);
+
+    for (Object key : constraints.keySet()) {
+      if (!payload.containsKey(key) || payload.get(key) == null) {
+        String value;
+        try {
+          value = documentContext.read(key.toString(), String.class);
+        } catch (InvalidPathException e) {
```
danielpeach, line 118: Is this exception thrown when the provided jsonpath is invalid? It might be nice to add validation at save time to ensure that this doesn't actually happen in practice. It also might be nice to at least log the error.

echo-pipelinetriggers/src/main/java/com/netflix/spinnaker/echo/pipelinetriggers/artifacts/ArtifactMatcher.java
```diff
@@ -86,6 +93,55 @@ public static boolean isConstraintInPayload(final Map constraints, final Map pay
...
+          return matches(constraints.get(key).toString(), value);
+        } else {
+          List<String> values;
+          try {
+            values = documentContext.read(key.toString());
+          } catch (ClassCastException e) {
+            return false;
+          }
+          if (values != null) {
+            return values.stream().anyMatch(v -> matches(constraints.get(key).toString(), v));
```
danielpeach, line 131: This is a little unexpected to me but I can't exactly put my finger on why it's strange. I'm not sure that this is what I would expect if I wrote jsonpath that results in an array...

echo-pipelinetriggers/src/main/java/com/netflix/spinnaker/echo/pipelinetriggers/artifacts/ArtifactMatcher.java
danielpeach: This is my only real concern. If what I've written isn't true I would still add a test that demonstrates that it works with multiple constraints.

edgarulg: thanks for taking a look, this does not work when multiple constraints are evaluated. I pushed the changes that will support multiple constraints and a test to demonstrate it.

echo-pipelinetriggers/src/test/groovy/com/netflix/spinnaker/echo/pipelinetriggers/artifacts/ArtifactMatcherSpec.groovy
edgarulg: Yes, `isConstraintInPayload` is being used in `PubSubHandler`. For this change, I'm only adding the JSONPath feat to the `WebhookHandler`. 

I combined these tests because the `isJsonPathConstraintInPayload` should have the same behavior that `isConstraintInPayload` for backward compatibility.

echo-pipelinetriggers/src/main/java/com/netflix/spinnaker/echo/pipelinetriggers/artifacts/ArtifactMatcher.java
edgarulg: Yes, it is for invalida jsonpath, if the exception is thrown it logs the error. The validation would be nice but I think it should be in a different PR.

echo-pipelinetriggers/src/main/java/com/netflix/spinnaker/echo/pipelinetriggers/artifacts/ArtifactMatcher.java
edgarulg: I changed the implementation for this, but basically if any string in the list match the regex it is considered a valid match.

echo-pipelinetriggers/src/test/groovy/com/netflix/spinnaker/echo/pipelinetriggers/artifacts/ArtifactMatcherSpec.groovy
danielpeach: Got it. I might make the variables more descriptive than `resultOld` and `resultNew` (maybe just `isConstraintInPayload` and `isJsonPathConstraintInPayload`). The other thing to consider is just duplicating the tests entirely.

danielpeach (approved):

PR #1079, opened 2021-02-03
fix(notifications): Support new Microsoft Teams webhook URLs

jeffreycoe:
This PR adds support for new webhook URLs for Microsoft Teams notifications.

See issue https://github.com/spinnaker/spinnaker/issues/6324 for more details.

### DIFF
--- a/echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/microsoftteams/MicrosoftTeamsNotificationService.java
+++ b/echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/microsoftteams/MicrosoftTeamsNotificationService.java
@@ -91,12 +91,9 @@ class MicrosoftTeamsNotificationService implements NotificationService {
 
     for (String webhookUrl : notification.getTo()) {
       log.info("Sending Microsoft Teams event notification");
-      log.debug("Teams Webhook Url: " + webhookUrl);
+      log.debug("Teams Webhook URL: " + webhookUrl);
 
-      String baseUrl = "https://outlook.office.com/webhook/";
-      String completeLink = webhookUrl;
-      String partialWebhookURL = completeLink.substring(baseUrl.length());
-      teamsService.sendMessage(partialWebhookURL, teamsMessage);
+      teamsService.sendMessage(webhookUrl, teamsMessage);
     }
 
     return new EchoResponse.Void();
--- a/echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/microsoftteams/MicrosoftTeamsNotificationService.java
+++ b/echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/microsoftteams/MicrosoftTeamsNotificationService.java
@@ -91,7 +91,7 @@ class MicrosoftTeamsNotificationService implements NotificationService {
 
     for (String webhookUrl : notification.getTo()) {
       log.info("Sending Microsoft Teams event notification");
-      log.debug("Teams Webhook URL: " + webhookUrl);
+      log.debug("Teams Webhook URL: {}", webhookUrl);
 
       teamsService.sendMessage(webhookUrl, teamsMessage);
     }
--- a/echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/microsoftteams/MicrosoftTeamsService.java
+++ b/echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/microsoftteams/MicrosoftTeamsService.java
@@ -38,14 +38,6 @@ public class MicrosoftTeamsService {
   }
 
   public Response sendMessage(String webhookUrl, MicrosoftTeamsMessage message) {
-    // As of Jan 2021, Microsoft has migrated to a new webhook URL where the hostname can be
-    // dynamic.
-    //
-    // In order to support both old and new URLs, and since retrofit 1.x does not support dynamic
-    // URLs,
-    // the endpoint and relative path must be derived from the full webhook URL.
-    // The RestAdapter does not provide any methods to change the endpoint URL once it's created,
-    // so this object must be instantiated with a new endpoint URL for each message.
     MicrosoftTeamsClient microsoftTeamsClient =
         new RestAdapter.Builder()
             .setConverter(new JacksonConverter())
--- a/echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/microsoftteams/MicrosoftTeamsService.java
+++ b/echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/microsoftteams/MicrosoftTeamsService.java
@@ -38,6 +38,8 @@ public class MicrosoftTeamsService {
   }
 
   public Response sendMessage(String webhookUrl, MicrosoftTeamsMessage message) {
+    // The RestAdapter instantiation needs to occur for each message to be sent as
+    // the incoming webhook base URL and path may be different for each Teams channel
     MicrosoftTeamsClient microsoftTeamsClient =
         new RestAdapter.Builder()
             .setConverter(new JacksonConverter())
@@ -52,17 +54,13 @@ public class MicrosoftTeamsService {
   }
 
   private String getEndpointUrl(String webhookUrl) {
-    String baseUrl = "";
-
     try {
       URL url = new URL(webhookUrl);
-      baseUrl = url.getProtocol() + "://" + url.getHost();
+      return url.getProtocol() + "://" + url.getHost();
     } catch (MalformedURLException e) {
       throw new InvalidRequestException(
           "Unable to determine base URL from Microsoft Teams webhook URL.", e);
     }
-
-    return baseUrl;
   }
 
   private String getRelativePath(String webhookUrl) {
@@ -82,7 +80,7 @@ public class MicrosoftTeamsService {
 
     // Remove slash from beginning of path as the client will prefix the string with a slash
     if (relativePath.charAt(0) == '/') {
-      relativePath = relativePath.substring(1, relativePath.length());
+      relativePath = relativePath.substring(1);
     }
 
     return relativePath;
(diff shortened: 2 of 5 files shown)


echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/microsoftteams/MicrosoftTeamsService.java
```diff
@@ -16,18 +16,64 @@
...
+            .setLog(new Slf4jRetrofitLogger(MicrosoftTeamsClient.class))
+            .build()
+            .create(MicrosoftTeamsClient.class);
+
+    return microsoftTeamsClient.sendMessage(getRelativePath(webhookUrl), message);
+  }
+
+  private String getEndpointUrl(String webhookUrl) {
+    String baseUrl = "";
+    Pattern pattern = Pattern.compile(BASE_URL_REGEX_PATTERN, Pattern.CASE_INSENSITIVE);
```
dreynaud, line 56: can we factor this out and compile the pattern only once during the initialization of MicrosoftTeamsService?

echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/microsoftteams/MicrosoftTeamsNotificationService.java
```diff
@@ -91,12 +91,9 @@ public boolean supportsType(String type) {
 
     for (String webhookUrl : notification.getTo()) {
       log.info("Sending Microsoft Teams event notification");
-      log.debug("Teams Webhook Url: " + webhookUrl);
+      log.debug("Teams Webhook URL: " + webhookUrl);
```
dreynaud, line 94:
```suggestion
      log.debug("Teams Webhook URL: {}", webhookUrl);
```

echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/microsoftteams/MicrosoftTeamsService.java
```diff
@@ -16,18 +16,64 @@
...
+  private RestAdapter.LogLevel retrofitLogLevel;
+
+  public MicrosoftTeamsService(Client retrofitClient, RestAdapter.LogLevel retrofitLogLevel) {
+    this.retrofitClient = retrofitClient;
+    this.retrofitLogLevel = retrofitLogLevel;
   }
 
   public Response sendMessage(String webhookUrl, MicrosoftTeamsMessage message) {
-    return microsoftTeamsClient.sendMessage(webhookUrl, message);
+    MicrosoftTeamsClient microsoftTeamsClient =
```
dreynaud, line 41: do we _have_ to create a new client per message now? Is that because the domain may be something different from outlook.office.com?

echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/microsoftteams/MicrosoftTeamsService.java
```diff
@@ -16,18 +16,64 @@
...
 import lombok.extern.slf4j.Slf4j;
+import retrofit.RestAdapter;
+import retrofit.client.Client;
 import retrofit.client.Response;
+import retrofit.converter.JacksonConverter;
 
 @Slf4j
 public class MicrosoftTeamsService {
-  private final MicrosoftTeamsClient microsoftTeamsClient;
+  private static final String BASE_URL_REGEX_PATTERN = "^(http|https)://.+?/";
```
dreynaud, line 30: do we have to live with a regex or could we use a somewhat standard class to parse URLs?

If this is really easiest to go with, it would be nice to see some tests showing what exactly this is extracting from the regex.

echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/microsoftteams/MicrosoftTeamsService.java
```diff
@@ -16,18 +16,64 @@
...
+
+  private String getEndpointUrl(String webhookUrl) {
+    String baseUrl = "";
+    Pattern pattern = Pattern.compile(BASE_URL_REGEX_PATTERN, Pattern.CASE_INSENSITIVE);
+    Matcher matcher = pattern.matcher(webhookUrl);
+
+    if (matcher.find()) {
+      baseUrl = matcher.group(0);
+    } else {
+      log.error("Webhook URL does not match base URL format. URL: " + webhookUrl);
```
dreynaud, line 62:
```suggestion
      log.error("Webhook URL ({}) does not match base URL format.", webhookUrl);
```

echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/microsoftteams/MicrosoftTeamsService.java
```diff
@@ -16,18 +16,64 @@
...
+    }
+
+    return baseUrl;
```
dreynaud, line 65: might as well throw an exception here, no? Returning an empty string is going to create other problems down the line

echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/microsoftteams/MicrosoftTeamsService.java
dreynaud: might as well throw an exception here, no? Returning an empty string is going to create other problems down the line

echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/microsoftteams/MicrosoftTeamsNotificationService.java
jeffreycoe: Fixed.

echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/microsoftteams/MicrosoftTeamsService.java
jeffreycoe: Looking back on this, I clearly was overthinking this and have no idea why I chose to do regex and not just use `java.net.URL`. 

I removed the regex use `java.net.URL` to parse the URL segments instead.

echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/microsoftteams/MicrosoftTeamsService.java
jeffreycoe: Unfortunately, yes. The domain could be something like outlook.office.com or [tenant-name].webhook.office.com, and the webhook path is slightly different now, too. This seemed to me to be the best option since the URL could be different even at an individual Teams channel level.

echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/microsoftteams/MicrosoftTeamsService.java
jeffreycoe: Removed regex and use `java.net.URL` instead.

echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/microsoftteams/MicrosoftTeamsService.java
jeffreycoe: Refactored to use `java.net.URL` and throws exception if URL is empty

echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/microsoftteams/MicrosoftTeamsService.java
```diff
@@ -16,18 +16,75 @@
...
+        throw new MalformedURLException();
+      }
+    } catch (MalformedURLException e) {
+      throw new InvalidRequestException(
+          "Unable to determine relative path from Microsoft Teams webhook URL.", e);
+    }
+
+    // Remove slash from beginning of path as the client will prefix the string with a slash
+    if (relativePath.charAt(0) == '/') {
+      relativePath = relativePath.substring(1, relativePath.length());
```
dreynaud, line 85:
```suggestion
      relativePath = relativePath.substring(1);
```

echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/microsoftteams/MicrosoftTeamsService.java
```diff
@@ -16,18 +16,75 @@
...
+
+    try {
+      URL url = new URL(webhookUrl);
+      baseUrl = url.getProtocol() + "://" + url.getHost();
+    } catch (MalformedURLException e) {
+      throw new InvalidRequestException(
+          "Unable to determine base URL from Microsoft Teams webhook URL.", e);
+    }
+
+    return baseUrl;
```
dreynaud, line 65: nit:

```suggestion
    try {
      URL url = new URL(webhookUrl);
      return url.getProtocol() + "://" + url.getHost();
    } catch (MalformedURLException e) {
      throw new InvalidRequestException(
          "Unable to determine base URL from Microsoft Teams webhook URL.", e);
    }
```

echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/microsoftteams/MicrosoftTeamsService.java
dreynaud: I see, thanks! I'd be really happy if you could add this explanation as a comment. Otherwise LGTM 👍

jeffreycoe: No problem - I added a comment here.

echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/microsoftteams/MicrosoftTeamsService.java
jeffreycoe: Fixed.

echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/microsoftteams/MicrosoftTeamsService.java
jeffreycoe: Fixed.

dreynaud (approved):
Thanks @jeffreycoe !

PR #1082, opened 2021-02-08
feat(trigger/docker): Add docker image digest to trigger context

german-muzquiz:
Fixes https://github.com/spinnaker/spinnaker/issues/6212

This change will add the docker image `digest` to the trigger context in pipeline execution json, under the key `trigger.digest`, so that it can be referenced by pipeline expressions.

This will only be added if `trackDigests` is enabled at docker account configuration.

### DIFF
--- a/echo-pipelinetriggers/src/main/java/com/netflix/spinnaker/echo/pipelinetriggers/eventhandlers/DockerEventHandler.java
+++ b/echo-pipelinetriggers/src/main/java/com/netflix/spinnaker/echo/pipelinetriggers/eventhandlers/DockerEventHandler.java
@@ -88,8 +88,7 @@ public class DockerEventHandler extends BaseTriggerEventHandler<DockerEvent> {
   protected Function<Trigger, Trigger> buildTrigger(DockerEvent dockerEvent) {
     return trigger ->
         trigger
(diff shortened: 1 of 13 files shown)


echo-pipelinetriggers/src/main/java/com/netflix/spinnaker/echo/pipelinetriggers/eventhandlers/DockerEventHandler.java
```diff
@@ -87,7 +87,10 @@ public boolean isSuccessfulTriggerEvent(DockerEvent dockerEvent) {
   @Override
   protected Function<Trigger, Trigger> buildTrigger(DockerEvent dockerEvent) {
     return trigger ->
-        trigger.atTag(dockerEvent.getContent().getTag()).withEventId(dockerEvent.getEventId());
+        trigger
+            .atTag(dockerEvent.getContent().getTag())
```
ncknt, line 91: It's small but can we make a new `atTag(String tag, String digest)`? We're building 3 trigger objects here (tag -> digest -> event). I'd even prefer a single call but I'd settle for two in the name of consistency :)

ncknt (approved):
LGTM

PR #1090, opened 2021-03-11
feat(travis): Always read properties from Travis triggers

jervi:
Travis doesn't need the value of the propertyFile field - it just needs it to be set to some value.
Corresponds with https://github.com/spinnaker/orca/pull/4089


echo-core/src/main/java/com/netflix/spinnaker/echo/build/BuildInfoService.java
```diff
@@ -74,8 +74,12 @@ public BuildEvent getBuildEvent(String master, String job, int buildNumber) {
     String job = event.getContent().getProject().getName();
     int buildNumber = event.getBuildNumber();
 
+    if (StringUtils.isEmpty(propertyFile) && master.contains("travis")) {
+      propertyFile = "travis";
+    }
+    String propertyFileFinal = propertyFile;
     if (StringUtils.isNoneEmpty(master, job, propertyFile)) {
-      return retry(() -> igorService.getPropertyFile(buildNumber, propertyFile, master, job));
+      return retry(() -> igorService.getPropertyFile(buildNumber, propertyFileFinal, master, job));
```
ajordens, line 82: Is this needed with the `orca` change? It seems not idea to have an explicit travis check in the core `BuildInfoService`.

jervi: Yes, unfortunately. This PR is for build _triggers_, the orca one is for build _stages_. The Jenkins/Travis relationship in echo is ugly (basically there's no way to differentiate them except using the name - Travis build masters are always prefixed with `travis-`).

PR #1106, opened 2021-06-03
fix(notification): Add application to githubStatus notification context

dbrougham:
Add the application to the github status notification. This helps ensure checks are unique per service when working in a monorepo.

Issue: https://github.com/spinnaker/spinnaker/issues/6442


echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/notification/GithubNotificationAgent.java
```diff
@@ -91,7 +93,7 @@ public void sendNotifications(
               content.getStageIndex());
     } else if (config.get("type").equals("pipeline")) {
       description = String.format("Pipeline '%s' is %s", content.getPipeline(), status);
-      context = String.format("pipeline/%s", content.getPipeline());
+      context = String.format("%s/pipeline/%s", application, content.getPipeline());
```
gregjones, line 96: I think this will be quite disruptive for existing users of the feature - if the existing context has been set to required in the github branch-protection settings, then this will leave github waiting for statuses that never match the required value. I think that if the change is made at the end instead, then github will still match it, as a prefix. i.e. change it to `pipeline/$pipeline/$application`

dbrougham: Thanks for the feedback!

Yeah I did think that any repositories using this feature may be required to update their GitHub checks if they are set to required. 

I couldn't seem to find anything online in the GitHub docs that mention a partial match working for the GitHub status. Will do some testing on this today.

Alternatively we could have a user specified context that is sent although this will require more code changes to implement, but would be backwards compatible, or we could look to use a feature flag to specify the full path for GitHub notification context

dbrougham: After some testing it doesn't look like the GitHub status will behave that way
https://docs.github.com/en/rest/reference/repos#statuses

We could add this to the release notes as a potential breaking change to be aware of?

Two alternate ways forward:
1. Put this functionality behind some kind of feature flag for echo `detailedGitHubStatus`
2. Allow users to specify a custom context in the notification pipeline config/deck UI

PR #1107, opened 2021-06-04
fix(templates/googlechat): custom template

dethi:
Copied the default template, but added some HTML tag to make the message more readable and also converted the additional instructions/messages with `markdownToHtml.convert(...)` since Google Chat support a [subset of HTML tags](https://developers.google.com/chat/reference/message-formats/cards#card_text_formatting).

fix https://github.com/spinnaker/spinnaker/issues/6236

Screenshot 2021-06-04 at 20 00 46


mattgogerly:
Can you share a before screenshot too? I don't use Google Chat notifications so I don't have a reference point.

dethi:
Screenshot 2021-12-09 at 11 12 45

mattgogerly:
@Mergifyio update

mattgogerly (approved):

PR #1110, opened 2021-07-09
fix(webhooks/stash): Adding jsonalias for refChanges field of StashWe…

ChrisTheShark:
…bhookEvent to account for Stash webhook payloads that follow this format.

This patch adds a simple json alias for the refChanges field of StashWebhookEvent to account for payloads that have a slightly different format of changes for the field name. Spinnaker issue 4427 reports this behavior as a NullPointerException on line 44 or 46 of StashWebhookEvent where refChanges is null. I experienced the NullPointerException originating at line 43. This shift in code location could be explained by changes in the file over time but I don't have the historical context. I built a custom echo image with this proposed fix to validate. This fix corrected Stash triggers in my case and has no impact to existing payloads that follow the refChanges format.

fix spinnaker/spinnaker#4427

### DIFF
--- a/echo-webhooks/src/main/groovy/com/netflix/spinnaker/echo/scm/StashWebhookEventHandler.java
+++ b/echo-webhooks/src/main/groovy/com/netflix/spinnaker/echo/scm/StashWebhookEventHandler.java
@@ -16,6 +16,7 @@
 
 package com.netflix.spinnaker.echo.scm;
 
+import com.fasterxml.jackson.annotation.JsonAlias;
 import com.fasterxml.jackson.databind.ObjectMapper;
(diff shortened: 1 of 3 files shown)


KellerII:
This is a much needed fix. As of now Spinnaker manifests deployments when leveraging Stash/Bitbucket are broken.

Tamoghnasen9:
+1.

danimorelo:
+1
This bug also prevents Managed Delivery to work when triggering from stash/bitbucket server

mattgogerly:
@Mergifyio update

mattgogerly (approved):

PR #1115, opened 2021-07-28
fix(notification): log exception on failure

deverton:
This change updates the logging so that when a notification agent fails the exception is logged.


mattgogerly (approved):

dbyron-sf:
@mergifyio update

dbyron-sf (approved):

PR #1209, opened 2022-10-13
fix(pipelineTriggers): handle invalid constraint regexes

ajbw:
Currently, if a user accidentally configures a pipeline trigger with an invalid regex for a constraint value, echo throws a `PatternSyntaxException` any time it receives an event that should be tested against the constraint.  As a result, all pipeline triggers using the same trigger type & source/subscription will be non-functional, not just the pipeline with the bad regex.

This patch catches `PatternSyntaxException`, logs an error about the bad regex for troubleshooting purposes, and returns `false` such that the bad regex will never match, allowing all other pipeline triggers to continue functioning normally.

We haven't raised an issue for this as it's a fairly trivial bug & fix.

### DIFF
--- a/echo-pipelinetriggers/src/main/java/com/netflix/spinnaker/echo/pipelinetriggers/artifacts/ArtifactMatcher.java
+++ b/echo-pipelinetriggers/src/main/java/com/netflix/spinnaker/echo/pipelinetriggers/artifacts/ArtifactMatcher.java
@@ -134,7 +134,12 @@ public class ArtifactMatcher {
     try {
       p = Pattern.compile(us);
     } catch (PatternSyntaxException ex) {
-      log.error("Invalid Regex pattern for constraint, will never match any payload: " + us);
+      log.error(
+          "Invalid regex pattern for constraint, will never match any payload: \""
+              + us
+              + "\": "
+              + ex.getMessage(),
+          ex);
       return false;
     }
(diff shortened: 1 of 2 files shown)


echo-pipelinetriggers/src/main/java/com/netflix/spinnaker/echo/pipelinetriggers/artifacts/ArtifactMatcher.java
```diff
@@ -129,7 +130,14 @@ public static boolean isJsonPathConstraintInPayload(final Map constraints, final
   }
 
   private static boolean matches(String us, String other) {
-    return Pattern.compile(us).asPredicate().test(other);
+    Pattern p;
+    try {
+      p = Pattern.compile(us);
+    } catch (PatternSyntaxException ex) {
+      log.error("Invalid Regex pattern for constraint, will never match any payload: " + us);
```
dbyron-sf, line 137: I bet there's some helpful information in the exception message.  I'd include it in the log message, or maybe the entire exception, so that the stack trace and any other cause shows up.

dbyron-sf: Thanks!

dbyron-sf (approved):

PR #1213, opened 2022-10-20
fix(pipelineTriggers): change buildNumber type from int to string

jaredstehler:
in order to support build numbers which are not strict integers (i.e. "123.1")

https://github.com/spinnaker/spinnaker/issues/6743


dbyron-sf:
LGTM, but would like to make sure the [orca](https://github.com/spinnaker/orca/pull/4316) and [igor](https://github.com/spinnaker/igor/pull/1049) PRs ready and have them go in "at the same time."

echo-test/src/main/groovy/com/netflix/spinnaker/echo/test/RetrofitStubs.groovy
```diff
@@ -64,7 +64,7 @@ trait RetrofitStubs {
   }
 
   BuildEvent createBuildEventWith(BuildEvent.Result result) {
-    def build = result ? new BuildEvent.Build(result == BUILDING, 1, result, null, []) : null
+    def build = result ? new BuildEvent.Build(result == BUILDING, '1', result, null, []) : null
```
jasonmcintosh, line 67: SUPER minor, but what about adding a test for a non-numeric build number as well as a numeric build number? :P And maybe a call out/comment on why?  I like this overall though!

jasonmcintosh (approved):

jaredstehler:
can this be merged yet?

PR #1233, opened 2022-12-09
feat(event): Add circuit breaker for events sending.

armory-abedonik:
**Description**

The event may have payload size up to 10 MB and even more, there are no strict limits. The number of events to be sent can be large. It depends on how frequently Spinnaker runs pipelines, and other internal jobs. As a result, if consumer API is unavailable or has high latency then Echo service is constantly restarting due to Out of Memory error.

**How to use?**

Add to `echo-local.yml`:

```
rest:
  enabled: true
  circuitBreakerEnabled: true

resilience4j:
  circuitbreaker:
    instances:
      sendEvent:
        failure-rate-threshold: 50
        minimum-number-of-calls: 5
        automatic-transition-from-open-to-half-open-enabled: true
        wait-duration-in-open-state: 5s
        permitted-number-of-calls-in-half-open-state: 3
        sliding-window-size: 10
        sliding-window-type: count_based
```


jasonmcintosh (approved):

dogonthehorizon (approved):

dogonthehorizon:
Has feature flag and is tested, backporting

dogonthehorizon:
@mergifyio backport release-1.29.x release-1.28.x release-1.27.x

PR #1292, opened 2023-04-20
feat(pipelinetriggers): only cache enabled pipelines with enabled triggers of specific types

dbyron-sf:
when pipelineCache.filterFront50Pipelines is true.  Determine the specific types from the types that the available BaseTriggerEventHandler beans support, with cron triggers as well. Adjust the logic in ManualEventHandler to work when the cache only contains a subset of pipelines.

When pipelineCache.filterFront50Pipelines is false, retain the current behavior to query front50 for all pipelines.

Note that both before and after this change, when ManualEventHandler gets a pipeline from the cache, it queries front50 for a single pipeline by id to make sure what it's triggering is up to date.  As currently implemented, front50 queries its database by id to generate the response.

Currently, deck triggers by pipeline application + name.  See for example
https://github.com/spinnaker/deck/blob/8abafe5c6a6680953aef7752b1a56cac093ec9ba/packages/core/src/pipeline/executions/executionGroup/ExecutionGroup.tsx#L138 and
https://github.com/spinnaker/deck/blob/8abafe5c6a6680953aef7752b1a56cac093ec9ba/packages/core/src/domain/IPipeline.ts#L86.

With this PR, ManualEventHandler now queries a new endpoint in front50 by application and name when there's no cached pipeline corresponding to a trigger.  Front50's database query in this case is likely slightly more expensive that querying by id, but not enough to be significant, especially since we're dealing with human-initiated events.  If the query by application and name fails, ManualEventTrigger then queries by pipeline id, so triggers by pipeline id (if there are any) are less performant than before as they involve another round trip to front50 to query by application and name first.

It does seem feasible, though not totally trivial, to change deck to trigger manual execution by pipeline id.  If this happens, it would make sense to change the order of queries in ManualEventHandler to query by id first.  That would restore the performance of manual execution by deck what it was before this PR.

depends on https://github.com/spinnaker/front50/pull/1252

### DIFF
--- a/echo-pipelinetriggers/src/main/java/com/netflix/spinnaker/echo/pipelinetriggers/PipelineCache.java
+++ b/echo-pipelinetriggers/src/main/java/com/netflix/spinnaker/echo/pipelinetriggers/PipelineCache.java
@@ -23,6 +23,8 @@ import com.netflix.spectator.api.Registry;
 import com.netflix.spectator.api.patterns.PolledMeter;
 import com.netflix.spinnaker.echo.model.Pipeline;
 import com.netflix.spinnaker.echo.model.Trigger;
+import com.netflix.spinnaker.echo.pipelinetriggers.eventhandlers.BaseTriggerEventHandler;
+import com.netflix.spinnaker.echo.pipelinetriggers.eventhandlers.TriggerEventHandler;
 import com.netflix.spinnaker.echo.pipelinetriggers.orca.OrcaService;
 import com.netflix.spinnaker.echo.services.Front50Service;
 import com.netflix.spinnaker.security.AuthenticatedRequest;
@@ -37,6 +39,7 @@ import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
 import javax.annotation.PostConstruct;
 import javax.annotation.PreDestroy;
+import lombok.Getter;
 import lombok.NonNull;
 import lombok.extern.slf4j.Slf4j;
 import org.springframework.beans.factory.annotation.Autowired;
@@ -55,6 +58,13 @@ public class PipelineCache implements MonitoredPoller {
   private final ObjectMapper objectMapper;
   private final PipelineCacheConfigurationProperties pipelineCacheConfigurationProperties;
 
+  /**
+   * If enabled by a feature flag, query front50 for (enabled) pipelines with only these (enabled)
+   * trigger types. A comma-separated string since that's what front50 expects. Getter only for
+   * testing.
+   */
+  @Getter private final String supportedTriggerTypes;
+
   private volatile Boolean running;
   private volatile Instant lastPollTimestamp;
 
@@ -70,7 +80,8 @@ public class PipelineCache implements MonitoredPoller {
       ObjectMapper objectMapper,
       @NonNull Front50Service front50,
       @NonNull OrcaService orca,
-      @NonNull Registry registry) {
+      @NonNull Registry registry,
+      @NonNull List<BaseTriggerEventHandler> triggerHandlers) {
     this(
         Executors.newSingleThreadScheduledExecutor(),
         pollingIntervalMs,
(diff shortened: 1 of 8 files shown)


echo-pipelinetriggers/src/main/java/com/netflix/spinnaker/echo/pipelinetriggers/PipelineCache.java
```diff
@@ -232,13 +282,15 @@ private static Map<String, List<Trigger>> extractEnabledTriggersFrom(List<Pipeli
...
+  /**
+   * Look up the latest version in front50 of a (potentially stale) pipeline config from the cache
+   * if anything fails during the refresh, we fall back to returning the cached version
+   */
   public Pipeline refresh(Pipeline cached) {
     try {
       Map<String, Object> pipeline = front50.getPipeline(cached.getId());
       Optional<Pipeline> processed = process(pipeline);
-      if (!processed.isPresent()) {
+      if (processed.isEmpty()) {
```
jasonmcintosh, line 293: Thanks for the cleanup :)

echo-pipelinetriggers/src/main/java/com/netflix/spinnaker/echo/pipelinetriggers/PipelineCache.java
```diff
@@ -66,45 +76,68 @@ public class PipelineCache implements MonitoredPoller {
...
-      @Value(
-              "${pipelineCache.parallelism:#{T(java.lang.Runtime).getRuntime().availableProcessors()}}")
-          int planParallelism,
+      @NonNull PipelineCacheConfigurationProperties pipelineCacheConfigurationProperties,
       ObjectMapper objectMapper,
       @NonNull Front50Service front50,
       @NonNull OrcaService orca,
-      @NonNull Registry registry) {
+      @NonNull Registry registry,
+      @NonNull List<BaseTriggerEventHandler> triggerHandlers) {
```
jasonmcintosh, line 84: Base vs. plain TriggerEventHandler?

echo-pipelinetriggers/src/main/java/com/netflix/spinnaker/echo/pipelinetriggers/PipelineCache.java
```diff
@@ -66,45 +76,68 @@ public class PipelineCache implements MonitoredPoller {
...
       int pollingIntervalMs,
       int pollingSleepMs,
-      int planParallelism,
+      @NonNull PipelineCacheConfigurationProperties pipelineCacheConfigurationProperties,
       ObjectMapper objectMapper,
       @NonNull Front50Service front50,
       @NonNull OrcaService orca,
-      @NonNull Registry registry) {
+      @NonNull Registry registry,
+      @NonNull List<BaseTriggerEventHandler> triggerHandlers) {
```
jasonmcintosh, line 107: Should this be TriggerEventHandler vs. BaseTriggerEventHandler?

echo-pipelinetriggers/src/main/java/com/netflix/spinnaker/echo/pipelinetriggers/PipelineCache.java
dbyron-sf: Using TriggerEventHandler here would give me ManualEventHandler, where BaseTriggerEventHandler doesn't.  Since pipelines don't typically define triggers whose type is `"manual"` it made sense to leave them out of here.

jasonmcintosh: Yeah only worry is a plugin or someone else who extends non-base.  Wondering if we could add an explicit exlucsion for manual trigger types or something.  I'm not as sure on this one.... I can leave it base, just one of those things that caught my eye.

jasonmcintosh (approved):

echo-pipelinetriggers/src/main/java/com/netflix/spinnaker/echo/pipelinetriggers/PipelineCache.java
dbyron-sf: Yeah, I'm not 100% how this will play out.  With any luck folks who enable the feature flag and have trouble will get in touch.

PR #1295, opened 2023-05-15
feat(cdevents-notification): Implementing produce CDEvents using Notification

rjalander:
Implementing Produce [CDEvents](https://cdevents.dev/) as per the RFC design 
https://github.com/spinnaker/governance/blob/master/rfc/cdevents-spinnaker.md#produce-cdevents-1

This PR has the changes to implement a new CDEventsNotificationAgent, to handle the CDEvent type of Notifications and send them to `address` configured as events-broker URL using [CDEvents-Java-SDK](https://github.com/cdevents/sdk-java)

**Note:**
Need some clarifications on type of CDEvents that Spinnaker needs to send,
1. Is this something will allow Spinnaker to send all type of CDEvents [core, source code, CI, CD and Operations Events](https://cdevents.dev/docs/) as a Notification
2. Or do we need to restrict to a subset of CDEvents(core, CD,..) which is suitable for Spinnaker

Current PR has the implementation to produce [Core CDEvents](https://github.com/cdevents/spec/blob/v0.1.2/core.md) for now, other events will be implemented after this PR review and discussion.

Dependent Spinnaker/deck PR - https://github.com/spinnaker/deck/pull/9997


echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/cdevents/CDEventsSenderService.java
```diff
@@ -0,0 +1,82 @@
...
+   * Sends CDEvent to the Configured ClouEvent broker URL.
+   *
+   * @param ceToSend
+   * @param url
+   * @return HttpURLConnection
+   * @throws IOException
+   */
+  public HttpURLConnection sendCDEvent(CloudEvent ceToSend, URL url) throws IOException {
+    HttpURLConnection httpUrlConnection = createConnection(url);
+    MessageWriter messageWriter = createMessageWriter(httpUrlConnection);
```
jasonmcintosh, line 41: A concern here is this is REALLY raw java layer for HTTP communications.  The standard in spinnaker is to use retrofit & ok-http client libs.  These automatically add things LIKE timeouts, URL validation, etc. that are pretty key settings for most environments.  I've not looked at the CloudEvents MessageWriter system... so may HAVE to do this, but definitely have concerns.

echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/notification/CDEventsNotificationAgent.java
```diff
@@ -0,0 +1,87 @@
...
+            "Received response from events broker : {} {} for execution id {}. {}",
+            response.getResponseCode(),
+            response.getResponseMessage(),
+            executionId,
+            new String(((TypedByteArray) response.getContent()).getBytes()));
+      }
+
+    } catch (Exception e) {
+      log.error("Exception occurred while sending CDEvent {}", e);
+      throw new CDEventsException("Exception occurred while sending CDEvent", e);
```
jasonmcintosh, line 84: As a rule, we should be throwing spinnaker exceptions not custom exceptions for these.

xibz: One thing to call out here, is unless you are modifying the retrofit exception handler, these custom exceptions may not be handled properly (I dont think). One thing we could consider doing is having some base spinnaker exception that custom exceptions could inherit from which could make throwing custom exceptions a little more flexible

xibz (changes requested):

echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/cdevents/CDEventsSenderService.java
rjalander: I am trying to use Retrofit client to send CDEvent but facing difficulties in deserializing as Retrofit doesn't directly use `CloudEventHttpMessageConverter` is the issue.
Can I use spring RestTemplate to send the CDEvent instead, please share your ideas?

dbyron-sf: With retrofit and a return type of Response (e.g. [here](https://github.com/spinnaker/echo/blob/31779124727a5eda13224a3ba9082c2c215996af/echo-core/src/main/java/com/netflix/spinnaker/echo/services/IgorService.java#L58)), I believe you'll get flexibility on converting the response.

rjalander: I could able to solve this issue by creating a custom [CDEventsHTTPMessageConverter](https://github.com/spinnaker/echo/pull/1295/files#diff-03fe23e811c27266271bbe77d1b5879778999b8aabf512271b53cf7c5430341e) as retrofit doesn't directly support CloudEventHttpMessageConverter, this converter uses Jackson converter to map the CloudEvent as String and send this String with the Requestbody

echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/cdevents/CDEventsBuilderService.java
```diff
@@ -0,0 +1,200 @@
...
+  public CloudEvent createCDEvent(
+      Map<String, Object> preference,
+      String application,
+      Event event,
+      Map<String, String> config,
+      String status,
+      String spinnakerUrl) {
+
+    String configType = Optional.ofNullable(config).map(c -> (String) c.get("type")).orElse(null);
+    String configLink = Optional.ofNullable(config).map(c -> (String) c.get("link")).orElse(null);
```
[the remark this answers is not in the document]
rjalander: yeah, throwing an existing FieldNotFoundException here

echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/cdevents/CDEventsBuilderService.java
```diff
@@ -0,0 +1,200 @@
...
+
+  public CloudEvent createCDEvent(
+      Map<String, Object> preference,
+      String application,
+      Event event,
+      Map<String, String> config,
+      String status,
+      String spinnakerUrl) {
+
+    String configType = Optional.ofNullable(config).map(c -> (String) c.get("type")).orElse(null);
```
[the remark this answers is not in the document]
rjalander: yeah, throwing an existing FieldNotFoundException here

echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/notification/CDEventsNotificationAgent.java
rjalander: `CDEventsException` is not needed now as using the retrofit request/response to send CDEvents notification

echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/cdevents/CDEventsBuilderService.java
```diff
@@ -0,0 +1,200 @@
...
+
+    SPDX-License-Identifier: Apache-2.0
+*/
+
+package com.netflix.spinnaker.echo.cdevents;
+
+import com.netflix.spinnaker.echo.api.events.Event;
+import dev.cdevents.CDEvents;
+import dev.cdevents.constants.CDEventConstants;
+import dev.cdevents.events.*;
```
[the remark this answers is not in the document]
rjalander: addressed

echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/cdevents/CDEventsBuilderService.java
```diff
@@ -0,0 +1,200 @@
...
+  private CloudEvent buildCloudEventWithCDEventType(
+      String cdEventsType,
+      String executionId,
+      String executionUrl,
+      String executionName,
+      String spinnakerUrl,
+      String status) {
+    CloudEvent ceToSend = null;
+    switch (cdEventsType) {
+      case "dev.cdevents.pipelinerun.queued":
```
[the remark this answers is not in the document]
rjalander: Yes I can pull that, there are Enum constants created with event types in the CDEvents SDK, 
```
PipelineRunFinishedEvent("dev.cdevents.pipelinerun.finished."),
PipelineRunQueuedEvent("dev.cdevents.pipelinerun.queued."),
```
May be I need to compare with substring of a `cdEventsType` to use with Enum constants

echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/cdevents/CDEventsBuilderService.java
```diff
@@ -0,0 +1,200 @@
...
+        ceToSend =
+            createTaskRunStartedEvent(executionId, executionUrl, executionName, spinnakerUrl);
+        break;
+      case "dev.cdevents.taskrun.finished":
+        ceToSend =
+            createTaskRunFinishedEvent(
+                executionId, executionUrl, executionName, spinnakerUrl, status);
+        break;
+      default:
+        throw new CDEventsException(
```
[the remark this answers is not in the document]
rjalander: As per my comment in the PR description, I will update the same PR/create another PR for all other events once this review is closed. 
`Current PR has the implementation to produce [Core CDEvents](https://github.com/cdevents/spec/blob/v0.1.2/core.md) for now, other events will be implemented after this PR review and discussion.`

My question here is does Spinnaker needs all types of events which CDEvents spec provides(https://cdevents.dev/docs/) or needs a subset of event types based on any Spinnaker requirements.
Like I am not sure whether [Source Code Control Events](https://cdevents.dev/docs/source-code-version-control/) are required for Spinnaker to send.

There is no such interface to create CDEvents(might be a good idea to propose this change from SDK itself as that interface/implementations can be used as common), 
I can update the current switch case to a map for flexibility and extensibility purpose.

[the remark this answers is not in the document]
xibz: > There is no such interface to create CDEvents(might be a good idea to propose this change from SDK itself as that interface/implementations can be used as common)

What I was meaning though is:

```java
public interface CDEventCreateTask {
    public CloudEvent create(String executionId, String executionUrl, String executionName, String spinnakerUrl, String status);
}
```

```java
public class CDEventCreateTaskRunFinished implements CDEventCreateTask {
// Code here
}
```

Then in this class

```java
private final Map<CDEventTypeEnum, CDEventCreateTask> createTasks = Map.of (
    CDEventTaskRunFinished, new CDEventCreateTaskRunFinished(),
);
```

That way the switch statement goes away and instead becomes something like

```java
CDEventCreateTask createTask = createTasks.get(eventTypeEnum);
if (createTask == null) {
    throw new SomeException("Invalid CDEvent create task");
}
```

echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/cdevents/CDEventsBuilderService.java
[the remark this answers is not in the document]
xibz: Yes, please do that :)

echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/cdevents/CDEventsBuilderService.java
```diff
@@ -0,0 +1,213 @@
...
+    return ceToSend;
+  }
+
+  private CloudEvent createTaskRunFinishedEvent(
+      String executionId,
+      String executionUrl,
+      String executionName,
+      String spinnakerUrl,
+      String status) {
+    TaskRunFinishedCDEvent cdEvent = new TaskRunFinishedCDEvent();
```
xibz, line 136: Then all these classes can just implement that interface I mentioned [above](https://github.com/spinnaker/echo/pull/1295/files/e7cf71404335dfe18a6391b54b0449006d33213e#r1359232738)

echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/cdevents/CDEventsBuilderService.java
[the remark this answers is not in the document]
rjalander: yes I'm in the same direction, was thinking aloud rather implementing in the Spinnaker is this a good idea to create these interface and classes in the CDEvents Java-SDK itself, 
but I know It will restrict to create CDEvents the way Spinnaker wants with any additional changes.
Will create these classes with little modifications here.
```java
 public interface CDEventCreator {
 //other type of events can have different parameters, so will remove the params
  CloudEvent createCDEvent();
}
```
Implement the classes for each event
```java
public class CDEventPipelineRunQueued implements CDEventCreator {

//each event can have different parameters, can go here
  private String source;
  private String subjectId;
  private String subjectSource;
  private String subjectPipelineName;
  private String subjectUrl;

  public CDEventPipelineRunQueued(String executionId, String executionUrl, String executionName, String spinnakerUrl) {
    this.source = spinnakerUrl;
    this.subjectId = executionId;
    this.subjectSource = spinnakerUrl;
    this.subjectPipelineName = executionName;
    this.subjectUrl = executionUrl;
  }

  @Override
  public CloudEvent createCDEvent() {
PipelineRunQueuedCDEvent cdEvent = new PipelineRunQueuedCDEvent();
}
```
creating a map of different events in this class,
```java
Map<CDEventTypeEnum, CDEventCreateTask> createTasks = Map.of (
      PipelineRunQueuedEnum, new CDEventPipelineRunQueued(executionId, executionUrl, executionName, spinnakerUrl)
    );
```

echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/cdevents/CDEventTaskRunStarted.java
```diff
@@ -0,0 +1,56 @@
...
+package com.netflix.spinnaker.echo.cdevents;
+
+import dev.cdevents.CDEvents;
+import dev.cdevents.events.TaskRunStartedCDEvent;
+import io.cloudevents.CloudEvent;
+import java.net.URI;
+
+public class CDEventTaskRunStarted implements CDEventCreator {
+
+  private String source;
```
xibz, line 26: We can probably move a lot of these to some base `CDEvent` class, since I imagine some of these will always be required for all `CDEventCreator`s.

rjalander: yes, making `CDEventCreator` class as an abstract and moving the common properties to this class

echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/cdevents/CDEventCreator.java
```diff
@@ -0,0 +1,67 @@
...
+  private String subjectUrl;
+
+  public CDEventCreator(String source, String subjectId, String subjectSource, String subjectUrl) {
+    this.source = source;
+    this.subjectId = subjectId;
+    this.subjectSource = subjectSource;
+    this.subjectUrl = subjectUrl;
+  }
+
+  public String getSource() {
```
xibz, line 36: Lombok can be used here

See
`@Getter`
`@Setter`
Makes the code a little more clean :)

echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/cdevents/CDEventCreator.java
```diff
@@ -0,0 +1,67 @@
...
+    limitations under the License.
+
+    SPDX-License-Identifier: Apache-2.0
+*/
+
+package com.netflix.spinnaker.echo.cdevents;
+
+import io.cloudevents.CloudEvent;
+
+public abstract class CDEventCreator {
```
xibz, line 21: This name is a little confusing. `CDEventCreator`? It's not really creating a CDEvent. It's more of a `BaseCDEvent` or something of that nature. Im sure there are better names than `BaseCDEvent`, but I think I'd much prefer that over `CDEventCreator` since no creating is happening here.

xibz (changes requested):
I just had a couple comments, but overall looks good. Can we also add docs around the `CDEventCreator`? Just something to outline what it is and link to CDEvents in general

echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/cdevents/CDEventsBuilderService.java
```diff
@@ -0,0 +1,110 @@
...
+    String configLink =
+        Optional.ofNullable(config)
+            .map(c -> (String) c.get("link"))
+            .orElseThrow(FieldNotFoundException::new);
+
+    String executionId =
+        Optional.ofNullable(event.content)
+            .map(e -> (Map) e.get("execution"))
+            .map(e -> (String) e.get("id"))
+            .orElseThrow(FieldNotFoundException::new);
```
nemesisOsorio, line 54: can we add the field name to the exceptions?
or log what is the missing field

echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/cdevents/CDEventsBuilderService.java
```diff
@@ -0,0 +1,110 @@
...
+
+    if (cdEventCreator == null) {
+      log.error("No mapping event type found for {}", cdEventsType);
+      log.error(
+          "The event type should be an event or substring of an event from the list of event types {}",
+          cdEventsMap.keySet());
+      throw new CDEventsException("No mapping eventType found to create CDEvent");
+    } else {
+      return cdEventCreator.createCDEvent();
+    }
```
nemesisOsorio, line 108: else not needed

echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/cdevents/CDEventsHTTPMessageConverter.java
```diff
@@ -0,0 +1,88 @@
...
+import retrofit.converter.Converter;
+import retrofit.mime.TypedByteArray;
+import retrofit.mime.TypedInput;
+import retrofit.mime.TypedOutput;
+
+public class CDEventsHTTPMessageConverter implements Converter {
+
+  private final ObjectMapper objectMapper;
+
+  private static final String MIME_TYPE = "application/json";
```
nemesisOsorio, line 42: I believe there is a constant in spring for this

echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/cdevents/CDEventsSenderClient.java
```diff
@@ -0,0 +1,26 @@
...
+    See the License for the specific language governing permissions and
+    limitations under the License.
+
+    SPDX-License-Identifier: Apache-2.0
+*/
+
+package com.netflix.spinnaker.echo.cdevents;
+
+import retrofit.client.Response;
+import retrofit.http.*;
```
nemesisOsorio, line 20: I believe we should avoid wildcards

echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/cdevents/CDEventsSenderService.java
```diff
@@ -0,0 +1,109 @@
...
+    CDEventsHTTPMessageConverter converterFactory = CDEventsHTTPMessageConverter.create();
+    RequestInterceptor authInterceptor =
+        new RequestInterceptor() {
+          @Override
+          public void intercept(RequestInterceptor.RequestFacade request) {
+            request.addHeader("Ce-Id", cdEvent.getId());
+            request.addHeader("Ce-Specversion", cdEvent.getSpecVersion().V1.toString());
+            request.addHeader("Ce-Source", cdEvent.getSource().toString());
+            request.addHeader("Ce-Type", cdEvent.getType());
+            request.addHeader("Content-Type", "application/json");
```
nemesisOsorio, line 53: reuse a constant

echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/cdevents/CDEventCreator.java
rjalander: yeah that looks good, using Lombok in other sub classes too.

echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/cdevents/CDEventCreator.java
rjalander: renamed to `BaseCDEvent`

echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/cdevents/CDEventsSenderClient.java
rjalander: yeah corrected, it was auto imported,

echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/cdevents/CDEventsHTTPMessageConverter.java
```diff
@@ -0,0 +1,88 @@
...
+    try {
+      JavaType javaType = objectMapper.getTypeFactory().constructType(type);
+      return objectMapper.readValue(body.in(), javaType);
+    } catch (JsonParseException e) {
+      throw new ConversionException(e);
+    } catch (JsonMappingException e) {
+      throw new ConversionException(e);
+    } catch (IOException e) {
+      throw new ConversionException(e);
+    }
```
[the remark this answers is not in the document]
rjalander: `IOException` can not be used with multi-catch as the other Exception classes are subclass of `IOException`

Using multi-catch for subclasses alone,
`catch (JsonParseException | JsonMappingException e)`

echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/cdevents/CDEventsHTTPMessageConverter.java
rjalander: yes, using MediaType.APPLICATION_JSON_VALUE from spring

echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/cdevents/CDEventsBuilderService.java
```diff
@@ -0,0 +1,110 @@
...
+                new CDEventTaskRunStarted(executionId, executionUrl, executionName, spinnakerUrl),
+            CDEventTypes.TaskRunFinishedEvent.getEventType(),
+                new CDEventTaskRunFinished(
+                    executionId, executionUrl, executionName, spinnakerUrl, status));
+    for (String keyType : cdEventsMap.keySet()) {
+      if (keyType.contains(cdEventsType)) {
+        cdEventCreator = cdEventsMap.get(keyType);
+        break;
+      }
+    }
```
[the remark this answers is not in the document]
rjalander: updated to use stream and filter

nemesisOsorio (changes requested):

echo-notifications/src/test/groovy/com/netflix/spinnaker/echo/notification/CDEventsNotificationAgentSpec.groovy
```diff
@@ -0,0 +1,66 @@
+/*
+ * Copyright 2016 Google, Inc.
```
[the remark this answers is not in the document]
rjalander: yes, updated

echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/cdevents/CDEventsSenderService.java
```diff
@@ -0,0 +1,110 @@
...
+      URL url = new URL(webhookUrl);
+      String endPointURL =
+          url.getPort() != -1
+              ? url.getProtocol() + "://" + url.getHost() + ":" + url.getPort()
+              : url.getProtocol() + "://" + url.getHost();
+      log.info("endpoint Url to send CDEvent {} ", endPointURL);
+      return endPointURL;
+    } catch (MalformedURLException e) {
+      throw new InvalidRequestException(
+          "Unable to determine base URL from Microsoft Teams webhook URL.", e);
```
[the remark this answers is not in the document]
rjalander: Apologies for the typo, corrected with valid message

echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/notification/CDEventsNotificationAgent.java
```diff
@@ -0,0 +1,85 @@
+/*
+ * Copyright 2020 Cerner Corporation
```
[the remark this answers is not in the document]
rjalander: Yes Updated

nemesisOsorio (approved):

jasonmcintosh (approved):
Overall better.  Frankly the APIs are a TOUCh rough on cdevents side.  Example:
* EVERY event right now looks like it duplicates that same call `cdEvent.setSource(URI.create(getSource()));` uniquely.  There isn't a base object that could handle this and reduce the duplication.  There appear to be about 3 or 4 fields which are duplicated across ALL cdevents stuff
* The URI parsing has a few places of "hey this is probably non ideal" on things like the format of spinnaker URls & such, but... for the moment it'd work.  Just a personal preference on UI urls vs. API urls and encoding.  Just depends on how downstream handles it, not enough to "block" the PR at this time.

xibz:
> EVERY event right now looks like it duplicates that same call cdEvent.setSource(URI.create(getSource())); uniquely. There isn't a base object that could handle this and reduce the duplication. There appear to be about 3 or 4 fields which are duplicated across ALL cdevents stuff

The SDKs are **very** low level currently for CDEvents. So no abstractions have been added to make things much easier. Generally when you generating from openapi, smithy, and any other API language, generally they may not generate base classes especially for languages like golang. It's completely up to the generator. However, what most people do is use that to generate clients but are not meant to be used as the base SDK. You see this in AWS SDKs where the clients are actually **very** low level, but there are higher level abstractions that make it easier. Due to CDEvents being still early in its life, these abstractions have no been made yet, unfortunately. This should get better with time though :).

jasonmcintosh:
Exactly :) Let's get this merged, and see how it gets better!

xibz (approved):
LGTM! :ship:

PR #1315, opened 2023-08-01
feat(rest/circuit-breaker): Optimize Circuit Breaker in Rest Events

sergio-quintero:
Previously, the resilience4j circuit breaker was isolated to the sendEvent() method. This worked but when Echo was processing large events, it tended to slow down and even crash. The reason was that Echo continues to receive events and write the Events to a Map using ObjectMapper.

This refactors the RestEventListener class to avoid mapping events with ObjectMapper if the Circuit Breaker does not permit the call because the state is OPEN or HALF_OPEN.

Features:
* Configure a Circuit Breaker per Rest Service

Optimizations:
* Do not write events with ObjectMapper if Circuit Breaker is open

Refactors:
* Refactor RestEventListener
* Refactor RestEventService
* Rewrite RestEventListener tests from Groovy-Spock to Java-JUnit-Mockito

To enable circuit breakers, you should enable it per service. Rest Service event names should be mapped to a resilience4j-circuitbreaker instance. Example using `my-service`:

```yaml
rest:
  enabled: true
  endpoints:
    - url: "https://my-service.com"
      eventName: "my-service"
      circuitBreakerEnabled: true

resilience4j.circuitbreaker:
  instances:
    my-service:
      slow-call-rate-threshold: 25
      slow-call-duration-threshold: 3s
      failure-rate-threshold: 25
      minimum-number-of-calls: 2
      automatic-transition-from-open-to-half-open-enabled: true
      wait-duration-in-open-state: 30s
      permitted-number-of-calls-in-half-open-state: 1
```


echo-rest/src/main/java/com/netflix/spinnaker/echo/events/RestEventListener.java
```diff
@@ -55,79 +68,188 @@ class RestEventListener implements EventListener {
...
+          m.put(StringUtils.defaultString(service.getConfig().getFieldName(), fieldName), eventMap);
+
+          eventMap = m;
+        }
+      }
+    } catch (Exception e) {
+      handleError(event, service, e);
+    }
+
+    return eventMap;
```
edgarulg, line 237: if an exception is thrown in the `transformEventToMap` method. Is it possible that an empty event will be sent in the `sendEvent` method?
Not sure if that's the behavior we have previously.

echo-rest/src/main/java/com/netflix/spinnaker/echo/events/RestEventListener.java
```diff
@@ -55,79 +68,188 @@ class RestEventListener implements EventListener {
...
+        StringUtils.defaultString(service.getConfig().getEventName(), "default");
+
+    CircuitBreaker circuitBreaker = circuitBreakerRegistry.circuitBreaker(circuitBreakerInstance);
+
+    try {
+      // acquire permission before writing the event to a Map
+      // this will save processing and memory usage if the Circuit Breaker is OPEN or HALF_OPEN
+      circuitBreaker.acquirePermission();
+
+      Map<String, Object> eventMap = transformEventToMap(event, service);
```
edgarulg, line 156: if an exception is thrown in the `transformEventToMap` method. does it try to send an event with errors?

echo-rest/src/test/java/RestEventListenerTest.java
```diff
@@ -0,0 +1,291 @@
...
+    Map<String, Object> expectedEvent =
+        listener.getMapper().convertValue(event, new TypeReference<>() {});
+
+    listener.processEvent(event);
+
+    Mockito.verify(restService, Mockito.times(1)).recordEvent(expectedEvent);
+  }
+
+  @Test
+  void shouldNotInvokeMultipleTimesAfterExceptionWhenCircuitBreakerEnabled() {
```
edgarulg, line 255: Can you include a test where an exception is thrown in the `transformEventToMap`method?

echo-rest/src/main/java/com/netflix/spinnaker/echo/events/RestEventListener.java
sergio-quintero: Good catch. Previously everything was in a single method and all the processing was done in a single try/catch block so if ObjectMapper threw an exception, the exception was logged and it never tried to send an empty event. I added a null check to avoid this scenario.

echo-rest/src/main/java/com/netflix/spinnaker/echo/events/RestEventListener.java
sergio-quintero: https://github.com/spinnaker/echo/pull/1315#discussion_r1283411288

echo-rest/src/test/java/RestEventListenerTest.java
sergio-quintero: Added a couple of tests for this.

edgarulg (approved):
LGTM but I would like to have more eyes on this PR @dbyron-sf

dbyron-sf:
This would be easier to follow if there were smaller, more specific commits.  Like, if converting the existing RestEventListenerSpec from groovy to java was one commit at the beginning, then it'd be easier to see what changed with the other code changes.  If you've got the stamina to split up the commits that way, it would help get confidence in the change.

dbyron-sf:
If there are other refactors (like making the members of RestEventListener final), it'd help to have those in separate commits as well.

dbyron-sf:
Yeah, as I stare at the PR there are a bunch of different things happening here.  Beyond avoiding the transformEventToMap stuff (adding that method could be another refactor commit), there also seems to be the ability to configure a circuit breaker, and also to have multiple circuit breakers, with a specific one associated with each rest endpoint.  It'll be so much easier to understand what's going on with refactor commits/feature commits and then maybe a fix commit or perf commit at the end to avoid unnecessary work when the circuit breaker is open.

sergio-quintero:
I agree. I'll split the changes into multiple commits.

nemesisOsorio (approved):

sergio-quintero:
Moved to Draft for now. This is WIP.

sergio-quintero:
@dbyron-sf @edgarulg @nemesisOsorio finished rebasing the commits. I ended up making some small changes compared to the initial PR. Hopefully it's easier to follow.

echo-rest/src/main/java/com/netflix/spinnaker/echo/config/RestProperties.java
```diff
@@ -47,5 +47,6 @@ public static class RestEndpointConfiguration {
     String headersFile;
     Boolean flatten = false;
     int retryCount = 1;
+    Boolean circuitBreakerEnabled = false;
```
nemesisOsorio, line 50: I like the idea to make it more granular

nemesisOsorio (approved):

echo-rest/src/main/java/com/netflix/spinnaker/echo/events/RestEventListener.java
```diff
@@ -39,7 +39,7 @@
 @ConditionalOnProperty("rest.enabled")
 @Getter
 @Setter
-class RestEventListener implements EventListener {
+public class RestEventListener implements EventListener {
```
dbyron-sf, line 45: What's the motivation for this?

sergio-quintero: Forgot to add a package to the test class. Fixed https://github.com/spinnaker/echo/commit/4568a320306ba2b5f897f9833d64c321437f02e3

dbyron-sf:
Much easier to follow now, thank you @sergio-quintero.  I think my remaining questions are about backwards compatibility.  Like, if someone uses this new code, and has `rest.circuit-breaker-enabled` set to true, but doesn't make any configuration changes (e.g. to define a circuit breaker instance), what happens?  Is there a test for this case?

sergio-quintero:
@dbyron-sf Before this PR, If someone only enabled `rest.circuit-breaker-enabled` without specifying any instance config, the circuit breaker would create a new instance named `sendEvent` with the default config. `sendEvent` name was hardcoded in the method annotation, so you needed to configure everything under that instance.

I added backwards compatibility (including a test) in https://github.com/spinnaker/echo/pull/1315/commits/36cd94ead62d27913c68c40aafec1e8b9c2d51ed

So now if someone is still using `rest.circuit-breaker-enabled` with or without any instance config, they can still use the circuit breaker as expected.

dbyron-sf:
Thanks @sergio-quintero , LGTM.  Could you write up some release notes for this?

dbyron-sf (approved):

sergio-quintero:
@dbyron-sf release notes https://github.com/spinnaker/spinnaker.io/pull/364

dbyron-sf:
@sergio-quintero can you rebase against the HEAD of master please?

PR #1342, opened 2023-09-26
feat(java17): build JRE 17 images, run tests with JRE 17

mattgogerly:
Number 3 on our Java 17 adventure.

See also https://github.com/spinnaker/front50/pull/1274, https://github.com/spinnaker/igor/pull/1143

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -16,13 +16,13 @@ jobs:
     if: startsWith(github.repository, 'spinnaker/')
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
         with:
           fetch-depth: 0
       - name: Set up QEMU
-        uses: docker/setup-qemu-action@v2
+        uses: docker/setup-qemu-action@v3
       - name: Set up Docker Buildx
-        uses: docker/setup-buildx-action@v2
+        uses: docker/setup-buildx-action@v3
       - uses: actions/setup-java@v3
         with:
           java-version: |
@@ -42,7 +42,7 @@ jobs:
       - name: Login to GAR
         # Only run this on repositories in the 'spinnaker' org, not on forks.
         if: startsWith(github.repository, 'spinnaker/')
-        uses: docker/login-action@v2
+        uses: docker/login-action@v3
         # use service account flow defined at: https://github.com/docker/login-action#service-account-based-authentication-1
         with:
           registry: us-docker.pkg.dev
@@ -51,7 +51,7 @@ jobs:
       - name: Build and publish slim container image
         # Only run this on repositories in the 'spinnaker' org, not on forks.
         if: startsWith(github.repository, 'spinnaker/')
-        uses: docker/build-push-action@v4
+        uses: docker/build-push-action@v5
         with:
           context: .
           file: Dockerfile.slim
@@ -65,7 +65,7 @@ jobs:
       - name: Build and publish ubuntu container image
         # Only run this on repositories in the 'spinnaker' org, not on forks.
         if: startsWith(github.repository, 'spinnaker/')
-        uses: docker/build-push-action@v4
+        uses: docker/build-push-action@v5
         with:
           context: .
           file: Dockerfile.ubuntu
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -10,13 +10,13 @@ jobs:
   build:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
         with:
           fetch-depth: 0
       - name: Set up QEMU
-        uses: docker/setup-qemu-action@v2
+        uses: docker/setup-qemu-action@v3
       - name: Set up Docker Buildx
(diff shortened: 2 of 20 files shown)


mattgogerly:
I have no idea why this test is randomly failing now.

dbyron-sf:
> I have no idea why this test is randomly failing now.

Some word wrapping behavior is different with java 17 maybe?

mattgogerly:
> > I have no idea why this test is randomly failing now.
> 
> Some word wrapping behavior is different with java 17 maybe?

Looks like it's [this](https://openjdk.org/jeps/355), introduced in Java 13. `stripIndent()` didn't exist as a Java method in 11, and came from Groovy instead. Now it does exist as of 13.

> For each [non-blank](https://docs.oracle.com/en/java/javase/13/docs/api/java.base/java/lang/String.html#isBlank()) line, min leading [white space](https://docs.oracle.com/en/java/javase/13/docs/api/java.base/java/lang/Character.html#isWhitespace(int)) characters are removed, and any trailing [white space](https://docs.oracle.com/en/java/javase/13/docs/api/java.base/java/lang/Character.html#isWhitespace(int)) characters are removed. [Blank](https://docs.oracle.com/en/java/javase/13/docs/api/java.base/java/lang/String.html#isBlank()) lines are replaced with the empty string.

Gotta love removing trailing whitespace in a method called `stripIndent()` 😅 

I guess we can either just remove the test, because it seemingly now "works" as of 13, or we can force it to call the Groovy version. Any preferences?

echo-notifications/src/test/groovy/com/netflix/spinnaker/echo/notification/ManualJudgmentTemplateTest.groovy
```diff
@@ -159,16 +160,17 @@ class ManualJudgmentTemplateTest extends Specification {
     rendered == expected
 
     where:
-
+    // @formatter:off
     // Note, there's a whitespace after each new line of the `instructions` because of `Jsoup.parse(...)`.
-    // This is because the field is HTML parsed to plaintext. Which also can contain Slack's psudo-markdown flavor.
-    expected = '''\
-        Stage <SPINNAKER_URL/#/applications/testapp/executions/details/exec-id?refId=stage-id&step=1|stage-name> for *testapp*'s *exe-name* pipeline build #*12345* is awaiting manual judgment.
+    // This is because the field is HTML parsed to plaintext. Which also can contain Slack's sudo-markdown flavor.
```
dbyron-sf, line 165: I think this is supposed to be
```suggestion
    // This is because the field is HTML parsed to plaintext. Which also can contain Slack's pseudo-markdown flavor.
```

echo-notifications/src/test/groovy/com/netflix/spinnaker/echo/notification/ManualJudgmentTemplateTest.groovy
```diff
@@ -17,6 +17,7 @@
 package com.netflix.spinnaker.echo.notification
 
 import com.netflix.spinnaker.echo.api.Notification
+import org.codehaus.groovy.runtime.StringGroovyMethods
```
dbyron-sf, line 20: Confusing that IntelliJ thinks this isn't used, yet tests fail without it.

echo-notifications/src/test/groovy/com/netflix/spinnaker/echo/notification/ManualJudgmentTemplateTest.groovy
```diff
@@ -159,16 +160,17 @@ class ManualJudgmentTemplateTest extends Specification {
     rendered == expected
 
     where:
-
```
dbyron-sf, line 162: Maybe the commit message could use a tweak since there are still some uses of `stripIndent` in the file?  Don't suppose you're fired up to convert this test to java?

echo-notifications/src/test/groovy/com/netflix/spinnaker/echo/notification/ManualJudgmentTemplateTest.groovy
mattgogerly: Yeah it actually isn't - I actually just changed the entire thing to be a Java String block rather than relying on `stripIndent`. I'll fix it up.

echo-notifications/src/test/groovy/com/netflix/spinnaker/echo/notification/ManualJudgmentTemplateTest.groovy
mattgogerly: I'll give it a go.

mattgogerly:
@Mergifyio update

dbyron-sf:
LGTM, but there's an error building the slim container image:
```
#9 [linux/amd64 2/4] RUN adduser --system --uid 10111 --group spinnaker
#9 0.115 adduser: group 'spinnaker' does not exist
#9 ERROR: process "/bin/sh -c adduser --system --uid 10111 --group spinnaker" did not complete successfully: exit code: 6
```

dbyron-sf (approved):

mattgogerly:
@Mergifyio update

PR #1357, opened 2023-11-09
fix(cdevents-webhooks) : Adding support for Artifact Constraints and Parameters 

rjalander:
This PR includes the changes to fix the CDEvents type Automated trigger to run the pipeline executions with Artifact Constraints and Parameters.

The structure of the CDEvents is defined in the [CDEvents-spec](https://github.com/cdevents/spec/blob/main/spec.md#overview)

The sample CDEvent data with Parameters in the `customData` object
````json
{
  "context": {
    "id": "5fb38d7d-28dd-47e4-ade2-cb21153cc8dc",
    "type": "dev.cdevents.artifact.published.0.1.0",
    "source": "https://ci-build.est.tech/",
    "version": "0.1.2"
  },
  "customData": {
    "parameters": {
      "stack": "prod"
    }
  },
  "customDataContentType": "application/json",
  "subject": {
    "type": "ARTIFACT",
    "content": {
      "pipelineName": "test-1",
      "outcome": "SUCCESS"
    }
  }
}
````
Sample CDEvent data with artifacts in the `customData` object
````json
{
  "context": {
    "id": "5fb38d7d-28dd-47e4-ade2-cb21153cc8dc",
    "type": "dev.cdevents.artifact.published.0.1.0",
    "source": "https://ci-build.est.tech/",
    "version": "0.1.2"
  },
  "customData": {
    "artifacts": [
      {
        "type": "custom/object",
        "name": "image-cdevents",
        "version": "1.0.1",
        "reference": "custom/object/artifact"
      }
    ]
  },
  "customDataContentType": "application/json",
  "subject": {
    "type": "ARTIFACT",
    "content": {
      "pipelineName": "test-1",
      "outcome": "SUCCESS"
    }
  }
}
````

Dependent PR : https://github.com/spinnaker/gate/pull/1732

### DIFF
--- a/echo-webhooks/echo-webhooks.gradle
+++ b/echo-webhooks/echo-webhooks.gradle
@@ -19,6 +19,7 @@ dependencies {
   implementation project(':echo-model')
   implementation "org.springframework.boot:spring-boot-starter-web"
   implementation "io.spinnaker.kork:kork-artifacts"
+  implementation "io.spinnaker.kork:kork-web"
   implementation "javax.validation:validation-api"
   implementation "io.cloudevents:cloudevents-spring:2.5.0"
   implementation "io.cloudevents:cloudevents-json-jackson:2.5.0"
--- a/echo-webhooks/echo-webhooks.gradle
+++ b/echo-webhooks/echo-webhooks.gradle
@@ -19,7 +19,7 @@ dependencies {
   implementation project(':echo-model')
   implementation "org.springframework.boot:spring-boot-starter-web"
   implementation "io.spinnaker.kork:kork-artifacts"
-  implementation "io.spinnaker.kork:kork-web"
+  implementation "io.spinnaker.kork:kork-core"
   implementation "javax.validation:validation-api"
   implementation "io.cloudevents:cloudevents-spring:2.5.0"
   implementation "io.cloudevents:cloudevents-json-jackson:2.5.0"
--- a/echo-webhooks/src/main/groovy/com/netflix/spinnaker/echo/controllers/WebhooksController.groovy
+++ b/echo-webhooks/src/main/groovy/com/netflix/spinnaker/echo/controllers/WebhooksController.groovy
@@ -138,13 +138,12 @@ class WebhooksController {
                                     @RequestBody CloudEvent cdevent,
                                     @RequestHeader HttpHeaders headers) {
     log.info("CDEvents Webhook received with source ${source} and with event type ${cdevent.getType()}")
-
     String ceDataJsonString  = headers.get("Ce-Data").get(0)
     log.info("CDEvent received with in the CloudEvent data request {}", ceDataJsonString)
     Map postedEvent
     try {
       postedEvent = mapper.readValue(ceDataJsonString, Map) ?: [:]
-      if (!postedEvent.get("customData") || !postedEvent.get("context") || !postedEvent.get("subject")) {
+      if (postedEvent.get("customData") == null || !postedEvent.get("context") || !postedEvent.get("subject")) {
         throw new RuntimeException("Invalid CDEvent data posted with the CloudEvent RequestBody - " + postedEvent);
       }
     } catch (Exception e) {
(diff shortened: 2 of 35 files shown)

### CODE REVIEW

File: echo-webhooks/src/main/groovy/com/netflix/spinnaker/echo/controllers/WebhooksController.groovy [hunk 1]
```diff
@@ -140,31 +138,43 @@ class WebhooksController {
...
-    event.details.source = source
-    event.details.type = "cdevents"
-    event.details.requestHeaders = headers
-    event.rawContent = ceDataJsonString
+    String ceDataJsonString  = headers.get("Ce-Data").get(0)
+    log.info("CDEvent received with in the CloudEvent data request {}", ceDataJsonString)
     Map postedEvent
     try {
       postedEvent = mapper.readValue(ceDataJsonString, Map) ?: [:]
+      if (postedEvent.get("customData") == null || !postedEvent.get("context") || !postedEvent.get("subject")) {
```
jasonmcintosh, line 146: We're missing a number of tests around this logic here it seems like.  Be good to have some better tests on this, and maybe a doc stating why this is "invalid CDEvent" data?  ALSO... "RuntimeException" would be an incorrect exception given spinnaker standards.

rjalander: There is a validation for `artifacts` and `parameters` in the existing [test](https://github.com/spinnaker/echo/pull/1357/files#diff-3d61ce6cb6a561b259741a1af9c4616d5f34f6250551c94b56f0bfe57eaaacffR888-R889) with valid CDEvents data(will separate each case with different test, If needed ?) and there is a new test to verify invalid CDEvents data.

In the place of "RuntimeException" I can change to `com.netflix.spinnaker.kork.web.exceptions.InvalidRequestException` but I need to add a dependency of `io.spinnaker.kork:kork-web` in the echo-webhooks.gradle project, If that is fine?

I have create a [PR](https://github.com/spinnaker/spinnaker.io/pull/392), user guide to trigger the pipeline using CDEvents, where I mentioned how valid CDEvents data can be  constructed as per [CDEvents-spec](https://github.com/cdevents/spec/blob/main/spec.md)

jasonmcintosh: The existing test does a lot in ONE test.  i'd prefer individual tests for different use cases with some different parameters situation to try to capture edge cases.  Given the change in logic, I think this is showing not all scenarios are covered correctly by the tests.  AKA this change should have caused a test failure and didn't.  

InvalidRequestException comes from kork-exceptions which is included via kork-core.  SO no need for kork-web.  This provides actual feedback/better response handling through various paths as a rule and why we use it, or sometimes the error messages get lost across services.

echo-webhooks/echo-webhooks.gradle
```diff
@@ -19,6 +19,7 @@ dependencies {
   implementation project(':echo-model')
   implementation "org.springframework.boot:spring-boot-starter-web"
   implementation "io.spinnaker.kork:kork-artifacts"
+  implementation "io.spinnaker.kork:kork-web"
```
jasonmcintosh, line 22: Shouldn't be necessary.  Kork-core should have this.  The exception class should be coming in transitively automatically via one of the other dependencies vs. the entire web which isn't necessary.

rjalander: Removed this, but currently kork-core doesn't exist in this file.
Adding kork-core dependency to find the exception class as transitive dependency. 
If I don't add any dependency the build fails in finding the exception class.

echo-webhooks/src/main/groovy/com/netflix/spinnaker/echo/controllers/WebhooksController.groovy
rjalander: Created separate test cases for each valid and invalid cases

jasonmcintosh (approved):

rjalander:
@jasonmcintosh Can you please merge this if there are no other comments. Thank you.

PR #1373, opened 2023-12-14
feat(jenkins): Enable Jenkins job triggers for jobs in sub-folders

ciurescuraul:
### Description
When defining a Jenkins job in a sub-folder, the path contains slashes. Consequently, the Spring framework is not capable of accurately matching the request.

### Problem
The trigger request fails because it is attempting to initiate a job that is not found. 

### Solution
A feature flag has been added. This flag influences if we should call the existing endpoint (which accepts the job name as a path variable) or the new one (which accepts the job name as a query parameter).

### Before
image

### After
image

### Related PR's
Orca: https://github.com/spinnaker/orca/pull/4618
Igor: https://github.com/spinnaker/igor/pull/1204


ovidiupopa07 (approved):
LGTM

jasonmcintosh (approved):
Nice!  There's some thought long term that... we have a lot of "feature flags", should move some of the concepts into Kork and maybe generic it all so 
`feature.<service>.flagName` kinda properties could be consistent across services.  Be a nice enhancement later.

echo-core/src/main/java/com/netflix/spinnaker/echo/services/IgorService.java
```diff
@@ -30,6 +31,12 @@ Map<String, Object> getBuild(
       @Path("master") String master,
       @Path(value = "job", encode = false) String job);
 
+  @GET("/builds/status/{buildNumber}/{master}")
+  Map<String, Object> getBuildStatusWithJobQueryParameter(
+      @NotNull @Path("buildNumber") Integer buildNumber,
+      @NotNull @Path("master") String master,
+      @NotNull @Query(value = "job") String job);
```
jasonmcintosh, line 38: ONLY call out is... encoding. Calling out the caution on the other PR, but ... not sure how much of an issue that really is.  IF job should be URL encoded due to any kinda characters - I don't THINK so at this point though.

PR #1405, opened 2024-03-29
feat(build): add echo-integration module to exercise the just-built docker image

dbyron-sf:
Also, add a healthcheck to the docker images, and a few other small cleanups


echo-web/config/echo.yml
```diff
@@ -12,8 +12,8 @@ front50:
 
 services:
   fiat:
-    enabled: ${services.fiat.enabled:false}
-    baseUrl: ${services.fiat.baseUrl:http://localhost:8089}
+    enabled: false
```
jasonmcintosh, line 15: Hrmm isn't this the config that allows overrides?  I cna't recall the impact of fixing this to false on this file...

jasonmcintosh: Actually think probably redundant b/c of the way this operates in newer spring... ignore this ;)

jasonmcintosh (approved):

echo-web/config/echo.yml
dbyron-sf: Maybe it's only spring 2.6 that complains about the circular reference.  There are truckloads of precedence rules, I expect existing overrides to keep working.

PR #1434, opened 2024-07-08
test(pipelinetriggers): demonstrate behavior of SpEL expression evaluation

dbyron-sf:
that uses a non-public constructor/method, and so requires special treatment under java 17.


j-sandy (approved):
LGTM !

PR #1442, opened 2024-09-05
refactor(dependency): replace groovy coordinates during upgrade of groovy 4.x

j-sandy:
Replacing the groovy coordinates from `org.codehaus.groovy` to `org.apache.groovy` supported by groovy 4.x and above versions. Upgrading `@Wither` to `@With`, as suggested [here](https://projectlombok.org/features/With).


j-sandy:
https://projectlombok.org/api/lombok/experimental/Wither

dbyron-sf:
included in https://github.com/spinnaker/echo/pull/1443

PR #1466, opened 2024-12-29
refactor(retrofit2): replace retrofit client with retrofit2 client

kirangodishala:
- This PR makes use of retrofit 2.x client in place of retrofit 1.9 and make required changes accordingly.
- Replaced all RetrofitError references with SpinnakerServerException and/or its sub classes. 
- All the retrofit1 dependencies and references are removed.
- Retrofit2 boilerplate code was already implemented in kork's kork-retrofit module and this PR makes use of it.


echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/cdevents/CDEventsConverterFactory.java
```diff
@@ -0,0 +1,85 @@
+/*
```
dbyron-sf, line 1: move to src/main/java?

echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/notification/CDEventsNotificationAgent.java
```diff
@@ -70,16 +71,22 @@ public void sendNotifications(
...
+      try {
+        log.info(
+            "Received response from events broker : {} {} for execution id {}. {}",
+            response.code(),
+            response.message(),
+            executionId,
+            response.body() != null ? response.body().string() : "");
+      } catch (IOException e) {
+        log.info(
+            "Received response from events broker : {} {} for execution id {}",
```
dbyron-sf, line 85: Include some indication that there was an error serializing the response body / some differentiation from the above message?

echo-notifications/echo-notifications.gradle
```diff
@@ -42,5 +42,6 @@ dependencies {
   testImplementation "org.apache.httpcomponents:httpclient"
   testImplementation "org.spockframework:spock-spring"
   testImplementation "org.springframework:spring-test"
+  testImplementation "com.squareup.retrofit2:retrofit-mock"
```
dbyron-sf, line 44: OK to remove
```
implementation "com.squareup.retrofit:converter-jackson"
```
from above?

echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/config/GoogleChatConfig.groovy
```diff
@@ -16,47 +16,40 @@
...
-            .setLogLevel(retrofitLogLevel)
-            .setLog(new Slf4jRetrofitLogger(GoogleChatClient.class))
+    def chatClient = new Retrofit.Builder()
+            .baseUrl(baseUrl)
+            .client(okHttpClientConfig.createForRetrofit2().build())
+            .addCallAdapterFactory(ErrorHandlingExecutorCallAdapterFactory.getInstance())
+            .addConverterFactory(JacksonConverterFactory.create())
             .build()
-            .create(GoogleChatClient.class)
+            .create(GoogleChatService.class)
```
dbyron-sf, line 50: Isn't this meant to stay GoogleChatClient.class since that's the retrofit2 interface?

echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/config/GoogleChatConfig.groovy
```diff
@@ -16,47 +16,40 @@
...
 
-    new GoogleChatService(chatClient)
+    new GoogleChatService(chatClient as GoogleChatClient)
```
dbyron-sf, line 52: and then no need for a cast here.

echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/googlechat/GoogleChatService.java
```diff
@@ -27,7 +27,7 @@ public GoogleChatService(GoogleChatClient googleChatClient) {
     this.googleChatClient = googleChatClient;
   }
 
-  Response sendMessage(String webhook, GoogleChatMessage message) {
-    return googleChatClient.sendMessage(webhook, message);
+  void sendMessage(String webhook, GoogleChatMessage message) {
```
dbyron-sf, line 30: Why change the signature here?  Doesn't the caller care about the response?  If not that seems like something to clean up in another PR.

echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/config/GithubConfig.java
```diff
@@ -16,47 +16,38 @@
...
-            .setEndpoint(githubEndpoint)
-            .setConverter(new JacksonConverter())
-            .setClient(retrofitClient)
-            .setLogLevel(retrofitLogLevel != null ? retrofitLogLevel : RestAdapter.LogLevel.BASIC)
-            .setLog(new Slf4jRetrofitLogger(GithubService.class))
-            .build()
-            .create(GithubService.class);
-
-    return githubClient;
+    return new Retrofit.Builder()
```
dbyron-sf, line 45: The above code doesn't have:
```
.setErrorHandler(SpinnakerRetrofitErrorHandler.getInstance())
```
so...is there still some RetrofitError-based exception handling?  Or maybe there isn't, and we're getting an improvement by moving to ErrorHandlingExecutorCallAdapterFactory....

echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/github/GithubService.java
```diff
@@ -16,20 +16,26 @@
...
+import retrofit2.http.POST;
+import retrofit2.http.Path;
 
 public interface GithubService {
   @POST("/repos/{repo}/statuses/{sha}")
-  Response updateCheck(
+  Call<Response<ResponseBody>> updateCheck(
       @Header("Authorization") String token,
-      @Path(value = "repo", encode = false) String repo,
+      @Path(value = "repo") String repo,
```
dbyron-sf, line 32: How come `encode = false` goes away?

echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/notification/GithubNotificationAgent.java
```diff
@@ -128,14 +127,8 @@ public void sendNotifications(
...
-    Response response = githubService.getCommit("token " + token, repo, sha);
-    ObjectMapper objectMapper = EchoObjectMapper.getInstance();
-    GithubCommit message = null;
-    try {
-      message = objectMapper.readValue(response.getBody().in(), GithubCommit.class);
-    } catch (IOException e) {
-      return sha;
-    }
+    GithubCommit message =
+        Retrofit2SyncCall.execute(githubService.getCommit("token " + token, repo, sha));
```
dbyron-sf, line 131: With the change in return type to GithubCommit from Response, retrofit2 is now doing the conversion, which is great.  Is there some equivalent try/catch block to add so this method returns the sha if the conversion fails?

echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/config/SlackConfig.groovy
```diff
@@ -51,20 +49,17 @@ class SlackConfig {
...
-                            RestAdapter.LogLevel retrofitLogLevel) {
+                            OkHttp3ClientConfiguration okHttpClientConfig) {
 
-    Endpoint slackEndpoint = newFixedEndpoint(config.baseUrl)
     log.info("Using Slack {}: {}.", config.useIncomingWebhook ? "incoming webhook" : "chat api", config.baseUrl)
 
-    def slackClient = new RestAdapter.Builder()
-        .setEndpoint(slackEndpoint)
-        .setConverter(new JacksonConverter())
-        .setClient(retrofitClient)
```
dbyron-sf, line 63: I'll stop commenting, but...another place with no SpinnakerRetrofitErrorHandler.

echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/github/GithubService.java
dbyron-sf: Aaah, the Path annotation in retrofit2 uses encoded, not encode, but it defaults to false....where encode defaults to true in retrofit1.  I think this means we'd add `encoded = true` to every @Path annotation that doesn't specify encode to avoid a behavior change....except that feels painful.

echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/slack/SlackNotificationService.groovy
```diff
@@ -22,6 +22,7 @@ import com.netflix.spinnaker.echo.controller.EchoResponse
 import com.netflix.spinnaker.echo.notification.NotificationService
 import com.netflix.spinnaker.echo.notification.NotificationTemplateEngine
 import groovy.util.logging.Slf4j
+import okhttp3.ResponseBody
```
dbyron-sf, line 25: unused import?

echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/slack/SlackNotificationService.groovy
```diff
@@ -62,8 +63,7 @@ class SlackNotificationService implements NotificationService {
           new SlackAttachment(subject, text, (InteractiveActions)notification.interactiveActions),
           address, true)
       }
-      log.trace("Received response from Slack: {} {} for message '{}'. {}",
-        response?.status, response?.reason, text, response?.body)
+      log.trace("Received response from Slack: {}", response?.string())
```
dbyron-sf, line 66: Does this show a gap in Retrofit2SyncCall that the entire Response is no longer available...only the ResponseBody?  I suppose if we get here there's no exception, so we know the response code is 2xx...and since this is a trace message it seems OK to drop that info.

echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/config/PagerDutyConfig.java
```diff
@@ -16,41 +16,39 @@
...
 public class PagerDutyConfig {
   private static final Logger log = LoggerFactory.getLogger(PagerDutyConfig.class);
 
-  @Bean
-  Endpoint pagerDutyEndpoint() {
-    return newFixedEndpoint("https://events.pagerduty.com");
+  private final String endpoint;
+
+  public PagerDutyConfig(
+      @Value("${pager-duty.endpoint:https://events.pagerduty.com}") String endpoint) {
```
dbyron-sf, line 39: Was this configurable before?

echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/pagerduty/PagerDutyNotificationService.groovy
```diff
@@ -82,24 +81,13 @@ class PagerDutyNotificationService implements NotificationService {
...
-          if (errorResponse.errors && errorResponse.errors.size() > 0) {
-            errorMessage = errorResponse.errors.join(", ")
-          }
-        }
+      } catch (SpinnakerServerException e){
         log.error('Failed to send page {} {} {}',
           kv('serviceKey', serviceKey), kv('message',
           notification.additionalContext.message),
-          kv('error', errorMessage)
+          kv('error', e.message)
```
dbyron-sf, line 88: We're losing a bit of functionality here.  With RetrofitError, errorMessage looked at the `errors` field in the response body.  Now we're only using e.message.  SpinnakerHttpException's message looks at the `message` field, not `errors`.  Seems reasonable to at least try to grab the `errors` property from SpinnakerHttpException's responseBody map.

echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/microsoftteams/MicrosoftTeamsService.java
```diff
@@ -16,51 +16,38 @@
...
-        new RestAdapter.Builder()
-            .setConverter(new JacksonConverter())
-            .setClient(retrofitClient)
-            .setEndpoint(getEndpointUrl(webhookUrl))
-            .setLogLevel(retrofitLogLevel)
-            .setLog(new Slf4jRetrofitLogger(MicrosoftTeamsClient.class))
+        new Retrofit.Builder()
+            .baseUrl(webhookUrl)
+            .client(okHttp3ClientConfiguration.createForRetrofit2().build())
+            .addConverterFactory(JacksonConverterFactory.create())
```
dbyron-sf, line 45: missing
```
.addCallAdapterFactory(ErrorHandlingExecutorCallAdapterFactory.getInstance())
```
?

echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/notification/MicrosoftTeamsNotificationAgent.java
```diff
@@ -141,13 +141,17 @@ public void sendNotifications(
...
+    ResponseBody response = teamsService.sendMessage(webhookUrl, teamsMessage);
+
+    try {
+      log.info(
+          "Received response from Microsoft Teams Webhook for execution id {}. {}",
+          executionId,
+          response.string());
+    } catch (IOException e) {
+      log.info(
+          "Received response from Microsoft Teams Webhook  for execution id {} but failed to deserialize",
```
dbyron-sf, line 153:
```suggestion
          "Received response from Microsoft Teams Webhook for execution id {} but failed to deserialize",
```

echo-telemetry/src/main/java/com/netflix/spinnaker/echo/config/TelemetryConfig.kt
```diff
@@ -77,8 +65,6 @@ open class TelemetryConfig {
     var instanceId = ULID().nextULID()
     var spinnakerVersion = "unknown"
     var deploymentMethod = DeploymentMethod()
-    var connectionTimeoutMillis = 3000
```
dbyron-sf, line 80: How can we get back this configurability?

echo-telemetry/src/test/kotlin/com/netflix/spinnaker/echo/telemetry/TelemetryEventListenerTest.kt
```diff
@@ -19,28 +19,29 @@ package com.netflix.spinnaker.echo.telemetry
...
 import com.netflix.spinnaker.echo.api.events.Event
-import com.netflix.spinnaker.echo.api.events.Event as EchoEvent
 import com.netflix.spinnaker.echo.api.events.Metadata
-import com.netflix.spinnaker.kork.proto.stats.Event as StatsEvent
+import com.netflix.spinnaker.kork.retrofit.exceptions.SpinnakerNetworkException
 import io.github.resilience4j.circuitbreaker.CircuitBreakerRegistry
-import io.mockk.Called
-import io.mockk.CapturingSlot
-import io.mockk.every
+import io.mockk.*
```
dbyron-sf, line 25: please avoid star imports

echo-telemetry/src/test/kotlin/com/netflix/spinnaker/echo/telemetry/TelemetryEventListenerTest.kt
```diff
@@ -131,18 +132,19 @@ class TelemetryEventListenerTest {
...
   @Test
-  fun `RetrofitError from service is ignored`() {
+  fun `SpinnakerNetworkException from service is ignored`() {
     val event = createLoggableEvent()
-
+    val request: Request = Request.Builder().url("http://url").build()
     every { telemetryService.log(any()) } throws
-      RetrofitError.networkError("url", IOException("network error"))
-
+      SpinnakerNetworkException(IOException("timeout"), request)
```
dbyron-sf, line 147: nit, but wouldn't `IOException("network error")` reduce the diffs slightly?

echo-notifications/echo-notifications.gradle
```diff
@@ -18,6 +18,7 @@ dependencies {
   implementation project(':echo-core')
   implementation project(':echo-model')
   implementation project(":echo-pipelinetriggers")
+  implementation project(":echo-test")
```
dbyron-sf, line 21: testImplementation?

echo-rest/echo-rest.gradle
```diff
@@ -15,19 +15,20 @@
...
 
 dependencies {
-  implementation "com.squareup.retrofit:retrofit"
-  implementation "com.squareup.retrofit:converter-jackson"
-  implementation "com.jakewharton.retrofit:retrofit1-okhttp3-client"
+  implementation "com.squareup.retrofit2:converter-jackson"
   implementation project(':echo-model')
   implementation project(':echo-api')
   implementation project(':echo-core')
+  implementation project(':echo-test')
```
dbyron-sf, line 22: testImplementation?

echo-pipelinetriggers/echo-pipelinetriggers.gradle
```diff
@@ -20,8 +20,11 @@ dependencies {
   implementation "io.spinnaker.fiat:fiat-api:$fiatVersion"
   implementation "io.spinnaker.fiat:fiat-core:$fiatVersion"
   implementation "com.squareup.retrofit:retrofit"
```
dbyron-sf: remove?

echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/slack/SlackInteractiveNotificationService.groovy
```diff
@@ -155,26 +155,25 @@ class SlackInteractiveNotificationService extends SlackNotificationService imple
...
+      .addConverterFactory(JacksonConverterFactory.create())
       .build()
-      .create(SlackHookService.class)
+      .create(SlackHookService.class);
   }
 
   interface SlackHookService {
     @POST('/{path}')
-    Response respondToMessage(@Path(value = "path", encode = false) path, @Body Map content)
+    Call<ResponseBody> respondToMessage(@Path(value = "path", encoded = false) path, @Body Map content)
```
dbyron-sf, line 177: Other places in this PR ditch `encoded = false` since it's the default now.

echo-notifications/echo-notifications.gradle
```diff
@@ -22,8 +22,6 @@ dependencies {
   implementation "org.springframework.boot:spring-boot-starter-web"
```
dbyron-sf, line 22: Not really related to this PR, but shouldn't the above
```
implementation project(":echo-test")
```
be
```
testImplementation project(":echo-test")
```
?

echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/notification/InteractiveNotificationCallbackHandler.groovy
```diff
@@ -99,9 +97,8 @@ class InteractiveNotificationCallbackHandler {
 
     // TODO(lfp): error handling (retries?). I'd like to respond to the message in a thread, but
     //  have been unable to make that work. Troubleshooting with Slack support.
-    // TODO(lfp): need to retrieve user's acccounts to pass in X-SPINNAKER-ACCOUNTS
```
dbyron-sf, line 102: Why change the comment here?  Is there something new happening with X-SPINNAKER-ACCOUNTS?

echo-core/src/test/groovy/com/netflix/spinnaker/echo/services/Front50ServiceSpec.groovy
```diff
@@ -68,196 +96,196 @@ class Front50ServiceSpec extends Specification {
...
     when:
-    def pipelines = front50.getPipelines()
+    def pipelines = Retrofit2SyncCall.execute(front50Service.getPipelines())
 
     then:
     pipelines.first().parallel
   }
 
-  @Ignore
-  def "list properties are immutable"() {
```
dbyron-sf, line 81: Can this test stay, even if it stays ignored?

echo-core/src/test/groovy/com/netflix/spinnaker/echo/services/Front50ServiceSpec.groovy
```diff
@@ -68,196 +96,196 @@ class Front50ServiceSpec extends Specification {
...
-        "type"        : "jenkins",
-        "master"      : "spinnaker",
-        "job"         : "Dummy_test_job",
-        "propertyFile": "deb.properties"
-      ]
-    ],
-    "application": "rush",
-    "appConfig"  : null,
-    "index"      : 1
+    [
```
dbyron-sf, line 140: Why the diffs here and below?

echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/jira/JiraNotificationService.java
```diff
@@ -189,10 +189,9 @@ private Map<String, Object> issueRequestBody(Notification notification) {
   }
 
   private Map errors(Exception exception) {
-    if (exception instanceof RetrofitError) {
+    if (exception instanceof SpinnakerHttpException) {
       try {
-        return mapper.readValue(
-            ((RetrofitError) exception).getResponse().getBody().in(), Map.class);
+        return ((SpinnakerHttpException) exception).getResponseBody();
       } catch (Exception e) {
```
dbyron-sf, line 195: Do we still need the catch block here, since SpinnakerHttpException.getResponseBody is a plain old accessor?

echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/config/GoogleChatConfig.groovy
kirangodishala: it's corrected now.

echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/config/GoogleChatConfig.groovy
kirangodishala: Agreed. Changed now.

echo-core/src/test/groovy/com/netflix/spinnaker/echo/services/Front50ServiceSpec.groovy
kirangodishala: It's ignored since it's inception. I feel it was meant to be improvised at later point but never happened so I thought the idea to make it a real test is expired. Do you want me to keep it anyway?

echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/github/GithubService.java
kirangodishala: I just learned that `encode = false` in retrofit1 and `encoded = false` retrofit2 have opposite meanings. 

https://square.github.io/retrofit/2.x/retrofit/retrofit2/http/Query.html#encoded--
```
Parameter names and values are URL encoded by default. Specify encoded=true to change this behavior.
```
`encode = false` in retrofit1 means - DO NOT ENCODE (it's already encoded)
`encoded = false` retrofit2  means - ITS NOT ENCODED (so please encode)

I will replace `encode = false` with `encoded = true` to preserve the behaviour.

echo-core/src/test/groovy/com/netflix/spinnaker/echo/services/Front50ServiceSpec.groovy
dbyron-sf: In the spirit of minimizing the diffs, I'd say keep it.

echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/slack/SlackInteractiveNotificationService.groovy
kirangodishala: It's addressed now.

echo-telemetry/src/main/java/com/netflix/spinnaker/echo/config/TelemetryConfig.kt
kirangodishala: Now it is part of ok-http-client ([OkHttpClientConfigurationProperties](https://github.com/spinnaker/kork/blob/master/kork-web/src/main/groovy/com/netflix/spinnaker/okhttp/OkHttpClientConfigurationProperties.groovy#L30)) configuration properties.

echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/config/PagerDutyConfig.java
kirangodishala: No. I added as it gives flexibility to change it. Also it will be useful for testing, though there are no tests for now.

echo-telemetry/src/main/java/com/netflix/spinnaker/echo/config/TelemetryConfig.kt
dbyron-sf: Seems like some behavior changes here:
1. `stats.connectionTimeoutMilils` and `stats.readTimeoutMillis` are no longer used
2. The default values have changed.  connect timeout was 3000, is now 5000.  read timeout was 5000 is now 120000.
3. It's no longer possible to configure the read and connection timeouts for TelemetryService.  The global values (`ok-http-client.connectTimeoutMs` and `ok-http-client.readTimeoutMs`) now apply to TelemetryService too.

People that depend on this are going to get surprised, and there isn't a great story at the moment.  How can we make it better?

echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/config/PagerDutyConfig.java
dbyron-sf: OK, `@Value` annotations pain me.  Would you mind adding a PagerDutyConfigurationProperties class?  I see `pager-duty.enabled` and `pager-duty.token`.

echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/notification/InteractiveNotificationCallbackHandler.groovy
kirangodishala: I removed by mistake. Added the comment back.

echo-telemetry/src/main/java/com/netflix/spinnaker/echo/config/TelemetryConfig.kt
kirangodishala: Made some changes to reinstate `stats.connectionTimeoutMilils` and `stats.readTimeoutMillis`

echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/config/PagerDutyConfig.java
kirangodishala: Done

echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/slack/SlackNotificationService.groovy
kirangodishala: removed the log statement

echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/notification/GithubNotificationAgent.java
kirangodishala: Added try-catch now

echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/notification/CDEventsNotificationAgent.java
kirangodishala: included additional info

echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/jira/JiraNotificationService.java
kirangodishala: We don't need it. Removed now.

echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/slack/SlackNotificationService.groovy
dbyron-sf: That's a bit more than I was going for.  I think it's still helpful to log the response body, and the outgoing request body (`text`).  I was saying it's OK to drop response?.status and response?.reason.

If we're really going to drop the log message, we no longer need to store the return value of slack.sendMessage, but again, I'd like to keep it.

It's not 100% clear to me whether in the old code, logging `response?.body` could throw an exception.  Accessing the body doesn't look like it can. retrofit.client.Response.getBody returns a TypedInput, but I believe the logging code calls toString, and can that?  There was no exception handling before, so it's probably OK to keep it that way, but okhttp3.ResponseBody.string() can throw an IOException.  I suppose it only gets called when trace logging is enabled, but still...if someone wants this log message it would be a bummer to die trying to log it.

echo-core/src/test/groovy/com/netflix/spinnaker/echo/services/Front50ServiceSpec.groovy
dbyron-sf: Aaah, I see now.  I like the new way better.  No more wacky response-building method.

echo-notifications/src/main/java/com/netflix/spinnaker/echo/notification/GithubNotificationAgent.java
```diff
@@ -128,22 +128,20 @@ public void sendNotifications(
...
   private String getBranchCommit(String repo, String sha) {
-    Response response = githubService.getCommit("token " + token, repo, sha);
-    ObjectMapper objectMapper = EchoObjectMapper.getInstance();
-    GithubCommit message = null;
+    GithubCommit message;
     try {
-      message = objectMapper.readValue(response.getBody().in(), GithubCommit.class);
-    } catch (IOException e) {
+      message = Retrofit2SyncCall.execute(githubService.getCommit("token " + token, repo, sha));
+    } catch (SpinnakerServerException e) {
```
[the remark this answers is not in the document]
kirangodishala: [Here](https://github.com/spinnaker/kork/blob/master/kork-retrofit/src/main/java/com/netflix/spinnaker/kork/retrofit/Retrofit2SyncCall.java#L36) the IOException is getting converted to SpinnakerNetworkException. And any other exception(while performing `call.execute()`) could be one of SpinnakerServerException & it's subclasses.

echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/slack/SlackNotificationService.groovy
kirangodishala: Ok. I added log.trace with a try-catch and included the log in the catch block as well with more info about the error.

echo-notifications/src/main/java/com/netflix/spinnaker/echo/notification/GithubNotificationAgent.java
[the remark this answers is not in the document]
dbyron-sf: Aaah, nice.  Thanks for the link.

echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/slack/SlackNotificationService.groovy
dbyron-sf: Thanks.

echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/slack/SlackNotificationService.groovy
```diff
@@ -62,8 +62,12 @@ class SlackNotificationService implements NotificationService {
           new SlackAttachment(subject, text, (InteractiveActions)notification.interactiveActions),
           address, true)
       }
-      log.trace("Received response from Slack: {} {} for message '{}'. {}",
-        response?.status, response?.reason, text, response?.body)
+      try {
+        log.trace("Received response from Slack: {}", response?.string())
+      } catch (IOException e) {
+        log.trace("Received response from Slack but unable to deserialize : {}", e.message)
```
dbyron-sf, line 68:
```suggestion
        log.trace("Received response from Slack for message '{}' but unable to deserialize", text, e)
```

dbyron-sf (approved):

PR #1476, opened 2025-02-26
fix(retrofit2):  fix `baseUrl must end in /` error

kirangodishala:
Retrofit2 client configuration doesn't allow baseUrl without tailing slash(`/`) if there is a slash in the middle. As a result  `https://somehost` and `https://somehost:3333` becomes valid but `https://somehost/api` throws `java.lang.IllegalArgumentException: baseUrl must end in /`. 

This PR addresses the issue by checking and adding trailing `/` if missing and removing the leading `/` from the endpoint.

### DIFF
--- a/echo-core/src/main/java/com/netflix/spinnaker/echo/util/RetrofitUtils.java
+++ b/echo-core/src/main/java/com/netflix/spinnaker/echo/util/RetrofitUtils.java
@@ -1,0 +1,42 @@
+/*
+ * Copyright 2025 OpsMx, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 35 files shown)


echo-core/src/main/java/com/netflix/spinnaker/echo/util/RetrofitUtils.java
```diff
@@ -0,0 +1,34 @@
+/*
+ * Copyright 2025 Netflix, Inc.
```
dbyron-sf, line 2:
```suggestion
 * Copyright 2025 OpsMx, Inc.
```
?

echo-core/src/main/java/com/netflix/spinnaker/echo/util/RetrofitUtils.java
```diff
@@ -0,0 +1,34 @@
...
+ * limitations under the License.
+ */
+
+package com.netflix.spinnaker.echo.util;
+
+import okhttp3.HttpUrl;
+
+public class RetrofitUtils {
+
+  public static String getBaseUrl(String suppliedBaseUrl) {
```
dbyron-sf, line 23: javadoc please.  Seems like we're gonna need this in kork, but this is fine at least for now.

dbyron-sf (approved):

dbyron-sf:

PR #1478, opened 2025-02-27
fix(retrofit2): fix `A @path parameter must not come after a @Query` error

kirangodishala:
Retrofit2, unlike retrofit1, is enforcing ordering of `@Path` and `@Query` params in the retrofit2 api interface method signature. 
When a `@Path` param is positioned after `@Query` param, then retrofit2 throws the following exception:
```
java.lang.IllegalArgumentException: A @Path parameter must not come after a @Query
```

There are two such cases which are fixed in this PR.


jasonmcintosh (approved):
Looks good thanks!

echo-core/src/test/java/com/netflix/spinnaker/echo/services/IgorServiceTest.java
```diff
@@ -0,0 +1,117 @@
...
+        get(urlEqualTo("/builds/properties/1/fileName/master?job=job"))
+            .willReturn(aResponse().withStatus(200).withBody("{\"mesage\":\"success\"}")));
+
+    Retrofit2SyncCall.execute(
+        igorService.getPropertyFileWithJobQueryParameter(1, "fileName", "master", "job"));
+
+    verify(1, getRequestedFor(urlEqualTo("/builds/properties/1/fileName/master?job=job")));
+  }
+
+  private interface IgorService {
```
dbyron-sf, line 102: Can't we use the "real" IgorService in a test?

dbyron-sf: That's OK.  I think it's better to have tests that cover the actual code.

kirangodishala: Ok, will change shortly.

dbyron-sf (approved):

dbyron-sf:

PR #1480, opened 2025-03-01
fix(rest): add Retrofit2SyncCall.execute to RestService in RestEventService

kirangodishala:
- added a test to demonstrate that RestEventService.sendEvent is failing to make the RestService.recordEvent api call
- fixed the issue and updated the test accordingly

### DIFF
--- a/echo-rest/src/test/java/com/netflix/spinnaker/echo/events/RestEventServiceTest.java
+++ b/echo-rest/src/test/java/com/netflix/spinnaker/echo/events/RestEventServiceTest.java
@@ -38,22 +38,23 @@ import io.github.resilience4j.circuitbreaker.CircuitBreakerConfig;
 import io.github.resilience4j.circuitbreaker.CircuitBreakerRegistry;
 import java.util.Map;
 import okhttp3.OkHttpClient;
-import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.AfterAll;
+import org.junit.jupiter.api.BeforeAll;
 import org.junit.jupiter.api.Test;
 import retrofit2.Retrofit;
 import retrofit2.converter.jackson.JacksonConverterFactory;
 
 public class RestEventServiceTest {
-  private Map<String, Object> eventMap;
-  private RestEventService restEventService;
-  private RestUrls.Service service;
+  private static Map<String, Object> eventMap;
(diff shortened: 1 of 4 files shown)


echo-rest/src/test/java/com/netflix/spinnaker/echo/events/RestEventServiceTest.java
```diff
@@ -0,0 +1,110 @@
...
+import org.junit.jupiter.api.Test;
+import retrofit2.Retrofit;
+import retrofit2.converter.jackson.JacksonConverterFactory;
+
+public class RestEventServiceTest {
+  private Map<String, Object> eventMap;
+  private RestEventService restEventService;
+  private RestUrls.Service service;
+
+  WireMockServer wireMockServer;
```
dbyron-sf, line 51: Pretty sure this way of using wiremock results in less code, but let's stick with what's here and get the fixes out.
```
  @RegisterExtension
  static WireMockExtension wiremock =
      WireMockExtension.newInstance().options(wireMockConfig().dynamicPort()).build();
```
Although, if we stick with this, does something need to call `wireMocKServer.stop()` ?

echo-rest/src/test/java/com/netflix/spinnaker/echo/events/RestEventServiceTest.java
```diff
@@ -0,0 +1,110 @@
...
+  int port;
+  String baseUrl = "http://localhost:PORT/api";
+
+  @BeforeEach
+  void setup() {
+    wireMockServer = new WireMockServer(WireMockConfiguration.options().dynamicPort());
```
dbyron-sf, line 57: Could this happen in a `@BeforeOnce` ?

kirangodishala: Done

echo-rest/src/test/java/com/netflix/spinnaker/echo/events/RestEventServiceTest.java
kirangodishala: Added wireMockServer.stop()

dbyron-sf (approved):

dbyron-sf:

PR #1485, opened 2025-03-04
fix(notifications): fix MicrosoftTeamsConfig which has unused retrofit1 parameters

kirangodishala:
- fixed MicrosoftTeamsConfig
- added a test to verify that MicrosoftTeamsService gets created now

### DIFF
--- a/echo-notifications/src/test/java/com/netflix/spinnaker/echo/config/MicrosoftTeamsServiceTest.java
+++ b/echo-notifications/src/test/java/com/netflix/spinnaker/echo/config/MicrosoftTeamsServiceTest.java
@@ -1,0 +1,43 @@
+/*
+ * Copyright 2025 OpsMx, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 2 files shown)


echo-notifications/src/test/java/com/netflix/spinnaker/echo/config/MicrosoftTeamsServiceTest.java
```diff
@@ -0,0 +1,44 @@
...
+@SpringBootTest(
+    classes = {
+      MicrosoftTeamsConfig.class,
+      Retrofit2TestConfig.class,
+      Retrofit2BasicLogTestConfig.class
+    },
+    properties = "microsoftteams.enabled=true")
+public class MicrosoftTeamsServiceTest {
+
+  @Autowired(required = false)
```
dbyron-sf, line 37: why required = false here?

kirangodishala: to avoid any error that may occur before even the control reaches test. But I think it's not needed, will remove.

kirangodishala: It's to avoid any error that would occur while autowiring. But I think it's not needed in our scenario. Removed now.

dbyron-sf (approved):

PR #1499, opened 2025-04-12
fix(ci/gcb): Fix silently failing update build call.

himanhsugusain:
Fixes the issue where cloud build event were not updating cloud build status, which in turn fails to update the UI with correct stage status.

### DIFF
--- a/echo-notifications/src/main/java/com/netflix/spinnaker/echo/notification/GoogleCloudBuildNotificationAgent.java
+++ b/echo-notifications/src/main/java/com/netflix/spinnaker/echo/notification/GoogleCloudBuildNotificationAgent.java
@@ -21,6 +21,7 @@ import com.netflix.spinnaker.echo.api.events.EventListener;
 import com.netflix.spinnaker.echo.model.pubsub.MessageDescription;
 import com.netflix.spinnaker.echo.services.IgorService;
 import com.netflix.spinnaker.kork.core.RetrySupport;
+import com.netflix.spinnaker.kork.retrofit.Retrofit2SyncCall;
 import com.netflix.spinnaker.security.AuthenticatedRequest;
 import lombok.RequiredArgsConstructor;
 import lombok.extern.slf4j.Slf4j;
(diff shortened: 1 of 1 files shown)


dbyron-sf:
It's probably a bit involved to add a test that fails without the fix, but we need one.  Perhaps @kirangodishala can point to an example to follow?

echo-notifications/src/main/java/com/netflix/spinnaker/echo/notification/GoogleCloudBuildNotificationAgent.java
```diff
@@ -43,23 +44,29 @@ public class GoogleCloudBuildNotificationAgent implements EventListener {
 
   @Override
   public void processEvent(Event event) {
+    log.info("Process Event: {}", event.getDetails());
```
jasonmcintosh, line 47: NIT:  This can be a LOT of logs potentially - should be debug :)

echo-notifications/src/main/java/com/netflix/spinnaker/echo/notification/GoogleCloudBuildNotificationAgent.java
```diff
@@ -43,23 +44,29 @@ public class GoogleCloudBuildNotificationAgent implements EventListener {
...
+                                messageDescription.getMessageAttributes().get("buildId"),
+                                messageDescription.getMessageAttributes().get("status"),
+                                RequestBody.create(
+                                    messageDescription.getMessagePayload(),
+                                    MediaType.parse("application/json"))))),
+            5,
+            2000,
+            false);
+      } catch (Exception e) {
+        log.error("Failed to update google build status: {}", e);
```
jasonmcintosh, line 68: Hrmm, wouldn't this NORMALLY log elsewhere on the processing?   E.g. https://github.com/spinnaker/echo/blob/4c0e2959f1a9613d23011e316df98b5862bf44e5/echo-core/src/main/java/com/netflix/spinnaker/echo/events/EventPropagator.java#L80 should have caught it there... if it wasn't handled someplace else on the other event listeners.  Kinda makes me wonder if there's some other missing stuff in echo.

kirangodishala:
> It's probably a bit involved to add a test that fails without the fix, but we need one. Perhaps @kirangodishala can point to an example to follow?

@himanhsugusain - you can refer [this PR](https://github.com/spinnaker/echo/pull/1480) to add a test to demonstrate the issue first and then add the fix.

dbyron-sf:
Replaced by https://github.com/spinnaker/echo/pull/1500.

